PR #113, opened 2016-07-17
Specify the command requires Python 2 in setup

odinho:
The  command will already say so when you try to use it.  But getting
the error already in installation or as metadata on pypi can be useful.


jcnelson:
Thanks @Velmont!

PR #143, opened 2016-10-01
streamline cli to be pythonic

pandu-rao:


blockstack_client/app.py
```diff
@@ -53,60 +58,64 @@ def app_register( name, app_name, app_account_id, app_url, app_storage_drivers=N
...
     pub_hex = keylib.ECPrivateKey(pk_hex).public_key().to_hex()
 
     # preferred storage drivers?
     conf = config.get_config(config_path)
-    if app_storage_drivers is None:
-        app_storage_drivers = conf['storage_drivers']
-
-    if proxy is None:
-        proxy = get_default_proxy(config_path=config_path)
+    app_storage_drivers = app_storage_drivers or conf['storage_drivers']
```
jcnelson, line 101: I'm a little confused by this choice of syntax.  Is there a chance that this could accidentally evaluate to a boolean, such as by way of misplaced parenthesis?  The reason we went with the two-line format (if xxx is None:) is to make it unambiguous that we're conditionally assigning a variable a value if it has not yet been initialized.

blockstack_client/cli.py
```diff
@@ -52,106 +46,97 @@
...
     import blockstack_client.actions as builtin_methods
-    return get_methods("cli_advanced_", builtin_methods )
+    all_methods = get_methods('cli_', builtin_methods)
+    advanced_methods = get_cli_advanced_methods()
+
+    return list(set(all_methods) - set(advanced_methods))
 
 
-def prompt_args( arginfolist, prompt_func ):
+def prompt_args(args, prompt_func):
```
jcnelson: Could we change `args` back to `arginfolist`?  The reason here is that `args` is often used to refer to a `Namespace(...)` instantiation returned by `argparse` (in blockstack_client/actions.py).  The variable name `arginfolist` here is meant to clarify that this is a list of information about the arguments (i.e. generated by blockstack_client/method_parser.py).

blockstack_client/client.py
```diff
@@ -77,38 +54,35 @@ def session(conf=None, config_path=CONFIG_PATH, server_host=None, server_port=No
     Returns the API proxy object.
     """
 
-    if conf is None and config_path is not None:
+    if not conf and config_path:
```
jcnelson: I'd prefer keeping this as `conf is None` instead of `not conf`.  The reason is that I'm trying to avoid situations where class operator overloading (e.g. of `__nonzero__()` or `__bool__()`) could introduce hard-to-decipher bugs and add accidental complexity.  If `conf` is neither `None` nor a `dict` (or something that behaves like a `dict`), I'd prefer for an exception to be raised.

blockstack_client/client.py
```diff
@@ -77,38 +54,35 @@ def session(conf=None, config_path=CONFIG_PATH, server_host=None, server_port=No
...
-        if storage_drivers is None:
-            storage_drivers = conf['storage_drivers']
-        if metadata_dir is None:
-            metadata_dir = conf['metadata']
-        if spv_headers_path is None:
-            spv_headers_path = conf['blockchain_headers']
-
-    if storage_drivers is None:
-        log.error("No storage driver(s) defined in the config file.  Please set 'storage=' to a comma-separated list of drivers")
+    if conf:
```
jcnelson: Also pointing these out as cases where checking for None explicitly would be preferred (since a lot could be going on behind the scenes here).

blockstack_client/client.py
```diff
@@ -149,36 +124,40 @@ def register_storage(storage_impl, conf):
...
+
+    key = key or {}
     if 'error' in key:
-        log.debug("Failed to fetch analytics key: %s" % key['error'])
-        return None 
+        log.debug('Failed to fetch analytics key: {}'.format(key['error']))
+        return
+
+    key = key.get('analytics_key')
+    if key:
```
jcnelson: Likewise, could we use `key is not None` or `key != None` in this case?

blockstack_client/client.py
```diff
@@ -149,36 +124,40 @@ def register_storage(storage_impl, conf):
...
-            key = proxy.get_analytics_key( uuid )
-        except Exception, e:
-            log.debug("Failed to get analytics key")
-            return None
+    if key:
+        return key
 
+    try:
+        proxy = proxy or get_default_proxy()
+        # TODO: proxy does not have a get_analytics_key function.
```
jcnelson: `proxy` here should be an instance of `BlockstackRPCClient`, which in turn dispatches method calls to a class that inherits from `xmlrpclib.ServerProxy`.  Method binding is late here--`get_analytics_key` is an RPC method on the server.

blockstack_client/config.py
```diff
@@ -987,62 +994,64 @@ def get_app_config(app_name, path=CONFIG_PATH):
...
-    v1_major, v1_minor, v1_patch = v1_parts[0].strip(), v1_parts[1].strip(), v1_parts[2].strip()
-    v2_major, v2_minor, v2_patch = v2_parts[0].strip(), v2_parts[1].strip(), v2_parts[2].strip()
-    if v1_major != v2_major:
-        return False
+    v1_major, v1_minor, v1_patch = get_version_parts(v1_parts, str)
+    v2_major, v2_minor, v2_patch = get_version_parts(v2_parts, str)
 
-    if v1_minor != v2_minor:
-        return False
+    # TODO: check why patch versions are not compared. see function docstring.
```
jcnelson: Good catch!  The doc string is wrong; we only care about the major and minor versions.

blockstack_client/config.py
```diff
@@ -987,62 +994,64 @@ def get_app_config(app_name, path=CONFIG_PATH):
...
-            parser.write(configfile)
+def get_version_parts(whole, func):
+    return [func(_.strip()) for _ in whole[0:3]]
 
 
-def semver_match( v1, v2 ):
+def semver_match(v1, v2):
     """
     Verify that two semantic version strings match:
     the major, minor, and patch versions must be equal.
```
jcnelson: My bad; this should say 'the major and minor versions must be equal; the patch version is not considered' or the like.

blockstack_client/config.py
```diff
@@ -313,403 +317,406 @@ def get_logger( debug=DEBUG ):
...
+                f.flush()
+                os.fsync(f.fileno())
+        except Exception as e:
+            log.exception(e)
+            return None
+
+    return u
+
+
+def configure(config_file=CONFIG_PATH, force=False, interactive=True):
```
jcnelson, line 566: Note to myself--the interactive variant of this method is not covered by the integration tests.  Changes here will either need to be tested by hand before merging, or we'll need an `expect` script handy.

jcnelson (commented):
Hi @pandu-rao, thank you so much for taking the time to clean up all of this code!  I left a few minor comments, but overall it looks great.  Because the diff is so big, we probably won't be able to merge it into 0.14, but we'll strive to merge it into 0.14.1 (once we've been able to run all the integration tests over it).

jcnelson:
Also, CC'ing @muneeb-ali and @shea256 to see if they'd like to give a code review as well.

blockstack_client/cli.py
pandu-rao: Got that. Not an issue. Will change this.

blockstack_client/client.py
pandu-rao: Makes sense.

blockstack_client/client.py
pandu-rao: Makes sense.

blockstack_client/client.py
pandu-rao: I'll change it to `key is not None`

blockstack_client/app.py
pandu-rao: >  Is there a chance that this could accidentally evaluate to a boolean, such as by way of misplaced parenthesis?

No, that won't happen. See below.

I've found that treating '', None, [] and {} in this manner works well in practice. Because the subsequent lines throw an error when some variable is used incorrectly.

However, I can use the explicit check for None. If that's what you prefer.

```
>>> False or {'key':'value'}['key']
'value'
>>> (False or {'key':'value'}['key'])
'value'
>>> 
```

pandu-rao: Another way to accomplish the assignment is as follows:

```
foo = bar if foo is None else foo
```

This is explicit, compact and (usually) readable. We can choose this for most cases and fall back to the block variant only in cases like this:

```
long_variable_name = longer_variable_name if long_variable_name is None else long_variable_name
```

For long variable names, this reads better:

```
if long_variable_name is None:
    long_variable_name = longer_variable_name 
```

blockstack_client/rpc_runner.py
```diff
@@ -23,74 +26,56 @@
...
-                sys.exit(0)
-            else:
-                sys.exit(1)
-
-
+            rc = local_rpc_start(portnum, config_dir=config_dir)
+            sys.exit(0 if rc else 1)
+    else:
+        print(usage, sys.stderr)
+        sys.exit(1)
```
jcnelson, line 81: Looks good to me!  Thank you for taking the time to make this module more Pythonic!

blockstack_client/method_parser.py
```diff
@@ -143,16 +149,14 @@ def build_method_subparsers( subparsers, method_infos, include_args=True, includ
...
             for arg in method_info['args']:
-                subparser.add_argument( arg['name'], type=arg['type'], help=arg['help'] )
+                subparser.add_argument(arg['name'], type=arg['type'], help=arg['help'])
 
         if include_opts:
             for opt in method_info['opts']:
-                subparser.add_argument( opt['name'], type=opt['type'], nargs='?', help=opt['help'] )
+                subparser.add_argument(opt['name'], type=opt['type'], nargs='?', help=opt['help'])
 
     return True
```
jcnelson, line 174: Thanks for taking the time to consolidate all those constants as variables :)

blockstack_client/config.py
```diff
@@ -45,12 +48,16 @@ def get_logger(debug=DEBUG):
 
 log = get_logger('blockstack-client')
 
+BLOCKSTACK_TEST = os.environ.get('BLOCKSTACK_TEST')
+BLOCKSTACK_TEST_NODEBUG = os.environ.get('BLOCKSTACK_TEST_NODEBUG')
+BLOCKSTACK_DEBUG = os.environ.get('BLOCKSTACK_DEBUG')
+BLOCKSTACK_TEST_FIRST_BLOCK = os.environ.get('BLOCKSTACK_TEST_FIRST_BLOCK')
```
jcnelson: Thanks for making this particular change!  Much more ergonomic.

blockstack_client/config.py
```diff
@@ -808,9 +810,11 @@ def get_tx_broadcaster(config_path=CONFIG_PATH):
     try:
         blockchain_broadcaster = connect_utxo_provider(writer_opts)
         return blockchain_broadcaster
-    except:
+    except Exception as e:
         log.exception(e)
```
jcnelson, line 867: Good catch!

blockstack_client/keys.py
```diff
@@ -21,124 +21,102 @@
...
+        return singlesig_privkey_to_string(privkey_info)
 
-    elif is_multisig( privkey_info ):
-        return multisig_privkey_to_string( privkey_info )
+    if is_multisig(privkey_info):
+        return multisig_privkey_to_string(privkey_info)
 
-    else:
-        return None
+    # TODO: check if we should be returning '' here
```
jcnelson: I prefer None personally.  The other methods that call this one use it to print out private keys, and returning None (to me) is a more straightforward way of saying "this is an invalid key."

blockstack_client/keys.py
```diff
@@ -245,82 +229,80 @@ def decrypt_private_key_info( privkey_info, password ):
...
-
-        elif is_singlesig( payment_privkey ):
-            pk_payment = virtualchain.BitcoinPrivateKey( payment_privkey ).to_hex()
-            ret['payment_privkey'] = pk_payment
-
         else:
-            raise ValueError("Invalid payment key info")
+            raise ValueError('Invalid owner key info')
+
+    if payment_privkey is None:
```
jcnelson, line 288: For legibility reasons, I'd prefer it if we keep this in the original notation:

```
if payment_privkey is not None:
   # process it

return ret
```

blockstack_client/keys.py
```diff
@@ -377,107 +357,102 @@ def get_data_or_owner_privkey( user_zonefile, owner_address, wallet_keys=None, c
...
+    if data_privkey is not None:
+        return {'status': True, 'privatekey': data_privkey}
+
+    # This is legacy code here.  The only time this should happen is
+    # when the user has a single owner key, and does not have a
+    # separate data key.
+    log.warn('No data private key set.  Falling back to owner keypair.')
+    owner_privkey_info = get_owner_privkey_info(wallet_keys=wallet_keys, config_path=config_path)
+    if owner_privkey_info is None:
+        # FIXME: We cannot both raise and return
```
jcnelson, line 370: Thanks for catching that.  I added a `raise` to see if I could get this method to fail more loudly in this case and the case below.  I probably added it while testing, and just forgot to clean up the `return`.

blockstack_client/tx.py
```diff
@@ -21,161 +21,163 @@
...
         log.exception(e)
-        log.error("Failed to broadcast transaction %s" % signed_tx)
+        log.error('Failed to broadcast transaction {}'.format(signed_tx))
         return {'error': 'Failed to broadcast transaction (caught exception)'}
 
     if 'error' in resp:
-        log.error("Failed to broadcast transaction: %s" % resp['error'])
+        log.error('Failed to broadcast transaction: {}'.format(resp['error']))
 
     return resp
```
jcnelson, line 183: Awesome, looks good!  Thank you for cleaning up this file!

blockstack_client/scripts.py
```diff
@@ -394,16 +405,10 @@ def tx_get_unspents(address, utxo_client, min_confirmations=TX_MIN_CONFIRMATIONS
...
-    # filter minimum confirmations 
-    ret = []
-    for d in data:
-        if not d.has_key('confirmations') or d['confirmations'] >= min_confirmations:
-            ret.append(d)
-
-    return ret
-
+    # filter minimum confirmations
+    return [d for d in data if d.get('confirmations', 0) >= min_confirmations]
```
jcnelson, line 414: Looks good so far, but I will test rigorously since this file directly influences how much money a user spends per transaction.  Thank you for taking the time to clean it up!

blockstack_client/keys.py
```diff
@@ -516,27 +490,23 @@ def get_pubkey_addresses( pubkey ):
...
+        uncompressed_address = pubkey.address()
+        compressed_address = compress(pubkey_bin)
+        hashed_address = bin_hash160(compressed_address)
+        compressed_address = bin_hash160_to_address(hashed_address, version_byte=version_byte)
     else:
-        raise Exception("Invalid public key")
+        raise Exception('Invalid public key')
 
-    return (compressed_address, uncompressed_address)
+    return compressed_address, uncompressed_address
```
jcnelson, line 511: Looks good; thank you for cleaning this up!  I will also be testing this one rigorously, since it affects how the user's wallet gets loaded and stored.

blockstack_client/keys.py
pandu-rao: Now that you mention it ...

```
if payment_privkey is not None:
   # process it
```

... maintains the symmetry between the `owner_privkey` and `payment_privkey` blocks.

In fact, both blocks can be further consolidated in this manner:
https://gist.github.com/pandu-rao/8d20134dd0b2c1f6be4611b16fd93164

blockstack_client/user.py
```diff
@@ -291,11 +293,9 @@ def put_immutable_data_zonefile(user_zonefile, data_id, data_hash, data_url=None
...
 
-    txtrec = None
+    txtrec = '#{}'.format(data_hash)
     if data_url is not None:
-        txtrec = '{}#{}'.format(data_url, data_hash)
-    else:
-        txtrec = '#{}'.format(data_hash)
+        txtrec = '{}{}'.format(data_url, txtrec)
 
     user_zonefile.setdefault('txt', [])
```
jcnelson, line 301: Looks good!

blockstack_client/user.py
```diff
@@ -21,256 +21,245 @@
...
+    for url in urls:
+        urirec = url_to_uri_record(url)
+        user['uri'].append(urirec)
 
-   return user
+    return user
 
 
-def is_user_zonefile( d ):
+def is_user_zonefile(d):
```
jcnelson, line 116: We could use `jsonschema` here to make this method cleaner.

blockstack_client/user.py
```diff
@@ -349,443 +332,429 @@ def get_immutable_url_from_txt( txtrec ):
...
 
-       if d_id == data_id:
-           if ret is None:
-               ret = h
-           elif type(ret) != list:
-               ret = [ret]
-               ret.append(h)
+        if ret is None:
+            ret = h
+        elif not isinstance(ret, list):
```
jcnelson, line 455: Great catch finding this bug!

blockstack_client/user.py
```diff
@@ -349,443 +332,429 @@ def get_immutable_url_from_txt( txtrec ):
...
-            if h != data_hash:
-                continue 
+            h = get_immutable_hash_from_txt(txtrec['txt'])
+            if data_hash != h:
+                continue
 
-            url = get_immutable_url_from_txt( txtrec['txt'] )
+            url = get_immutable_url_from_txt(txtrec['txt'])
         except:
+            # TODO: Check if we should use more fine-grained exception handling
```
jcnelson, line 483: I'd generally prefer to err on the side of false-negatives than false-positives.  It's better to assume there is no valid URL if `get_immutable_hash_from_txt` or `get_immutable_url_from_txt` fail, even if they fail for unexpected reasons.  I think the best thing here would be to try to catch individual exceptions, but retain a catch-all `except:` block that printed a very loud error message so I can go fix it later :)

blockstack_client/user.py
```diff
@@ -349,443 +332,429 @@ def get_immutable_url_from_txt( txtrec ):
...
     for txtrec in user_zonefile['txt']:
         try:
             d_id = txtrec['name']
-            h = get_immutable_hash_from_txt( txtrec['txt'] )
+            h = get_immutable_hash_from_txt(txtrec['txt'])
             assert storage.is_valid_hash(h)
-            ret.append( (d_id, h) )
+            ret.append((d_id, h))
         except:
+            # TODO: Check if we should use more fine-grained exception handling
```
jcnelson, line 507: Same idea as above.  If we can think of finer-grained exceptions, we should add them, but let's retain the catch-all `except:` and have it print a very loud error message.

blockstack_client/user.py
```diff
@@ -349,443 +332,429 @@ def get_immutable_url_from_txt( txtrec ):
...
-       version = None
-       try:
-           unpacked_data_id, version = unpack_mutable_data_md( packed_data_txt )
-       except:
-           continue
+        unpacked_data_id, version = None, None
+        try:
+            unpacked_data_id, version = unpack_mutable_data_md(packed_data_txt)
+        except:
+            # TODO: Check if we should use more fine-grained exception handling
```
jcnelson, line 594: Same as above

blockstack_client/user.py
```diff
@@ -349,443 +332,429 @@ def get_immutable_url_from_txt( txtrec ):
...
             continue
 
-        unpacked_data_id = None
-        version = None
+        unpacked_data_id, version = None, None
         try:
-            unpacked_data_id, version = unpack_mutable_data_md( packed_data_txt )
+            unpacked_data_id, version = unpack_mutable_data_md(packed_data_txt)
         except:
+            # TODO: Check if we should use more fine-grained exception handling
```
jcnelson, line 620: Same as above

blockstack_client/user.py
```diff
@@ -349,443 +332,429 @@ def get_immutable_url_from_txt( txtrec ):
...
     """
     uri_records = mutable_info.get('uri')
-    if uri_records is None:
-        return None 
-
-    urls = [u['target'].strip('"') for u in uri_records]
-    return urls
 
+    if uri_records is None:
+        return None
```
jcnelson, line 759: Looks good, thanks!

blockstack_client/user.py
```diff
@@ -58,7 +58,7 @@ def url_to_uri_record(url, datum_name=None):
             try:
                 assert len(scheme.split('+')) == 2
                 scheme, proto = scheme.split('+')
-            except:
+            except (AssertionError, ValueError):
                 msg = 'WARN: Scheme "{}" has no known transport protocol'
                 log.debug(msg.format(scheme))
```
jcnelson, line 64: Much better than what we had before :)

blockstack_client/profile.py
```diff
@@ -195,90 +176,112 @@ def load_name_profile(name, user_zonefile, data_address, owner_address, use_zone
...
+        msg = 'Missing user data public key and address; cannot verify profile'
+        raise Exception(msg)
 
     if user_data_pubkey is None:
-        log.warn("No data public key set; falling back to hash of data and/or owner public key for profile authentication")
+        msg = (
+            'No data public key set; falling back to hash of data '
+            'and/or owner public key for profile authentication'
+        )
+        log.warn(msg)
```
jcnelson, line 203: Just going forward, which is more Pythonic?  Having a separate `msg` variable, or making the `log.warn()` call span multiple lines?

blockstack_client/profile.py
```diff
@@ -635,51 +632,49 @@ def zonefile_data_publish(fqu, zonefile_txt, server_list, wallet_keys=None):
     successful_servers = []
```
jcnelson, line 686: Note to self: use `jsonschemas` here to validate the returned response.

blockstack_client/profile.py
```diff
@@ -694,27 +689,27 @@ def zonefile_data_replicate( fqu, zonefile_data, tx_hash, server_list, config_pa
...
 
     # replicate to blockstack servers
-    res = zonefile_data_publish( fqu, zonefile_data, server_list ) 
+    res = zonefile_data_publish(fqu, zonefile_data, server_list)
     if 'error' in res:
         return res
 
     return {'status': True, 'servers': res['servers']}
-
-
```
jcnelson, line 739: Looks good!

blockstack_client/profile.py
```diff
@@ -300,8 +301,7 @@ def get_name_zonefile(name, storage_drivers=None, create_if_absent=False, proxy=
     Even then, only the *public* data key needs to be set.
     """
 
-    if proxy is None:
-        proxy = get_default_proxy()
+    proxy = get_default_proxy() if proxy is None else proxy
 
     # find name record first
     if name_record is None:
```
jcnelson, line 316: Looks good!

blockstack_client/wallet.py
```diff
@@ -163,96 +150,102 @@ def get_child_keypairs(self, count=1, offset=0, include_privkey=False):
...
     By default, the owner and payment keys will be key bundles set up to require 2-of-3 signatures.
-    @payment_privkey_info, @owner_privkey_info, and @data_privkey_info can either be individual private keys, or 
+    @payment_privkey_info, @owner_privkey_info, and @data_privkey_info can either be individual private keys, or
     dicts with {'redeem_script': ..., 'private_keys': ...} defined.
 
     Return the new wallet on success.
     Return {'error': ...} on failure
     """
 
+    def get_privkey_info(multisig):
```
jcnelson: Suggest renaming to `make_privkey_info`, since I'm pretty sure `get_privkey_info_...` is already a commonly-used method prefix.

blockstack_client/wallet.py
```diff
@@ -726,72 +718,67 @@ def get_wallet(config_path=CONFIG_PATH):
...
-                log.error("RPC error: %s" % wallet_data['error'])
-                raise Exception("RPC error: %s" % wallet_data['error'])
-
-        except Exception, e:
-            log.exception(e)
-            return {'error': 'Failed to get wallet'}
+    if not local_proxy:
+        return None
 
+    try:
```
jcnelson, line 724: Note to self: use `jsonschemas` to validate the RPC daemon's response.

blockstack_client/wallet.py
```diff
@@ -924,9 +916,8 @@ def dump_wallet(config_path=CONFIG_PATH, wallet_path=None, password=None):
         if 'error' in res:
             return res
 
-    wallet = get_wallet( config_path=config_path )
+    wallet = get_wallet(config_path=config_path)
     if wallet is None:
         return {'error': 'Failed to load wallet'}
 
     return {'status': True, 'wallet': wallet}
-
```
jcnelson, line 932: Looks good, thanks!  Also looking forward to using your wallet schema to validate wallets loaded from disk :)

blockstack_client/storage.py
```diff
@@ -51,686 +46,704 @@
...
 
-   # sanity check: only support single-sig private keys 
-   if not keys.is_singlesig(privatekey):
-       log.error("Only single-signature data private keys are supported")
-       return False
 
-   fq_data_id = str(fq_data_id)
-   assert is_fq_data_id( fq_data_id ) or is_valid_name(fq_data_id), "Data ID must be fully qualified or must be a valid blockchain ID (got %s)" % fq_data_id
-   assert privatekey is not None
+# TODO: Check use of mutable default arguments
```
jcnelson: Not sure what you mean here?

blockstack_client/storage.py
```diff
@@ -1184,10 +1203,13 @@ def __init__(self, full_response=False, config_path=CONFIG_PATH):
...
+    def blockstack_open(self, req):
         """
         Open a blockstack URL
         """
-        bh = BlockstackURLHandle( req.get_full_url(), full_response=self.full_response, config_path=self.config_path )
-        return bh
+        bh = BlockstackURLHandle(
+            req.get_full_url(), full_response=self.full_response,
+            config_path=self.config_path
+        )
```
jcnelson, line 1217: Looks good!  Thank you for taking a crack at this :)

blockstack_client/storage.py
```diff
@@ -1193,7 +1191,6 @@ def readlines(self, sizehint=None):
 
 
 class BlockstackHandler(urllib2.BaseHandler):
-
     """
     URL opener for blockstack:// URLs.
     Usable with urllib2.
```
jcnelson, line 1202: Looks good!

blockstack_client/snv.py
```diff
@@ -20,22 +20,9 @@
     You should have received a copy of the GNU General Public License
     along with Blockstack-client. If not, see <http://www.gnu.org/licenses/>.
 """
-import argparse
```
jcnelson, line 23: Wow, I didn't realize all these were no longer necessary!  A sign of how much this module has evolved, I guess :)

blockstack_client/snv.py
```diff
@@ -243,10 +252,11 @@ def txid_to_serial_number(txid, bitcoind_proxy, proxy=None):
         tx_index = block_data['tx'].index(txid)
     except:
         # not actually present
-        log.error("Transaction %s is not present in block %s (%s)" % (txid, block_id, block_hash))
-
-    return "%s-%s" % (block_id, tx_index)
+        msg = 'Transaction {} is not present in block {} ({})'
+        log.error(msg.format(txid, block_id, block_hash))
 
+    # BUG: block_id is undefined
```
jcnelson, line 262: YES!  Thank you for catching this!  This method isn't used anywhere yet, but it will be in the near future (i.e. it converts a hard-to-remember transaction ID (a sha256 hash) into a phone number-like string of digits that identifies the block height and the offset into the block where the transaction occurs (ex: 393543-257)).

blockstack_client/snv.py
```diff
@@ -457,42 +501,54 @@ def get_name_creation_consensus_info(name, blockchain_record, bitcoind_proxy, pr
...
         # we can trust that the consensus-bearing transaction is on the blockchain.
         # now, what's the creation consensus hash's block number?
         # (NOTE: this trusts Blockstack)
-        creation_consensus_block_id = get_block_from_consensus(creation_consensus_hash, proxy=proxy)
-        if type(creation_consensus_hash_id) == dict and 'error' in ret:
+        creation_consensus_block_id = get_block_from_consensus(
+            creation_consensus_hash, proxy=proxy
+        )
+
+        # BUG: undefined variables creation_consensus_hash_id and ret
```
jcnelson, line 541: Good catch!  Thanks!

blockstack_client/snv.py
```diff
@@ -396,31 +426,45 @@ def verify_consensus_hash_from_tx(tx, fqname, candidate_consensus_hash):
...
+    log.error(msg.format(
+        name_consensus_hash_mix.encode('hex'),
+        candidate_consensus_hash,
+        tx_name_consensus_hash_mix.encode('hex')
+    ))
+
+    return None
 
 
 def get_name_creation_consensus_info(name, blockchain_record, bitcoind_proxy, proxy=None, serial_number=None):
```
jcnelson, line 478: Should have mentioned in the comments--this method isn't used anywhere, and I don't know if it'll be used in the future.

blockstack_client/snv.py
```diff
@@ -575,19 +629,19 @@ def find_last_historical_op(history, opcode):
...
 
             if prev_opcode == opcode:
-                return (int(prev_blocks[i + 1]), prev_op)
+                return int(prev_blocks[i + 1]), prev_op
 
-    return (None, None)
+    return None, None
 
 
 def get_name_update_consensus_info(name, blockchain_record, bitcoind_proxy, proxy=None):
```
jcnelson, line 661: Should have also tagged this one as deprecated.

blockstack_client/snv.py
```diff
@@ -897,32 +971,30 @@ def snv_lookup(verify_name, verify_block_id, trusted_serial_number_or_txid_or_co
...
         try:
             trusted_block_id = int(parts[0])
+            # TODO: Check why this variable is unused
             trusted_tx_index = int(parts[1])
         except:
-            log.error("Malformed serial number '%s'" % trusted_serial_number_or_txid_or_consensus_hash)
+            log.error('Malformed serial number "{}"'.format(trusted_serial_number_or_txid_or_consensus_hash))
             return {'error': 'Did not receive a valid serial number'}
 
         trusted_tx = serial_number_to_tx(trusted_serial_number_or_txid_or_consensus_hash, bitcoind_proxy)
```
jcnelson, line 1038: It shouldn't affect correctness, but the idea behind `trusted_block_id` and `trusted_tx_index` was to verify that `trusted_serial_number_or_txid_or_consensus_hash` was a serial number (i.e. $block_height-$tx_index).  We can replace `trusted_serial_number_or_txid_or_consensus_hash` with `'{}-{}' % (trusted_block_id, trusted_tx_index)` in this branch of the conditional.

blockstack_client/snv.py
```diff
@@ -897,32 +971,30 @@ def snv_lookup(verify_name, verify_block_id, trusted_serial_number_or_txid_or_co
...
         if trusted_tx is None:
             return {'error': 'Unable to convert given serial number into transaction'}
 
         # tx must have a consensus hash
+        # TOOD: Check why return values are ignored
         op, payload = parse_tx_op_return(trusted_tx)
```
jcnelson, line 1044: This is an artifact of an earlier version of this file.  The call to `parse_tx_op_return` below can be removed, since `get_consensus_hash_from_tx` takes care of this (before `get_consensus_hash_from_tx` exited, we used the `payload` to extract the `trusted_consensus_hash` based on what `op` was).

blockstack_client/snv.py
```diff
@@ -932,22 +1004,25 @@ def snv_lookup(verify_name, verify_block_id, trusted_serial_number_or_txid_or_co
...
+    historic_namerec = snv_name_verify(
+        verify_name, trusted_block_id, trusted_consensus_hash,
+        verify_block_id, verify_consensus_hash
+    )
 
     return historic_namerec
 
 # backwards compatibility
 lookup_snv = snv_lookup
-
```
jcnelson, line 994: Thank you for taking a crack at this file!  I love that you left comments in here about whether or not variables hadn't been declared, or whether or not variables were unused :)

blockstack_client/data.py
```diff
@@ -278,39 +268,37 @@ def list_update_history( name, current_block=None, proxy=None ):
...
-            info = proxy.getinfo()
-            if info.has_key('last_block_processed'):
-                current_block = int(info['last_block_processed']) + 1
-            elif info.has_key('last_block'):
-                current_block = int(info['last_block']) + 1
-            else:
-                raise Exception("Invalid getinfo reply")
-        except Exception, e:
-            log.error("Invalid getinfo reply")
+        info = proxy.getinfo()
```
jcnelson: Part of the original `try/except` block's job earlier was to catch invalid data returned by `proxy.getinfo()`.  I'm working on a patch to `proxy.py` that raises an exception of the server-replied data doesn't match our schema, but the `try/except` block should still be here to cover that case.

blockstack_client/data.py
```diff
@@ -170,29 +147,40 @@ def delete_mutable_data_version(conf, data_id):
...
         # failed for whatever reason
-        log.warn("Failed to remove version file '%s'" % (version_file_path))
-        return False
+        msg = 'Failed to remove version file "{}"'
+        log.warn(msg.format(version_file_path))
+
+    return False
+
+
+def is_legacy_profile(user_zonefile):
```
jcnelson: I like that this check is now consolidated :)  However, some clarification on the nature of these checks is needed.  The reason for these checks isn't to determine if the profile has a legacy format; the reason is to verify that the data whose hash is on the blockchain corresponds to a well-formed zonefile/profile pair.  Both checks are needed because a specially-crafted legacy profile can also pass as a well-formed zonefile, and we only want to support storage operations if we're certain that the given information is _actually_ a zonefile (i.e. if a user has a legacy profile that also looks like a zonefile, then (s)he should first run the `migrate` command to switch over to a zonefile that is guaranteed to not be a legacy profile).

I would consider calling this method `is_obsolete_zonefile`.

blockstack_client/data.py
```diff
@@ -1225,73 +1272,90 @@ def data_put( blockstack_url, data, proxy=None, wallet_keys=None, data_privkey=N
...
+            parts['blockchain_id'], parts['data_id'],
+            proxy=proxy, wallet_keys=wallet_keys
+        )
+        return ret
 
-        else:
-            ret = delete_mutable( parts['blockchain_id'], parts['data_id'], proxy=proxy, wallet_keys=wallet_keys )
+    # app-specific delete
+    fields = parts['fields']
+    # BUG: required parameter "data_privkey" is not provided as argument
```
jcnelson, line 1355: Great catch!  I double-checked and it looks like `data_get`, `data_delete`, and `data_list` aren't covered by the integration test framework.  I'll make a note to cover them.

blockstack_client/data.py
```diff
@@ -1356,12 +1438,11 @@ def set_data_pubkey( name, data_pubkey, proxy=None, wallet_keys=None, txid=None
...
-    rc = store_name_zonefile( name, user_zonefile, txid )
+    rc = store_name_zonefile(name, user_zonefile, txid)
     if not rc:
         result['error'] = 'Failed to store zonefile'
         return result
 
     # success!
     result['status'] = True
     return result
-
```
jcnelson, line 1367: Awesome work!

blockstack_client/data.py
```diff
@@ -1419,8 +1430,9 @@ def set_data_pubkey(name, data_pubkey, proxy=None, wallet_keys=None, txid=None):
...
 
         update_result = async_update(
-            name, user_zonefile_txt, None, owner_privkey_info, payment_privkey_info,
-            config_path=proxy.conf['path'], proxy=proxy, queue_path=proxy.conf['queue_path']
+            name, user_zonefile_txt, None, owner_privkey_info,
+            payment_privkey_info, config_path=proxy.conf['path'],
+            proxy=proxy, queue_path=proxy.conf['queue_path']
         )
 
         if 'error' in update_result:
```
jcnelson, line 1441: Looks good!

blockstack_client/method_parser.py
```diff
@@ -64,7 +64,8 @@ def parse_methods(method_list):
     command_pattern = re.compile('^command:[ \t]+([^ \t]+)[ ]*(.*)[ ]*$')
     help_pattern = re.compile('^help:[ \t]+(.+)$')
 
-    arg_opt_pattern = '^{}[ \t]+([^ \t]+)[ \t]+\((.+)\)[ \t]+"([^"]+)"$'
+    # NOTE: pattern must be defined using double-quotes
+    arg_opt_pattern = "^{}[ \t]+([^ \t]+)[ \t]+\((.+)\)[ \t]+'([^']+)'$"
     arg_pattern = re.compile(arg_opt_pattern.format('arg:'))
     opt_pattern = re.compile(arg_opt_pattern.format('opt:'))
```
jcnelson, line 71: Awesome!  Thanks for doing this!

blockstack_client/data.py
pandu-rao: Got that. I've added it back.

blockstack_client/keys.py
```diff
@@ -112,7 +112,6 @@ def privkey_to_string(privkey_info):
     if is_multisig(privkey_info):
         return multisig_privkey_to_string(privkey_info)
 
-    # TODO: check if we should be returning '' here
     return None
```
jcnelson, line 117: :+1:

blockstack_client/config.py
```diff
@@ -1045,8 +1046,7 @@ def semver_match(v1, v2):
     v1_major, v1_minor, v1_patch = get_version_parts(v1_parts, str)
     v2_major, v2_minor, v2_patch = get_version_parts(v2_parts, str)
 
-    # TODO: check why patch versions are not compared. see function docstring.
-
+    # NOTE: patch versions are not relevant here.
     return [v1_major, v1_minor] == [v2_major, v2_minor]
```
jcnelson, line 1088: :+1:

blockstack_client/data.py
```diff
@@ -273,12 +273,15 @@ def list_update_history(name, current_block=None, proxy=None):
...
+            info = proxy.getinfo()
+            if 'last_block_processed' in info:
+                current_block = int(info['last_block_processed']) + 1
+            elif 'last_block' in info:
+                current_block = int(info['last_block']) + 1
+            else:
+                raise Exception('Invalid getinfo reply')
+        except Exception as e:
             log.error('Invalid getinfo reply')
             return None
```
jcnelson, line 287: Looks good :)  In my tree, I've patched `proxy.getinfo()` to use `jsonschema` to verify that the result is well-formatted as well.

blockstack_client/data.py
```diff
@@ -1190,7 +1190,7 @@ def list_mutable_data(name, proxy=None, wallet_keys=None):
         # user_profile will contain an error message
         return user_profile
 
-    if is_legacy_profile(user_zonefile):
+    if is_obsolete_zonefile(user_zonefile):
         # zonefile is really a legacy profile
         return {'data': []}
```
jcnelson, line 1199: Thanks for renaming :)

blockstack_client/actions.py
```diff
@@ -127,6 +129,54 @@
...
+        return result
+
+    # continue processing the result
+
+    return result
+
+
 def check_valid_name(fqu):
     """
     Verify that a name is valid.
```
jcnelson, line 184: Looks great!  This will be very helpful for adding future commands.

blockstack_client/wallet.py
```diff
@@ -216,7 +216,7 @@ def get_privkey_info(multisig):
         return {'error': 'Multisig payment private key info is not supported'}
 
     if owner_privkey_info is None:
-        owner_privkey_info = get_privkey_info(multisig)
+        owner_privkey_info = make_privkey_info(multisig)
 
     if not is_singlesig(owner_privkey_info) and not is_multisig(owner_privkey_info):
         return {
```
jcnelson, line 222: Thanks!  This keeps the code a bit clearer.

blockstack_client/storage.py
```diff
@@ -705,15 +709,16 @@ def delete_immutable_data(data_hash, txid, privkey):
...
     go and delete the associated mutable data.
     """
 
     global storage_handlers
 
+    only_use = [] if only_use is None else only_use
+
     # sanity check
     if not keys.is_singlesig(privatekey):
         log.error('Only single-signature data private keys are supported')
```
jcnelson, line 737: I'm so glad you caught this!  This would have been a huge headache to debug.

blockstack_client/method_parser.py
```diff
@@ -97,6 +99,15 @@ def parse_methods(method_list):
...
+            if unsupported_pragmas:
+                log.exception('Unsupported pragmas: {}'.format(unsupported_pragmas))
+                raise ValueError
+
+            if 'rpc' not in command_pragmas:
+                command_pragmas.append('norpc')
+
             command_help = re.findall(help_pattern, help_line)[0]
         except Exception as e:
             log.exception(e)
```
jcnelson, line 113: Awesome!  This makes things clearer.

blockstack_client/profile.py
pandu-rao: For reference. It is better to use a msg variable. This works well if you need to do additional formatting.

blockstack_client/rpc.py
```diff
@@ -578,267 +623,252 @@ def local_rpc_exit_handler( sig, frame ):
...
 
-        if pass_password:
-            del os.environ['BLOCKSTACK_CLIENT_WALLET_PASSWORD']
+    rc = local_rpc_action('start', config_dir=config_dir)
 
-        if rc != 0:
-            log.error("Failed to start RPC endpoint; exit code was %s" % rc)
-            return False
+    # we do not need the password anymore. so remove it from the environment
+    os.environ.pop('BLOCKSTACK_CLIENT_WALLET_PASSWORD', None)
```
jcnelson, line 850: I didn't know `dict` had a `pop()` method :)

blockstack_client/proxy.py
```diff
@@ -94,38 +94,47 @@ class BlockstackRPCClient(object):
...
+                    msg = 'Server replied invalid JSON'
+                    if BLOCKSTACK_TEST is not None:
+                        log.debug('{}: {}'.format(msg, res))
+
+                    log.error(msg)
+                    res = {'error': msg}
+
+                self.log_debug_timeline('end', key, r)
 
                 return res
```
jcnelson, line 474: Looks good to me!

blockstack_client/rpc.py
```diff
@@ -848,8 +878,14 @@ def start_rpc_endpoint(config_dir=blockstack_config.CONFIG_DIR, password=None):
...
-        return {'error': 'Failed to start RPC endpoint (in working directory %s).\nPlease check your password, and verify that the working directory exists and is writeable.' % config_dir}
+        msg = (
+            'Failed to start RPC endpoint (in working directory {}).\n'
+            'Please check your password, and verify that the working '
+            'directory exists and is writeable.'
+        )
+
+        return {'error': msg.format(config_dir)}
 
     return {'status': True}
```
jcnelson, line 891: Looks good to me!

blockstack_client/rpc.py
```diff
@@ -578,267 +623,252 @@ def local_rpc_exit_handler( sig, frame ):
...
+
+    local_rpc_unlink_pidfile(rpc_pidpath)
+    local_rpc_server_stop(rpc_srv)
 
-    local_rpc_unlink_pidfile( rpc_pidpath )
-    local_rpc_server_stop( rpc_srv )
+    return True
+
+
+def rpc_kill(pidpath, pid, sig, unlink_pidfile=True):
```
jcnelson, line 735: Thanks for adding this method!

blockstack_client/config.py
```diff
@@ -360,6 +361,24 @@ def get_logger(debug=DEBUG):
 }
 
 
+# NOTE: duplicated from blockstack-core and streamlined.
```
jcnelson, line 364: I have since modified blockstack-core to use the client's method.

blockstack_client/proxy.py
```diff
@@ -483,83 +915,258 @@ def get_consensus_range(block_id_start, block_id_end, proxy=None):
...
+            if json_is_error(block_resp):
+                return block_resp
+    except ValidationError as e:
+        resp = json_traceback(resp.get('error'))
 
     return resp
 
 
-def get_op_history_rows( name, proxy=None ):
+def get_op_history_rows(name, proxy=None):
```
jcnelson, line 1262: A lot of this functionality used to live in the server.  One of the heaviest-weight I/O-bound RPC requests a client can make is to ask for all of the name operations that occurred at a particular block height (`get_nameops_at`).  Since the server's internal representation of the name state is reminiscent of a version control system (i.e. each name record has a "head" state and a bunch of deltas that, when applied to it, restore it to a previous version of itself), calculating a previous name state required performing a bunch of disk queries.  This wasn't a problem in 0.13 since the entire DB lived in RAM.  But in 0.14, this became a bottleneck--the old `get_nameops_at` RPC call could take a long time to complete and, due to the black magic that powers Python's garbage collector, could sometimes run out of memory.

This method and a few of them below it are used to implement a more scalable version of `get_nameops_at`.  Instead of making a single RPC call, the client does the following:
- get all the "head" name state that corresponds to names that were affected at the given block height (which can be done in batches)
- for each name, get the list of deltas (it's "history", which again can be done in batches)
- merge the name's head and deltas to calculate the name operation at that block height

blockstack_client/proxy.py
```diff
@@ -685,3 +1336,205 @@ def is_name_owner(fqu, address, proxy=None):
         return False
 
     return address == blockchain_record.get('address', '')
+
+
+def get_zonefile_inventory(hostport, bit_offset, bit_count, timeout=30, my_hostport=None, proxy=None):
```
jcnelson, line 1730: This method and the various `_atlas_` methods used to live server-side as well.  I moved them here because (1) it's important to validate their return values (and we had been doing so _manually_ before you showed me `jsonschema`), and (2) it's useful for diagnostics and advanced CLI usage for the client to be able to query its upstream server's Atlas state.  In the future, for example, having these Atlas methods here will make it a lot easier to have the CLI replicate its zonefiles to several Atlas nodes at once, and do more advanced things like try to calculate the entire Atlas peer graph.

blockstack_client/storage.py
```diff
@@ -981,7 +968,7 @@ def blockstack_immutable_data_url_parse(url):
         data_id, blockchain_name, namespace_id, data_hash = m.groups()
         blockchain_id = '{}.{}'.format(blockchain_name, namespace_id)
 
-        if not is_valid_name(blockchain_id):
+        if not is_name_valid(blockchain_id):
             log.debug('Invalid blockstack ID "{}"'.format(blockchain_id))
             raise ValueError('Invalid blockstack ID')
```
jcnelson, line 974: Looks good!  Thanks for taking the time to streamline this :)

muneeb-ali:
@pandu-rao your commit were included and we merged this repo with [blockstack-core](http://github.com/blockstack/blockstack-core). Thanks again for the improvements!

PR #144, opened 2016-10-02
Update actions.py

seriocomedy:


jcnelson:
Thank you for sending us these corrections!  We'll merge them into the next release.

PR #160, opened 2016-11-02
get_names_in_namespace ignores count argument

larrysalibra:
`blockstack get_names_in_namespace 0 10` ignores the `count` argument (in this example 10) and always returns at least 100 names (unless there are less than 100 remaining names).


jcnelson:
Ah, thanks for catching that!

PR #161, opened 2016-11-02
Ensure str value for addresses 

muneeb-ali:
b58check_version_byte() expects a str and I was getting errors in the registrar. Other external calls might run into similar issues. 

@jcnelson please check that this doesn't break something for you.


jcnelson:
Should be fine.  We expect strings anyway.

PR #164, opened 2016-11-17
Invalid Name: More exact error message

bshouse:
Names are not allowed to contain uppercase letters. Explicitly state this in the error message.


jcnelson:
Thanks! Merged :)

PR #165, opened 2016-11-20
minor fix: 1 BTC = 100,000,000 Satoshi

tthsn:
Found this issue after setup a blockstack on bitcoin regtest, then send 6.6 btc to it.
Then run blockstack balance/wallet to show a balance, but the bitcoin value is different.
I guess this issue is due to btc is so expensive that developer don't have so much zeros to display. :)


jcnelson:
Thanks!  Merged :)

PR #181, opened 2016-12-08
Support specify config file in non-test mode

tthsn:
for running another configuration coexistence with the default ~/.blockstack/client.ini

* config file is still a must in test mode
* config file is an optional in non-test mode


jcnelson:
Thanks for this PR!  Will merge in 0.14.1

PR #186, opened 2017-01-11
Fix output for insufficient funds to actually output the needed amount

vbrandl:
The output prints the current balance instead of the fee


jcnelson:
Looks good.  Thanks for catching this!

PR #193, opened 2017-02-12
Change Y/n to y/N for Continue?

leoj3n:
Don't see this problem in your develop branch, but suggesting this to master just to be sure.


muneeb-ali:
@leoj3n  Apologies but you'll need to re-send this PR on the [blockstack-core](http://github.com/blockstack/blockstack-core) repo. We merged the CLI there. Thanks again for sending the PR and we'll merge it, if you resend :-)

PR #198, opened 2017-03-29
Develop core auth & storage

jcnelson:
This PR adds the necessary code to generate Core authentication requests and fetch Core session tokens.


shea256:
@jcnelson Did you commit the folder `tests/coreTests`? I'm not seeing it for some reason in my local code.

shea256:
Awesome thanks @jcnelson! Can you please make a few changes?

1. Can we update use of http to fetch, to keep the requests consistent across the library? From what I can tell, it should allow us to simplify our code a little.
1. Can we get the core auth tests to run automatically without a local core node? I want to do this so we can have the tests running when we push on CircleCI. You can take a look at how @larrysalibra got this to work with the proof tests: https://github.com/blockstack/blockstack.js/blob/master/tests/unitTests/src/unitTestsProofs.js#L14. As you can see, we're using FetchMock to simulate the fetch calls without actually making the external calls. With the functions that make calls to the core API, we could use a sample response and config the response using FetchMock.
1. Once we've gotten all of our tests interacting to run automatically, can we add our test functions to this file: https://github.com/blockstack/blockstack.js/blob/master/tests/unitTests/src/index.js? CircleCI will automatically run `npm run test` and then we'll get the chance to have fully automated test coverage.

shea256:
And then here are just some minor things for styling. Here's some references to go by: https://github.com/airbnb/javascript. And here are just a couple guidelines:

1. Update all uses of `var` in ES6 for `const` if possible and otherwise `let` if the variable is to be changed (let is safer as it's block scoped rather than function scoped).
1. Update all uses of "==" and "!=" for the safer "===" and "!==" (it's annoying, I know).
1. Update the naming to follow the standard javascript camelCase.
1. Replace all uses of `function() {` with `() => {` (safer scoping).
1. Ensure we have spaces before parentheses in control statements (like `if (isJedi) { }`).
1. Investigate whether `sprintf-js` can be replaced by console.log with ES6 template literals.

I can add in a linter to help out with this. Two things I have to start doing are (a) use semicolons (b) fix my multi-line imports and exports to be standardized.

shea256:
We decided to avoid mocking responses (as that would be too complex). Instead we're going to have CircleCI start up a simple core node with only storage enabled and using our integration tests. @jcnelson please correct me if I have any of this wrong.

jcnelson:
> 2. Can we get the core auth tests to run automatically without a local core node? I want to do this so we can have the tests running when we push on CircleCI. You can take a look at how @larrysalibra got this to work with the proof tests: https://github.com/blockstack/blockstack.js/blob/master/tests/unitTests/src/unitTestsProofs.js#L14. As you can see, we're using FetchMock to simulate the fetch calls without actually making the external calls. With the functions that make calls to the core API, we could use a sample response and config the response using FetchMock.
> 3. Once we've gotten all of our tests interacting to run automatically, can we add our test functions to this file: https://github.com/blockstack/blockstack.js/blob/master/tests/unitTests/src/index.js? CircleCI will automatically run npm run test and then we'll get the chance to have fully automated test coverage.

Okay, I've set it up so CircleCI will fetch and install the integration test framework, and run the blockstack.js authentication and storage tests against a live Core node in the CircleCI VM.

This has to be a separate target from `unitTests`, since we have to take a lot of extra steps to get the necessary software up and running.

larrysalibra:
@shea256 
> Two things I have to start doing are (a) use semicolons

Oh man. I stopped using semicolons to match your JS style.

No semi-colons is also my fav feature of python and swift:
img_1723

We returning to the semi-colon dark side?

shea256:
@jcnelson Awesome work! Excited to have this up and running. Now I can easily contribute to the storage code and ensure I'm not breaking anything. Great way to protect the master branch too.

@larrysalibra Good call. We can keep no semicolons if that's what you prefer as well. We can add the linter flag that flips the "unnecessary semicolon" enforcement.

jcnelson:
Okay, I think I've addressed all of the lint issues.  I've also separated out the storage client and put it here for now: https://github.com/jcnelson/blockstack-storage-js

jcnelson:
So, can I merge this PR?

larrysalibra:
@jcnelson i'm going to merge this in the next couple days. sorry about the delay!

jcnelson:
:+1:

PR #219, opened 2017-06-23
authApp.js promise bug fix

scott-martin:
The main thing here is a fix to the way promises were being used in `authApp.js`. I also removed the unnecessary callback parameter in `signUserIn` and `loadUserData`. Finally, added a `toJSON` method to the Person class to make it more pleasant to use with redux.


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "blockstack",
   "version": "0.3.10",
   "description": "The Blockstack Javascript library for identity and authentication.",
-  "main": "lib/index",
+  "main": "dist/blockstack.js",
```
larrysalibra, line 5: `dist/blockstack.js` is the browserified version intended for apps using this package in the browser in an app that doesn't have its own browserify/webpack system. The previous entry point, `lib/index` allows package users to use this on node or in the browser via their preferred browserify/webpack system. Does this make sense? Is there a better way we can achieve this?

larrysalibra:
This is really great @scott-martin ! Thank you so much! Taking a look now and adding some comments.

shea256:
@scott-martin Really appreciate your contributions here. Awesome to be working with you.

package.json
scott-martin: I needed to be able to specify my own fork of the library in my package.json. The only way I could get that to work was by changing this field. I was using a webpack bundler myself, and it didn't complain about giving it an already packaged file. I'd have to look deeper to know for sure. Probably fine to just leave this off.

scott-martin:
Hey @larrysalibra noticed you closed this. Will you be merging in the fixes in authApp at least? Without those I'm kind of stuck with staying on my own fork. Any changes you need from me to resubmit?

larrysalibra:
@scott-martin I don't recall closing it. Could have a been a fat finger on my part. Apologies if so!

larrysalibra:
@scott-martin Oh! I see what happened..apparently github automatically closes a pull request if you delete a branch:
screen shot 2017-06-29 at 10 12 26 pm

scott-martin:
Oh ok, I didn't know that either. No worries. The only parts I really need are the changes to `signUserIn` and `loadUserData`. The `signUserIn` change fixes an actual bug. And the `loadUserData` change gets rid of a callback that's really awkward to use. Appreciate the quick response!

larrysalibra:
Thank you @scott-martin ! We're working out some issues with the branch I merged this into regarding storage. Once those are figured out (hopefully today) I'll push a new build to npm!

Issue #251, opened 2017-08-29
Add additional service verification support


larrysalibra:
Currently we support facebook, twitter and github.

Here's a list of all the service types we want to eventually support:

![29775517-6d7acffa-8bd3-11e7-9fb5-81ba34772014](https://user-images.githubusercontent.com/597182/29836820-5a8f1454-8cc4-11e7-8dd6-c143bfce9c78.png)

We don't have to tackle these all at once. 

Let's add a few high priority ones in this sprint and then open issues for others and mark them with the label `help-wanted`.

guylepage3:
Twitter
Facebook
GitHub
Instagram
LinkedIn
Tumblr
Reddit
Pinterest
YouTube
AngelList
StackOverflow
Hacker News
Snapchat

shea256:
Also considering:

- Yelp
- Foursquare
- SoundCloud
- GoodReads

Let's prioritize the following:

- Instagram
- Hacker News
- LinkedIn

larrysalibra:
We can break these out out into one issue per service: these issues would be great for community members to take stab at. Documenting the interface for services would help as well.

larrysalibra:
If you'd like to add an additional service, please create a new issue for that service and paste a link to the issue into a comment here. We'll happy accept pull requests. Let us know if you have any questions!

Issue #250, opened 2017-08-29
Implement new address based verifications system


larrysalibra:
We're moving from a name-based verification system to an address based verification system that will allow users to verify control over their social media accounts by the same person that controls the private key of the address that owns their avatar.

This is the format of verification message:
`Verifying my avatar on Blockstack is owned by the address 19MoWG8u88L6t766j7Vne21Mg4wHsCQ7vk https://avatar.blockstack.org/address/19MoWG8u88L6t766j7Vne21Mg4wHsCQ7vk`

yknl:
Implemented in 8b4bedcfea95f97089d86a18189637f95ce7f24f

larrysalibra:
@yknl 

We're going to tweak the verification message. See https://github.com/blockstack/blockstack-browser/issues/765 for discussion & reasoning.

New format
```
Verifying my blockchain ID is secured with the address 19MoWG8u88L6t766j7Vne21Mg4wHsCQ7vk

https://id.blockstack.org/address/19MoWG8u88L6t766j7Vne21Mg4wHsCQ7vk
```

larrysalibra:
This has already shipped .

shea256:
@larrysalibra Can we please confirm that the verification message is the following?

```
Verifying my Blockstack ID is secured with the address 19MoWG8u88L6t766j7Vne21Mg4wHsCQ7vk
https://explorer.blockstack.org/address/19MoWG8u88L6t766j7Vne21Mg4wHsCQ7vk
```

shea256:
Actually I see #263 is open. I'll comment on there.

PR #253, opened 2017-09-06
Address based verifications + Additional social verification

yknl:
#250 #251 
- **Implemented new address based verifications**
- **Fixed proof URL check bugs**
- **Refactored verification code**
- **Add support for Instagram account verification**
 Verifies username and verification message based on post meta tag content
- **Add support for Hacker News account verification**
 Checks for the verification message in user profile "about" section
- **Add support for LinkedIn account verification**
 Verifies username and verification message in post


larrysalibra:
@yknl I tried this out this morning with the browser and address-based verifications work really well! 

A few things I like love your thoughts on:

1. We are adding support for address verifications while not breaking existing name based verifications. At some point in the future we may want to phase out name-based verifications because they're not as secure (it's not clear which blockstack universe the names exist on).

To achieve this with our current `validateProofs` interface, we'd have to call `validateProofs` twice, once with the owner address as the `identifier` and once with the name as the `identifier` and then merge the results.

We should try to hide this complexity from the user so they only have to run `validateProofs` once for a given profile.

What if we changed the interface to something like this:

`validateProofs(profile, ownerAddress, name = null)` ?

...and changed the function to run both checks name and address based verifications if a name is provided?

In the future, when we decide to start to phase out name-based proofs, we could add an additional argument

`validateProofs(profile, ownerAddress, name = null, enableLegacyProofs = false)` so that users have to explicitly opt-in to continue to use name-based proofs.

We could validate that `ownerAddress` address is a valid bitcoin address and raise an error otherwise so that users passing a name as the second argument would know that the interface has changed.

2. Something I forgot to mention before: We've been slowly adding [Flow static type checking](https://flow.org) to this library and more recently the browser. As part of our testing suite, the flow type checker and linter both run (and must pass) before the other automated tests are run.
Each file opts in by adding `// @flow` or `/* @flow */ to the top of the file. I've been adding it as I go. (like whenever I happen to work on a file) Can you take a stab at adding it to the proofs-related code? 

3. Another thing we've been trying to improve in this library is documentation. We're using [documentationjs](http://documentation.js.org) to do this. They support jsdoc style comments and also support Flow type annotations!

Going forward, we're trying to document functions that should be part of our public API. You can see an example here: https://github.com/blockstack/blockstack.js/blob/master/src/auth/authApp.js#L67 `validateProofs` is a function we should add to our documented public API. Can you take a stab at this too? 

Feel free to document any other functions you feel make sense as well - marking them `@private` will make sure they don't appear in the public API docs: http://blockstack.github.io/blockstack.js/index.html

Let me know if this stuff makes sense or if we should take a different approach!

Really can't wait to get this integrated into our products! Thanks for taking this on @yknl !

yknl:
@larrysalibra 

1. Sounds good to me. We will be introducing a breaking API change though, so we'll need to make the corresponding change in the Blockstack browser when updating to this version of Blockstack.js.

Another approach that we can take is to just move name-based verifications to `validateNameBasedProofs()`. This will still add complexity to user code for now, but will hopefully encourage users to switch entirely to address-based verifications. (Since we will eventually phase out name-based verifications anyway) This will also leave us with a cleaner interface for the address-based `validateProofs()` and mitigate a possible breaking change in the future when name-based verifications are phased out. Let me know your thoughts on this.

Regarding 2 & 3:
I will look into flow static type checking and documentation for this issue as well as going forward.

larrysalibra:
@yknl 

> This will still add complexity to user code for now, but will hopefully encourage users to switch entirely to address-based verifications. (Since we will eventually phase out name-based verifications anyway) 

We'll need to implement checking both address and legacy proofs in browser, explorer and our resolver. Writing it once in this library instead of multiple times in user code make it easier to make sure we're getting the same behavior across the ecosystem.

If we hide the complexity in this library, we can deprecate legacy name-based proof checking in the future by making the changes here and updating `blockstack.js` in a non-breaking way. If we rely on user code for this, then we also have to get app devs to update their code as well in addition to installing a new version of `blockstack.js`

In general, we should try to hide as much complexity from users of our library as possible. Do this make sense?

yknl:
@larrysalibra Makes sense, we'll take your approach and hide away the complexity.

larrysalibra:
@yknl one more tweak of the verification message text: https://github.com/blockstack/blockstack.js/issues/250#issuecomment-329478555

will review the changes you pushed yesterday in a bit! thanks for the quick turn around!

larrysalibra:
Looks great @yknl ! Thank you!

PR #258, opened 2017-09-21
Fixes mismatched case error

kantai:
fix mismatched casing in proof check when there's mixed casing in an identifier --

example:

https://core.blockstack.org/v2/users/ablankstein.id

```
...
"account": [
        {
          "@type": "Account", 
          "identifier": "AaronBlankstein", 
          "proofType": "http", 
          "proofUrl": "https://twitter.com/AaronBlankstein/status/849369534968496129", 
          "service": "twitter"
        },
...
```


larrysalibra:
Thanks @kantai - will include this in the next release.

PR #260, opened 2017-09-28
Add keyfile support to blockstack.js

kantai:
This is required for multi-reader storage support, will depend on a new version of blockstack-storage-js.


kantai:
Closing this, as this is no longer how we plan to implement multi-player storage in `blockstack.js`

PR #262, opened 2017-10-03
Add ECIES encryption to putFile/getFile

kantai:
Adds support for using the AppPrivateKey to perform encryption and decryption of data stored using putFile/getFile.

This adds `encrypt` and `decrypt` boolean parameters to `putFile` and `getFile`, respectively. These parameters default to false.

This is an implementation of ECIES (see [wiki](https://en.wikipedia.org/wiki/Integrated_Encryption_Scheme)), which uses the `crypto` and `elliptic` libraries -- `elliptic` was already an explicit dependency, and `crypto` was an implicit dependency (though one that was already getting used explicity for getRandomBytes), so this shouldn't change our library dependency tree.

### DIFF
--- a/dist/blockstack.js
+++ b/dist/blockstack.js
@@ -943,25 +943,17 @@ function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { de
 var ecurve = new _elliptic.ec('secp256k1');
 
 function aes256CbcEncrypt(iv, key, plaintext) {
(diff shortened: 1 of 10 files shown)


larrysalibra:
Awesome @kantai ! Can't wait to try this out!

larrysalibra:
beautiful @kantai

screen shot 2017-10-10 at 5 29 27 am

PR #274, opened 2017-10-23
Allow encryption/decryption without putFile/getFile

nivas8292:


src/storage/index.js
```diff
@@ -26,6 +26,19 @@ export function getFile(path: string, decrypt: boolean = false) {
     })
 }
 
+/**
+ * Encrypts the data provided.
+ * @param {String|Buffer} content - the data to store in the file
+ * @return {Promise} that resolves if the operation succeed and rejects
```
larrysalibra, line 32: This function doesn't return a Promise.

larrysalibra (changes requested):
This is really great @nivas8292 - thank you so much for it!

Can you take a stab at adding some test coverage for this?

Issue #276, opened 2017-10-24
Encrypt app private key in authentication response using app transit key


larrysalibra:

kantai:
Closing this out, as this was implemented and merged back in October.

PR #278, opened 2017-10-25
Use transit key for encrypting auth request secrets

kantai:
This addresses issue #276 

More discussion over at `blockstack-browser`:

https://github.com/blockstack/blockstack-browser/issues/958
https://github.com/blockstack/blockstack-browser/pull/960


dominictarr:
👍 this encrypts just the app private key field, which means without the app transit key, it is not possible to decrypt the private key. Theirfore, when the app makes requests to the core server, it will not leak the private key, so the JWT authorization is still safe to pass around.

PR #286, opened 2017-11-03
Revert "use crypto randomBytes instead of math.random"

larrysalibra:
Reverts blockstack/blockstack.js#267


larrysalibra:
@kantai  
This was causing 
```
  var payload = {
    jti: (0, _utils.makeUUID4)(),
```
to fail with `RangeError: offset is not uint`

PR #299, opened 2017-11-21
Fix Invalid-Hex-String

### LINKED ISSUE #296: Using `putFile` on the same file repeatedly sometimes results in a `TypeError`
When using `putFile` with the encrypt parameter set to `true` a `TypeError: Invalid hex string` error is sometimes thrown. The issue does not happen when encrypt is `false`

I've found the bug to be more consistent when writing to the same file multiple times in a row. In my case, I am using `putFile` every time user preferences are changed for it to be auto-saving (rather than edit-save)

### Stack Trace

stacktrace1
stacktrace2

kantai:
Pretty fun bug: Roughly 1/16th of encrypt calls would generate a EICS shared-secret key with a leading '0' in hex notation. The 'elliptic' library uses 'bn.js' for BigNums, and the `toString('hex')` method of that class does _not_ add a leading '0' -- this results in a 63-character rather than a 64-character hex string, which `Buffer` will complain about.

Using the `toBuffer()` method of the BN object gets around this problem.

This should fix #296 and also the intermittent CircleCI test failures.


larrysalibra:
Thanks @kantai !

PR #301, opened 2017-11-22
Fix #296 #299 Fix Invalid Hex String Error when using `encryptECIES`

### LINKED ISSUE #296: Using `putFile` on the same file repeatedly sometimes results in a `TypeError`

brandonparee:
This issue was caused when the length of `sharedSecret.toString('hex')` was not 64.

To work around this issue a new recursive function `getEncryptionKeys` was created that will return an object with `sharedSecret` and `ephemeralPK` to be used by `encryptECIES`. The function will only return a value if the converted hex string for `sharedSecret` is 64 characters long. The function will run again otherwise.

### DIFF
--- a/dist/blockstack.js
+++ b/dist/blockstack.js
@@ -254,7 +254,7 @@ function signUserOut() {
     window.location = redirectURL;
   }
 }
-},{"../index":11,"../profiles":13,"../utils":35,"./authConstants":2,"./authMessages":3,"./index":7,"custom-protocol-detection-blockstack":298,"jsontokens":381,"query-string":478}],2:[function(require,module,exports){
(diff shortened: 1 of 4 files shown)


larrysalibra:
Thanks so much for this @brandonparee - @kantai or I will take a look at this in the next couple days!

kantai (commented):
Thanks for the PR!

I think this is the right approach, the BN.toBuffer method isn't reliable. However, the 63 character hexstring is due to dropping a leading 0 in the representation of the buffer, so we can just return an appended '0' instead, so I'll merge this into a branch on our repo and add some changes.

PR #302, opened 2017-11-23
Invalid Hex String in Encryption

kantai:
This integrates @brandonparee's fixes for the invalid hex string, and then modifies it to fix broken 63 character hex strings, rather than recursing and generating another ephemeral key. Tested this on a local blockstack-todos, and it looks like it fixes the original issue, and then the TypeErrors coming from the use of BN.toBuffer

### DIFF
--- a/dist/blockstack.js
+++ b/dist/blockstack.js
@@ -254,7 +254,7 @@ function signUserOut() {
     window.location = redirectURL;
   }
 }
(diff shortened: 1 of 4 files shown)


larrysalibra:
Thanks @kantai - I'll test this out tomorrow too and make a new release assuming there are no problems.

larrysalibra:
Thanks again for this @brandonparee & @kantai - this tested out for me as well locally. I've pushed a new build v0.13.7 that incorporates the fixes and updated blockstack-todos to use it.

Issue #311, opened 2017-12-11
Implement direct-to-hub gaia hub storage client


larrysalibra:
Storage communication will be directly between apps & Gaia Hubs without passing through an intermediary local Blockstock Core daemon.

Related issues:
Bearer token generation code: https://github.com/blockstack/blockstack.js/issues/310
Pass gaia hub root url in authentication response token: https://github.com/blockstack/blockstack.js/issues/309

brandonparee:
Am I correct in assuming this will solve https://github.com/blockstack/blockstack-core/issues/690?

Also if needed, I can help test this out in a live environment once a PR is brought in.

kantai:
This has been merged and released as 0.14.0 on `npm`

Issue #310, opened 2017-12-11
Move gaia hub bearer token generation code from browser into blockstack.js


larrysalibra:

larrysalibra:
This shipped! Thanks @kantai !

Issue #309, opened 2017-12-11
Pass gaia hub root url as part of generating an authentication response


larrysalibra:

larrysalibra:
@larrysalibra will do acceptance testing on this issue.

larrysalibra:
This shipped.

PR #314, opened 2017-12-13
Direct gaia hub storage operations

kantai:
This PR implements:

#309 Pass gaia hub root url as part of generating an authentication response
#310 Move gaia hub bearer token generation code from browser into blockstack.js  
#311 Implement direct-to-hub gaia hub storage client

In general, the way it works, is:

1. it bumps the version on the Auth Request
2. tries to read the hubURL from the Auth Response. if it cannot, it defaults to `hub.blockstack.org`
3. stores a user's hubConfig to localStorage

This PR includes a _new_ interface as well, that people may want to advise on. Now that we have a bumped version for the auth request, we need to actually do semantic version checks on those requests. To do this, I added a `isLaterVersion` function to `src/utils.js`

Testing this doesn't require an updated browser (if you use an old browser, the sign in process in the browser will still do the datastore setup and core session token generation even though the app won't use either)

Improvements this should give us:
1. Faster and more reliable writes/reads (and the errors should be more understandable)
2. URLs which match files (i.e., if you putFile(todos.json), the stored file is a `gaia.blockstack.org/hub/<app-addr>/todos.json`
3. Ability to serve applications which use storage over HTTPS (no longer making requests to localhost means no longer having mixed-content errors).

Once the blockstack-browser updates are live, we'll also be able to use storage from the next Windows native build and webapp builds.

### DIFF
--- a/dist/blockstack.js
+++ b/dist/blockstack.js
@@ -201,7 +201,7 @@ function handlePendingSignIn() {
             }
           }
         }
-        var hubUrl = 'https://hub.blockstack.org';
+        var hubUrl = _authConstants.BLOCKSTACK_DEFAULT_GAIA_HUB_URL;
         if ((0, _utils.isLaterVersion)(tokenPayload.version, '1.2.0') && tokenPayload.hubUrl !== null && tokenPayload.hubUrl !== undefined) {
           hubUrl = tokenPayload.hubUrl;
         }
@@ -280,6 +280,7 @@ var BLOCKSTACK_STORAGE_LABEL = exports.BLOCKSTACK_STORAGE_LABEL = 'blockstack';
 var DEFAULT_BLOCKSTACK_HOST = exports.DEFAULT_BLOCKSTACK_HOST = 'https://blockstack.org/auth';
 var DEFAULT_SCOPE = exports.DEFAULT_SCOPE = ['store_write'];
 var BLOCKSTACK_APP_PRIVATE_KEY_LABEL = exports.BLOCKSTACK_APP_PRIVATE_KEY_LABEL = 'blockstack-transit-private-key';
+var BLOCKSTACK_DEFAULT_GAIA_HUB_URL = exports.BLOCKSTACK_DEFAULT_GAIA_HUB_URL = 'https://hub.blockstack.org';
 },{}],3:[function(require,module,exports){
 (function (Buffer){
 'use strict';
@@ -3451,6 +3452,8 @@ var _bigi2 = _interopRequireDefault(_bigi);
 
 var _authApp = require('../auth/authApp');
 
+var _authConstants = require('../auth/authConstants');
+
 function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
 
 var BLOCKSTACK_GAIA_HUB_LABEL = exports.BLOCKSTACK_GAIA_HUB_LABEL = 'blockstack-gaia-hub-config';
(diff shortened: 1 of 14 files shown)


jackzampolin:
fixes https://github.com/blockstack/blockstack-browser/issues/631

larrysalibra:
@kantai this appears to work for me on locally on todos. And it's much faster! Awesome work!

What do the errors about not being able to decrypt the `appPrivateKey` mean?

(I'm using a browser that hasn't been upgraded)

screen shot 2017-12-14 at 10 39 24 pm

kantai:
`appPrivateKey` errors occur because on the browser authRequest version check (https://github.com/blockstack/blockstack-browser/blob/master/app/js/auth/components/AuthModal.js#L220), version is checked with `===`, not a semantic version check, so the AuthResponse doesn't try to encrypt the appPrivateKey, which leads to those two 'failed decryption' messages.

That's being fixed in the browser branch I'm working on -- https://github.com/blockstack/blockstack-browser/tree/feature/direct-gaia-app-support -- it uses `isLaterVersion` from `blockstack.js` to perform the check

larrysalibra:
> version is checked with ===, not a semantic version check, so the AuthResponse doesn't try to encrypt the appPrivateKey, which leads to those two 'failed decryption' messages.

Does this mean if app devs update blockstack.js and users keep using an older version of the browser, their app private key will be potentially exposed (eg. in referrer headers)?

kantai:
> Does this mean if app devs update blockstack.js and users keep using an older version of the browser, their app private key will be potentially exposed (eg. in referrer headers)?

Yes -- 

As alternative to this is to hold off on updating the version of the authRequest message and instead add a field (which would be a non-breaking change) that would allow a new browser version (next release 0.21.5) to determine that the app is capable of communicating with gaia hubs directly without performing a version check. We would also update the version check in browser to perform a release semantic versioning check. Then in future releases, we'd be able to use better version checks.

larrysalibra:
> As alternative to this is to hold off on updating the version of the authRequest message and instead add a field (which would be a non-breaking change) that would allow a new browser version (next release 0.21.5) to determine that the app is capable of communicating with gaia hubs directly without performing a version check.

I think this approach makes sense. We don't want to leave users in a position where they're exposing private keys.

larrysalibra (changes requested):
@kantai this works really great for me especially with the fixes you pushed.

A question: can we remove blockstack-storage as a dependency with this pull request?

My two concerns are:

1. That we're potentially leaking unencrypted private keys of users who use the current browser when an app upgrades to this. (previously discussed in an earlier comment)

2. This doesn't have any test coverage. (unless i'm missing it)

Do you think we can make changes to address both of these and still support our time frame?

kantai:
@larrysalibra 

Yes -- both of these can be addressed pretty quickly. I'll add the test cases and the authRequest changes tomorrow.

brandonparee:
While testing this PR I noticed that if a user was previously logged into an application they would need to log out and back into the application in order to properly set `hubUrl` in `userData`. 

To work around this I've put in a PR (#317) that will make sure that `hubUrl` is set before calling `connectToGaiaHub`.

kantai:
Changes in the current version:

1. Unit tests for the gaia hub management functions (the higher level functions aren't tested in the unit cases, as they read config settings from local storage) 
2. 404s (file doesn't exist) returns `null` rather than raising exception
3. Version un-bumped -- now old versions of the blockstack browser will continue to encrypt the app private key
4. Incorporates @brandonparee's PR
5. Removes `blockstack-storage` dependency

larrysalibra (commented):
Looks great! Thanks!

Issue #308, opened 2017-12-02
Meteor integration not compatible with crypto package


raulk:
We are creating an `accounts-blockstack` Meteor module to integrate Blockstack login with Meteor's account system. Unfortunately, the usage of the `crypto` module (which is now present in vanilla Node) is incompatible and causes an error.

I deleted it from `npm_modules` manually, and it works properly. Please migrate away from this package.

Error:

```
Uncaught Error: Cannot find module 'crypto'
    at makeMissingError (modules-runtime.js?hash=5e485d3e2a49d2506f7ca0df72fcd6a3216a83a5:241)
    at require (modules-runtime.js?hash=5e485d3e2a49d2506f7ca0df72fcd6a3216a83a5:251)
    at secp256k1.js (modules.js?hash=dc5b20b05aad051d7248e15645099d47399039ba:5289)
    at fileEvaluate (modules-runtime.js?hash=5e485d3e2a49d2506f7ca0df72fcd6a3216a83a5:353)
    at require (modules-runtime.js?hash=5e485d3e2a49d2506f7ca0df72fcd6a3216a83a5:248)
    at index.js (modules.js?hash=dc5b20b05aad051d7248e15645099d47399039ba:5260)
    at fileEvaluate (modules-runtime.js?hash=5e485d3e2a49d2506f7ca0df72fcd6a3216a83a5:353)
    at require (modules-runtime.js?hash=5e485d3e2a49d2506f7ca0df72fcd6a3216a83a5:248)
    at signer.js (modules.js?hash=dc5b20b05aad051d7248e15645099d47399039ba:5169)
    at fileEvaluate (modules-runtime.js?hash=5e485d3e2a49d2506f7ca0df72fcd6a3216a83a5:353)
```

raulk:
Usages of crypto are:

```
.../blockstack.js/src/encryption.js:
    7:   const cipher = crypto.createCipheriv('aes-256-cbc', key, iv)
   12:   const cipher = crypto.createDecipheriv('aes-256-cbc', key, iv)
   17:   return crypto.createHmac('sha256', key).update(content).digest()
   33:   const hashedSecret = crypto.createHash('sha512').update(sharedSecret).digest()
   71:   const initializationVector = crypto.randomBytes(16)
```

All these functions are present in the standard Node crypto package: https://nodejs.org/api/crypto.html

raulk:
Are you perhaps keeping it around for the browser distribution?

kantai:
[...]

Issue #305, opened 2017-11-30
crypto is deprecated


timsu:
Blockstack's package.json currently depends on "crypto", which npm indicates is deprecated: https://www.npmjs.com/package/crypto

kantai:
Thanks for raising this issue @timsu --

I'm testing out simply removing the dependency now (as it's included in nodeJS by default)

PR #315, opened 2017-12-14
Remove deprecated crypto

kantai:
Per issue #305 and #308, we shouldn't be listing the `crypto` library in our package.json anymore.

This is included by default in nodejs now, and so simply deleting it from the package.json works -- even with webpack.


cwackerfuss-rl:
Thank you!!

PR #324, opened 2017-12-19
Multi-reader storage using app storage buckets

yknl:
We now have a functional API for multi-reader storage via app index files which looks like this
```function getUserAppIndex(name: string, appOrigin: string, profileLookupURL: string = 'https://core.blockstack.org/v2/users/')```

Returns the app index file for a given user and app by looking up the user profile on Blockstack Core and then fetching the appropriate app index file.

We're using Blockstack ID to look up user profiles here. We could potentially allow look ups via owner address, but how people are going to find other people's address is a problem.

Effectively, a Blockstack ID is required for multi-reader storage. When free subdomains are enabled, this shouldn't be a big issue.

There are also 2 new methods to write/get the current user's app index file: `getAppIndexFile` and `putAppIndexFile`, which essentially wraps the `putFile` and `getFile` methods.

### DIFF
--- a/src/profiles/profileLookup.js
+++ b/src/profiles/profileLookup.js
@@ -4,17 +4,17 @@ import { resolveZoneFileToProfile } from './profileZoneFiles'
 /**
  * Look up a user profile by blockstack ID
  *
- * @param {string} name The Blockstack ID of the profile to look up
+ * @param {string} username The Blockstack ID of the profile to look up
  * @param {string} [zoneFileLookupURL=http://localhost:6270/v1/names/] The URL
  * to use for zonefile lookup 
  * @returns {Promise} that resolves to a profile object
  */
-export function lookupProfile(name: string, zoneFileLookupURL: string = 'http://localhost:6270/v1/names/') {
+export function lookupProfile(username: string, zoneFileLookupURL: string = 'http://localhost:6270/v1/names/') {
   return new Promise((resolve, reject) => {
-    if (!name) {
+    if (!username) {
       reject()
     }
-    const url = `${zoneFileLookupURL.replace(/\/$/, '')}/${name}`
+    const url = `${zoneFileLookupURL.replace(/\/$/, '')}/${username}`
     try {
       fetch(url)
         .then(response => response.text())
--- a/src/profiles/profileLookup.js
+++ b/src/profiles/profileLookup.js
@@ -4,7 +4,7 @@ import { resolveZoneFileToProfile } from './profileZoneFiles'
 /**
  * Look up a user profile by blockstack ID
  *
- * @param {Object} name The Blockstack ID of the profile to look up
+ * @param {string} name The Blockstack ID of the profile to look up
  * @param {string} [zoneFileLookupURL=http://localhost:6270/v1/names/] The URL
  * to use for zonefile lookup 
  * @returns {Promise} that resolves to a profile object
--- a/src/profiles/profileLookup.js
+++ b/src/profiles/profileLookup.js
@@ -1,35 +1,37 @@
 /* @flow */
+import { resolveZoneFileToProfile } from './profileZoneFiles'
 
 /**
  * Look up a user profile by blockstack ID
  *
  * @param {Object} name The Blockstack ID of the profile to look up
- * @param {string} [profileLookupURL=https://core.blockstack.org/v2/users/] The URL
- * to use for profile lookup 
+ * @param {string} [zoneFileLookupURL=http://localhost:6270/v1/names/] The URL
+ * to use for zonefile lookup 
  * @returns {Promise} that resolves to a profile object
  */
-export function lookupProfile(name: string, profileLookupURL: string = 'https://core.blockstack.org/v2/users/') {
+export function lookupProfile(name: string, zoneFileLookupURL: string = 'http://localhost:6270/v1/names/') {
   return new Promise((resolve, reject) => {
     if (!name) {
       reject()
     }
-    const url = `${profileLookupURL.replace(/\/$/, '')}/${name}`
+    const url = `${zoneFileLookupURL.replace(/\/$/, '')}/${name}`
     try {
       fetch(url)
         .then(response => response.text())
         .then(responseText => JSON.parse(responseText))
         .then(responseJSON => {
-          if (responseJSON.hasOwnProperty(name)) {
-            resolve(responseJSON[name].profile)
+          if (responseJSON.hasOwnProperty('zonefile')
+            && responseJSON.hasOwnProperty('address')) {
+            resolve(resolveZoneFileToProfile(responseJSON.zonefile, responseJSON.address))
           } else {
             reject()
           }
         })
-        .catch(() => {
-          reject()
+        .catch((e) => {
+          reject(e)
         })
     } catch (e) {
-      reject()
+      reject(e)
     }
   })
 }
--- a/src/profiles/profileLookup.js
+++ b/src/profiles/profileLookup.js
@@ -1,0 +1,35 @@
+/* @flow */
+
+/**
+ * Look up a user profile by blockstack ID
+ *
+ * @param {Object} name The Blockstack ID of the profile to look up
+ * @param {string} [profileLookupURL=https://core.blockstack.org/v2/users/] The URL
+ * to use for profile lookup 
+ * @returns {Promise} that resolves to a profile object
+ */
+export function lookupProfile(name: string, profileLookupURL: string = 'https://core.blockstack.org/v2/users/') {
+  return new Promise((resolve, reject) => {
+    if (!name) {
+      reject()
+    }
+    const url = `${profileLookupURL.replace(/\/$/, '')}/${name}`
+    try {
+      fetch(url)
+        .then(response => response.text())
+        .then(responseText => JSON.parse(responseText))
+        .then(responseJSON => {
+          if (responseJSON.hasOwnProperty(name)) {
+            resolve(responseJSON[name].profile)
+          } else {
+            reject()
+          }
+        })
+        .catch(() => {
+          reject()
+        })
+    } catch (e) {
+      reject()
+    }
+  })
+}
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: src/profiles/profileLookup.js [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Look up a user profile by blockstack ID
+ *
+ * @param {Object} name The Blockstack ID of the profile to look up
+ * @param {string} [profileLookupURL=https://core.blockstack.org/v2/users/] The URL
+ * to use for profile lookup 
+ * @returns {Promise} that resolves to a profile object
+ */
+export function lookupProfile(name: string, profileLookupURL: string = 'https://core.blockstack.org/v2/users/') {
```
larrysalibra: Resolving profiles should be done in javascript like in the browser.

We can move this browser logic to blockstack.js to achieve this:
https://github.com/blockstack/blockstack-browser/blob/master/app/js/utils/profile-utils.js
https://github.com/blockstack/blockstack-browser/blob/master/app/js/utils/zone-utils.js#L36

One reason this needs to happen is that local core api endpoints don't have the `/v2/users` endpoint (unless i'm mistaken) - I believe this is only resolver feature. 

We'd want to use this end point:

```
curl http://localhost:6270/v1/names/user1.id
{"status": "registered", "zonefile": "$ORIGIN user1.id\n$TTL 3600\n_http._tcp URI 10 1 \"https://gaia.blockstack.org/hub/1FmW15z7yj6t2eiWCjHXq34aAUKBsEF4AS/0/profile.json\"\n", "expire_block": 55245, "blockchain": "bitcoin", "last_txid": "04662e7f514d2e23fa782045abe7a60f01a0cb4cf18b943a9730069681d4cc6e", "address": "1FmW15z7yj6t2eiWCjHXq34aAUKBsEF4AS", "zonefile_hash": "f58c767473e9291ea654965d55b12f1591d11550"}
```

larrysalibra (changes requested):
This is really great Ken! I tried it out with a slightly modified todos list app and the app index got added to my profile file perfectly.

One thing we definitely need those is test suite coverage for this.

I think we also have room to improve the API and naming of this feature. Will give some thought to that tomorrow.

src/profiles/profileLookup.js
kantai: Yep, that's correct -- you only get the `/v2/users/` endpoint on the resolver.

yknl:
I've updated the code to use the `/v1/names/` endpoint for profile lookups. And will add test coverage to the new methods.

What do you guys think about abstracting away the app index file with a `putPublicFile` method. This will do 2 things in one call:
1. Store the file 
2. Add the url of the file to `app_index.json` so it looks like this:
```
{
  files: {
    myPublicData: 'https://gaia.blockstack.org/hub/1fhUqfN3JYNt121wuB4Z2fPC7k4Jn32N1R/data.json'
  },
  // ...other fields
}
```

Developers who need to customize the structure of the `app_index.json` file can still do so using the `putAppIndexFile` method.

Another thing I want to propose is to modify the `putFile` method so that you can specify the public key used for encryption. This should enable private messaging applications where data can be encrypted and read by a specific user. Let me know your thoughts.

kantai:
> What do you guys think about abstracting away the app index file with a putPublicFile method. This will do 2 things in one call:
>
>    Store the file
>    Add the url of the file to app_index.json so it looks like this:

I think that this is nice to provide, as this is a use case for many applications. However, the counter-argument I see is that we should start with sort of the minimum API set required to get something like multi-reader storage, and not make any assumptions yet about how it'd be used. For, example, for peer-to-peer messaging the index.json would only be used to set up the peering, rather than describing files. This change would also not abstract away the index.json completely, because the application would still need to know what to do with `getUserAppIndex`.

Maybe the right thing to do is to say that we should see how demo apps look first, and then figure out if there's common patterns we can clean up via API.

> Another thing I want to propose is to modify the putFile method so that you can specify the public key used for encryption. This should enable private messaging applications where data can be encrypted and read by a specific user. Let me know your thoughts.

I like this idea -- I'll open an issue for it and we can discuss it over there.

tutley:
One comment to add here - you might also want to think of how people might use this index file in relation to other libraries that are out there. So for example, if you just give people the ability to create and update an index file for their app and structure it however they want to, they may end up using a single-file database library to interact with the file, such as SQL lite or https://github.com/hamiltop/MongoLiteDB

src/storage/index.js
```diff
@@ -1,22 +1,55 @@
...
+    app: null,
+    zoneFileLookupURL: 'http://localhost:6270/v1/names/'
+  }
+
+  const opt = Object.assign({}, defaults, options)
+
   return getOrSetLocalGaiaHubConnection()
-    .then((gaiaHubConfig) => fetch(getFullReadUrl(path, gaiaHubConfig)))
+    .then((gaiaHubConfig) => {
+      if (opt.user && opt.app) {
```
larrysalibra: If `options.user` is specified and `options.app` isn't provided, this function should assume the same app. This ways apps can read files in other apps by only providing a username.

larrysalibra (changes requested):
@yknl This generally works well for me! great job on this!

I ran into one issue I think we should address before merging this that happens with multiple public `putFile` requests:

If you call a number of successive multi-player `putFile`, these execute in parallel and the app index file won't always include all of the files.

In the example below, I write 4 files that are intended to be public, but the end only the last one shows up in the index.

```
	let options = {
          public: true
        }

        blockstack.putFile('a.json', JSON.stringify({'test': 'a'}), options)
        blockstack.putFile('b.json', JSON.stringify({'test': 'b'}), options)
        blockstack.putFile('c.json', JSON.stringify({'test': 'c'}), options)
        blockstack.putFile('d.json', JSON.stringify({'test': 'd'}), options)
```


```
curl https://gaia.blockstack.org/hub/1AUGZZDHVVffDSNn47vBXjoimKixDEpFJ9/app_index.json
{"files":{"todos.json":"https://gaia.blockstack.org/hub/1AUGZZDHVVffDSNn47vBXjoimKixDEpFJ9/todos.json","d.json":"https://gaia.blockstack.org/hub/1AUGZZDHVVffDSNn47vBXjoimKixDEpFJ9/d.json"}}
```

What if instead storing an app index file, we just stored the app's gaia hub storage bucket base url. So instead of `https://gaia.blockstack.org/hub/1AUGZZDHVVffDSNn47vBXjoimKixDEpFJ9/app_index.json`, we store `https://gaia.blockstack.org/hub/1AUGZZDHVVffDSNn47vBXjoimKixDEpFJ9/` in the user's profile as the storage location for that app. 

Not needing to update an index file would avoid these concurrency issues, and also save us an http request. We could also simplify our `putFile` API by removing the need for `public` since everything is public unless you encrypt it.

This would also let us remove the concept of an app index and trying to figure out how to name it.

Thinking about this from the user's perspective, what an app as doing when it is asking for this permission is to publicize your use of this particular app. Making unencrypted files in known locations discoverable is a side effect of this.  (we should think about this in respect to the language of the authentication dialog)

---


Getting a bit late here, but I will have additional feedback tomorrow.

yknl:
Thanks @larrysalibra, this is great feedback! I've removed the need for an app index file and it solved the concurrency issue, simplified the API and improved performance significantly during testing.

src/storage/README.md
```diff
@@ -47,9 +51,30 @@ version. These features will be rolled out in future updates._
...
 ```
+
+## Reading another user's unencrypted file
+In order for files to be publicly readable, the app must request 
+the `publish_data` scope during authentication.
+
+```JavaScript
+ let options = { 
+   user: 'ryan.id', // the Blockstack ID of the user for which to lookup the file
+   app: 'BlockstackApp.com' // origin of the app this file is stored for
```
larrysalibra: This example should be an origin: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin#Syntax

src/storage/README.md
```diff
@@ -47,9 +51,30 @@ version. These features will be rolled out in future updates._
...
+ blockstack.getFile("/message.txt", options)
  .then((fileContents) => {
     // get & decrypt the contents of the file /message.txt
     assert(fileContents === "Secret hello!")
  });
 ```
+
+## Reading another user's unencrypted file
+In order for files to be publicly readable, the app must request 
+the `publish_data` scope during authentication.
```
larrysalibra, line 67: I like that `publish_data` is similar to what facebook calls their write permission: https://developers.facebook.com/docs/facebook-login/permissions#reference-publish_actions

src/storage/README.md
```diff
@@ -47,9 +51,30 @@ version. These features will be rolled out in future updates._
...
+
+```JavaScript
+ let options = { 
+   user: 'ryan.id', // the Blockstack ID of the user for which to lookup the file
```
larrysalibra, line 71: What about calling this `username` since this is a Blockstack ID's `username`?

src/storage/index.js
larrysalibra: I'm still seeing this issue.

larrysalibra (changes requested):
Looks like we're almost there! Very excited about this.

larrysalibra:
I just pushed a commit to default to the current app origin if a user doesn't specify it.

I wanted to modify the tests to cover that change in behavior and noticed that we don't have tests for the`getFile` and `putFile` - or any behavior except for `getFile` 404. @yknl it would be great if you could add tests for the new file reading and writing functionality.

larrysalibra (approved):

larrysalibra:
@yknl Thanks for making those changes. I added 2 more tests: 1 for getting a file from another user in the current app and one for specifying a different app.

I think we're ready to merge this!

larrysalibra:
This was merged: https://github.com/blockstack/blockstack.js/commit/fd8dc3fbfcb8993954f1da7275af00881fd8c0d6

Apparently github really doesn't like when you merge on the command line and delete the remote branch before pushing the merge commits to develop (like the git flow tool does) - it marked this pull request as closed instead of merged.

Issue #320, opened 2017-12-18
Encrypted File Operations handling of `null`


kantai:
Right now, if a file doesn't exist, `getFile` should return `null`, however, when encryption and decryption is enabled, the decryption routine attempts to decrypt `null`, resulting in some kind of Type Error. This should return `null` instead.

larrysalibra:
This is fixed and deployed in blockstack.js v0.14.1.

Thank you @kantai !

PR #326, opened 2017-12-19
Fix null handling problems with encryption

kantai:
Simple fix for Issue #320 -- if the return file is `null` (happens on 404), don't try to decrypt it, just return `null`.

### DIFF
--- a/dist/blockstack.js
+++ b/dist/blockstack.js
@@ -5347,7 +5347,7 @@ module.exports={
   "_args": [
     [
       "bigi@1.4.2",
-      "/home/aaron/devel/blockstack.js"
+      "/Users/larry/git/blockstack.js"
     ]
   ],
   "_from": "bigi@1.4.2",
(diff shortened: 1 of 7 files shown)


larrysalibra (approved):
Works for me!

larrysalibra (changes requested):
Can you add test coverage for this behavior?

kantai:
We'll need to add support for localStorage to the test framework I think.

larrysalibra:
@kantai we've used this to mock localstorage elsewhere https://www.npmjs.com/package/node-localstorage

kantai:
I was planning on using this package:

https://www.npmjs.com/package/mock-local-storage

But `node-localstorage` looks like it'll work too

kantai:
I went with `mock-local-storage`, because `node-local-storage` uses the local file system, and I don't want the test to depend on the file system.

larrysalibra:
Thanks for the tests @kantai!

PR #333, opened 2017-12-26
Improved Stability in the Buffers used in Encryption

kantai:
Turns out, the quick fix for the buffers used in encryption wasn't a _forever_ fix, because the binary-normal form library used by elliptic, can in fact return many different kinds of interestingly formatted hex string (basically, it doesn't pad), so the quick patch of just prepending a `0` was limited. Anyways, now I just pad the hex string to 64 characters. This _should_ resolve all those bad hex string TypeErrors that the library was giving before.

Added a unit test which derives 2 BNs known to produce 62 and 63 length hex strings, and checks to make sure that they are getting padded correctly. Prior versions will fail that test.

One note for the future -- I ended up changing around my padding function because node 6 doesn't have `padStart`, we should think about a policy going forward for language version support. Do we want to just support NodeJS LTS + Stable?

### DIFF
--- a/src/encryption.js
+++ b/src/encryption.js
@@ -42,7 +42,9 @@ export function getHexFromBN(bnInput: Object) {
     return hexOut
   } else if (hexOut.length < 64) {
     // pad with leading zeros
-    return hexOut.padStart(64, '0')
+    // the padStart function would require node 9
+    const padding = '0'.repeat(64 - hexOut.length)
+    return `${padding}${hexOut}`
   } else {
     throw new Error('Generated a > 32-byte BN for encryption. Failing.')
   }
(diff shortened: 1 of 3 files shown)


larrysalibra:
> Do we want to just support NodeJS LTS + Stable?

We definitely need a version support policy:

I opened an issue about this: https://github.com/blockstack/blockstack.js/issues/334


Taking a look at the PR now.

src/encryption.js
```diff
@@ -35,13 +35,18 @@ function sharedSecretToKeys(sharedSecret : Buffer) {
            hmacKey: hashedSecret.slice(32) }
 }
 
-function getHexFromBN(bnInput: Object) {
+export function getHexFromBN(bnInput: Object) {
```
larrysalibra, line 38: What's BN?

larrysalibra (approved):
This works well on my computer, so approving it.

Thanks @kantai !!

Issue #339, opened 2018-01-08
Verify authentication redirect uri is on app origin


larrysalibra:
The `redirect_uri` an app requests should be on the same origin that the request is for so that rogue apps can't get credentials for other apps.

larrysalibra:
This was merged into develop https://github.com/blockstack/blockstack.js/commit/6bd329553603128f58dce8b67181f0d86ba37580

Issue #338, opened 2018-01-08
Verify auth requests come from the claimed app origin


larrysalibra:
We want to make sure that rogue apps can't request authentication responses for other apps. Doing so would allow them to access the private keys of the other apps.
`verifyAuthRequest` currently checks the following:

```
      isExpirationDateValid(token),
      isIssuanceDateValid(token),
      doSignaturesMatchPublicKeys(token),
      doPublicKeysMatchIssuer(token)
```

Right now a rogue app `https://bob.rogue` could generate an authentication response for `https://alice.legit`

By setting the `domain_name` field to `https://alice.legit`, creating a manifest.json file spoofing `https://alice.legit`'s metadata and hosting it at `https://bob.rogue/manifest.json` (or anywhere for that matter) and setting the `redirect_uri` to a page on `https://bob.rogue/`

To prevent this, `verifyAuthRequest` should do the following checks:

* that the manifest file is located on the app origin (`domain_name `) https://github.com/blockstack/blockstack.js/issues/209
* that `redirect_uri` is to a location on the same origin as `domain_name` https://github.com/blockstack/blockstack.js/issues/339

The browser should use `verifyAuthRequest` and should display the verified `app_origin` to the user so that he can make an informed decision about which app he's actually authorizing.

larrysalibra:
This was merged into develop: https://github.com/blockstack/blockstack.js/commit/6bd329553603128f58dce8b67181f0d86ba37580

larrysalibra:
This was merged https://github.com/blockstack/blockstack.js/commit/6bd329553603128f58dce8b67181f0d86ba37580

Issue #209, opened 2017-05-13
verifyAuthRequest should verify manifest is from same app origin


larrysalibra:

larrysalibra:
A fix to this was merged into develop.

PR #340, opened 2018-01-08
Verify app origin in authentication requests

larrysalibra:
This pull request adds 2 verifications to `verifyAuthRequest` to verify that the manifest file and redirect url are on the same app origin as the one for which authentication is being requested.

It also adds a function `verifyAuthRequestAndLoadManifest` which loads the app manifest if the authentication request is valid - this is used in this commit https://github.com/blockstack/blockstack-browser/commit/d0a6a4185775e120d1e2d6a66045af91afa420fd in the browser branch https://github.com/blockstack/blockstack-browser/tree/feature/verify-app-origin to add app origin verification to the authentication process.

I've also added a utility function to check for same origin, additional test coverage and comments for some of the verification functions.

It addresses these issues: #209 #338 #339

### DIFF
--- a/src/auth/authVerification.js
+++ b/src/auth/authVerification.js
@@ -134,6 +134,13 @@ export function isRedirectUriValid(token: string) {
   return isSameOriginAbsoluteUrl(payload.domain_name, payload.redirect_uri)
 }
 
+/**
+ * Verify authentication request is valid
+ * @param  {String} token [description]
+ * @return {Promise} that resolves to true if the auth request
+ *  is valid and false if it does not
+ *  @private
+ */
 export function verifyAuthRequest(token: string) {
   return new Promise((resolve, reject) => {
     if (decodeToken(token).header.alg === 'none') {
@@ -182,10 +189,10 @@ export function verifyAuthRequestAndLoadManifest(token: string) {
 
 /**
  * Verify the authentication response is valid
- * @param {String} token the authentication request token
+ * @param {String} token the authentication response token
  * @param {String} nameLookupURL the url use to verify owner of a username
- * @return {Promise} that resolves to true if it is valid and false if it does not
- * @private
+ * @return {Promise} that resolves to true if auth response
+ * is valid and false if it does not
  */
 export function verifyAuthResponse(token: string, nameLookupURL: string) {
   return new Promise((resolve) => {
--- a/src/auth/authVerification.js
+++ b/src/auth/authVerification.js
@@ -1,5 +1,5 @@
 import { decodeToken, TokenVerifier } from 'jsontokens'
-import { getAddressFromDID, publicKeyToAddress } from '../index'
+import { getAddressFromDID, publicKeyToAddress, isSameOriginAbsoluteUrl } from '../index'
 
 export function doSignaturesMatchPublicKeys(token) {
   const payload = decodeToken(token).payload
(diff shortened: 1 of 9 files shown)


src/auth/authVerification.js
```diff
@@ -144,7 +164,37 @@ export function verifyAuthRequest(token) {
...
 }
 
-export function verifyAuthResponse(token, nameLookupURL) {
+/**
+ * Verify the authentication response is valid and
+ * fetch the app manifest file if valid. Otherwise, reject the promise.
+ * @param  {String} token the authentication request token
+ * @return {Promise} that resolves to the app manifest file in JSON format
+ * or rejects if the auth request or app manifest file is invalid
+ * @private
```
kantai, line 173: Should this be marked private if we're going to use this function in blockstack-browser?

kantai (commented):
This looks great to me. Tested this with sign-in on an old application, with blockstack-browser@v0.21.6 linked to this version of blockstack.js and it worked seamlessly.

My only question is about how we're marking things public versus private. Should we still mark functions private if we use them in blockstack-browser?

larrysalibra:
Thanks so much for taking a look at this @kantai !

> My only question is about how we're marking things public versus private. Should we still mark functions private if we use them in blockstack-browser?

At the moment, I have been marking as public functions that we want app developers to use in their apps. That's because the documentation script uses that tag to decide if it should show up on http://blockstack.github.io/blockstack.js/index.html - we use a bunch of functionality in the browser that wasn't really designed with the idea that it would be a public api. I'd label this in the category of technical debt.

I think we also need to make clear in the documentation which functions app developers should be considering versus functions that only make sense if you're making an alternative to the browser. What do you think?

kantai:
> That's because the documentation script uses that tag to decide if it should show up on http://blockstack.github.io/blockstack.js/index.html - we use a bunch of functionality in the browser that wasn't really designed with the idea that it would be a public api. I'd label this in the category of technical debt.

Yeah, that seems like the best way to do this for now.

> I think we also need to make clear in the documentation which functions app developers should be considering versus functions that only make sense if you're making an alternative to the browser. What do you think?

Yeah, ultimately, we'll probably want to categorize the API, so that 'things app developers should use' can be separate from 'things to use if you want to interact with the blockstack blockchain' or 'things to use when you want to build an alternative browser' -- will look for an appropriate place to add this to a documentation issue.

vsund:
Couldn't we simply go for making these both categories the same? In a way the Browser is just an app built with `blockstack.js`. Yes, it has some special capabilities, but after all it's an app.

I remember that this was discussed somewhere but unfortunately can't remember the result, so throwing this in here to consider :)

kantai:
> Couldn't we simply go for making these both categories the same? In a way the Browser is just an app built with blockstack.js. Yes, it has some special capabilities, but after all it's an app.

Yeah, long-term, that's the idea, but there's different kind of apps, which is why I think it'd be a good idea to  categorize the API endpoints. Anyways, this discussion should move over to #341

kantai:
Anyways, @larrysalibra, this looks good to merge.

larrysalibra:
Thanks - this is merged into develop despite github not showing it as merged. Apparently deleting the branch before pushing the merge commit breaks that github functionality.

The merge commit: https://github.com/blockstack/blockstack.js/commit/6bd329553603128f58dce8b67181f0d86ba37580

Issue #249, opened 2017-08-28
Add test for optional redirect feature 


larrysalibra:
We need to add a test for #243

abhishek71994:
Hello, I saw the label and felt that I could contribute. If you point me towards the direction in which I should approach, that would be very helpful. 
Thanks.

citizens:
@larrysalibra I took a stab at writing the test: #344

larrysalibra:
This was fixed by @citizens ! Thank you!

PR #344, opened 2018-01-10
Resolve issue #249

citizens:
Took a stab at adding tests for the optional redirect for `signUserOut` #249. I'm new to the codebase, but did what I could to emulate the existing style.


larrysalibra (approved):
this looks great to me! Thank you!

PR #349, opened 2018-01-16
Correct private-labelling in documentation

kantai:
Corrects the documentation of:

* encryptECIES
* decryptECIES
* setLocalGaiaHubConnection

Marks them as private, because they are not exported from index.js.

Because this is only a documentation clean-up, this shouldn't require a new release or any additional coverage testing.


yknl:
This looks good to me, thanks @kantai.
@larrysalibra will have to merge this one in, as I don't have the correct permissions on this repo.

larrysalibra (approved):
Thanks for the quick turn around on this @kantai

PR #361, opened 2018-02-02
switch babel preset to env #360

larrysalibra:
Addresses #360 by removing the deprecated `babel-preset-es2015` package and replacing it with `babel-preset-env`.

I tested this by running the build process and test suite.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["flow", "es2015"]
+  "presets": ["flow", "env"]
 }
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["flow", "env"]
+  "presets": ["env", "flow"]
 }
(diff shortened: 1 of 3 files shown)


kantai:
I've also tested this in the test suite -- plus the blockstack-browser test suite by linking it in. That works great --

I think this PR should also update the `.babelrc` and the `compile-tests` though:

```
~/devel/blockstack.js
aaron@cornelius $ cat .babelrc 
{
  "presets": ["flow", "es2015"]
}
~/devel/blockstack.js
aaron@cornelius $ cat package.json | grep es2015
    "compile-tests": "rm -rf tests/unitTests/lib; rm -rf tests/authTests/lib; babel --presets es2015 tests/unitTests/src -d tests/unitTests/lib; babel --presets es2015 tests/authTests/src -d tests/authTests/lib;",
```

larrysalibra:
great catch @kantai !

larrysalibra:
@kantai pushed a commit to make those changes.

larrysalibra:
Here's info on babel preset ordering: https://babeljs.io/docs/plugins/#pluginpreset-ordering

kantai (approved):
This looks great and works for me @larrysalibra -- will merge to develop.

PR #363, opened 2018-02-07
Add code coverage results to `npm run test`

kantai:
Adds code coverage results to `npm run test` using the `nyc` npm package.

Example output here:

nyc-code-coverage

The output can be configured to be smaller (not per-file) by changing the `reporter=` flag in the package.json to be `text-summary` instead of `text`.

You can test this by running `npm run test`

I found the code coverage results in `blockstack-browser` to be really helpful (and motivating) when adding tests for changes, so I figured that adding that to `blockstack.js` would be helpful as well. If people don't like this output, that's okay, we don't necessarily need to add this.


larrysalibra:
I like this! Checking it out now!

larrysalibra:
screen shot 2018-02-07 at 11 33 43 pm

beautiful @kantai !

larrysalibra (approved):
Works great for me!

PR #365, opened 2018-02-09
Send bitcoind credentials in request header for local regtest environment

yknl:
Send bitcoind credentials in request header for local regtest environment so that we can use the CORS proxy.

In order to use the bitcoind node in the regtest docker container, we need to pass the requests through the CORS proxy. The credentials needed to be passed in an authorization header instead of in the URL.

### DIFF
--- a/dist/blockstack.js
+++ b/dist/blockstack.js
@@ -1277,115 +1277,47 @@ Object.defineProperty(exports, "__esModule", {
   value: true
 });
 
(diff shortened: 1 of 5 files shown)


kantai (approved):
These changes look good to me. Sorry it took me a while to get this reviewed -- it's not easy to find pending reviews in my github notifications.

kantai:
Will resolve conflicts now.

yknl:
Hey @kantai , I've resolved the conflicts, going to push it now.

Issue #362, opened 2018-02-07
Implement transaction broadcast api


larrysalibra:
Implement the transaction broadcast javascript api defined here:

https://github.com/blockstack/blockstack-browser/issues/1201

Each API method will make a request to the transaction broadcast service.

Will mock responses until we've confirmed the specific transaction broadcast api.

larrysalibra:
@kantai @yknl i've added function signatures for the transaction broadcast service api here: https://github.com/blockstack/blockstack.js/commit/aec835eafe62d7e9fe2d09cce79cc085719008f7#diff-afe8509c39d4f5df1c2be1db4b65842cR106

Please take a look at these to see if your understanding is the same as mine.

I've also put a comment in each function with a proposed http rest api for the endpoint of transaction broadcast service API. @kantai please let me know if you're planning to or already using something different.

We also need to define the various error conditions along with more specific return values for the 3 functions.

kantai:
@larrysalibra -- this matches my understanding, and when implementing the broadcaster, I'll match that API.

One note --  `broadcastZoneFile` -- there's another function in `network.js` called `publishZoneFile` -- do you think these should be integrated the same way that `broadcastTransaction` is integrated? Or at the very least, they should have more differentiated names?

yknl:
The function signatures make sense to me as well. 

> One note -- broadcastZoneFile -- there's another function in network.js called publishZoneFile

Wouldn't `broadcastZoneFile` replace `publishZoneFile`?

kantai:
Ideally you still allow both behaviors -- broadcasting a zonefile directly to a blockstack core endpoint and using a service to watch for a transaction and then broadcast.

larrysalibra:
> Ideally you still allow both behaviors -- broadcasting a zonefile directly to a blockstack core endpoint and using a service to watch for a transaction and then broadcast

We run into this in two places: both `broadcastZoneFile` and `broadcastTransaction`

The way i've started to handle this with `broadcastTransaction` is if you provide a transaction to watch it is sent to the transaction broadcast service otherwise it is sent to the utxo broadcast endpoint.

We can do the same with `broadcastZoneFile`: if a transaction to watch is provided, send to the service otherwise send to the core endpoint.

If we allow configuration of a network without a broadcast service, we could raise an error in the case that the dev tries to watch a transaction but there's not broadcast service configured.

Thoughts @kantai @yknl ?

kantai:
[...]

PR #366, opened 2018-02-12
Broadcast service javascript api implementation

larrysalibra:
This pull request adds the broadcast transaction service API documented in #362 with some slight changes to the request format:

```
      /*
       * POST /v1/broadcast/transaction
       * Request body:
       * JSON.stringify({
       *  transaction,
       *  transactionToWatch,
       *  confirmations
       * })
      */

      /*
       * POST /v1/broadcast/zone-file
       * Request body:
       * JSON.stringify({
       *  zoneFile,
       *  transactionToWatch
       * })
       */

      /*
       * POST /v1/broadcast/registration
       * Request body:
       * JSON.stringify({
       * preorderTransaction,
       * registerTransaction,
       * zoneFile
       * })
       */
```

The public developer API methods added in this pull request check to make sure the developer has passed required parameters, rejecting with an error other wise as part of making our developer API more helpful.

They also reject with an error if a remote service has an error.

As part of implementing the error handling, this PR adds a basic error framework based on a previous discussion we had https://github.com/blockstack/blockstack-core/issues/422#issue-227806997

This is something we should flesh out and further define in a future sprint.

There still needs to be a bit of work done on the specific JSON objects returned by each of these functions. At the moment it just passes whatever the API sends back.

@kantai we should sync up on what info the service will be returning. 

I've also enabled lint coverage of our test framework in this pull request. Apologies for not including this in a separate pull request.

You can test this by running the test framework until the broadcast service is complete.


kantai:
This looks great (thanks for all the linting as well!) to me @larrysalibra. The integration tests are passing for me, so the changes there are okay. 

One note on this API -- should we support the zone file being optionally passed as a Base64 string?

larrysalibra:
> One note on this API -- should we support the zone file being optionally passed as a Base64 string

What's the benefit of that?

kantai:
Really, it would just make storing arbitrary data in the zonefile easier -- rather than trying to pass the data in a json string, it could be passed as a base64 encoding of the bytes. I am convinced this isn't necessary for now, and if we want to expand the API later we can.

kantai (approved):
This looks good to merge, @larrysalibra

Issue #374, opened 2018-02-18
'Uncaught (in promise) null' error when querying profile with no apps key for a public file


dhealy05:
When calling blockstack.getFile on a user with no apps key, 'Uncaught (in promise) null' throws, and nothing is return from the call. This breaks, for example, loops such as the following:

  const filename = 'public_index.json'
  var contacts = ['djh.id', 'pat.id', 'larry.id']
  for(var i = 0; i < contacts.length; i++){
    const publicOptions = {
      username: contacts[i],
      app: 'http://localhost:3000',
      zoneFileLookupURL: 'https://core.blockstack.org/v1/names/'
    }
    var result = await blockstack.getFile(filename, { ...publicOptions })
    console.log(result)
  }

Because djh.id has public_index.json, the first call responds with the relevant file. However, pat.id breaks, and larry.id is never reached. Desired behavior: console.log(result) == null; this is the current behavior when querying for a file that does not exist, for a user *with* an apps key. The error is in storage/index.js; see pull request #375

kantai:
Thanks for raising this issue. This is the same root issue as #432 -- namely, we want a better error message in this case. However, in your example, you'd still need to check for a promise rejection in this case.

I'm closing this issue in favor of #432.

PR #375, opened 2018-02-18
Fix for #374

dhealy05:
Return null from blockstack.getFile for a user with no apps key, instead of reject()ing. This was tested using blockstack.js prior to the update to arrow notation, and has been adapted; it has not been tested with the arrow notation code.


kantai:
@dhealy05 -- for issues like this, we want a policy of rejecting with an error object, rather than returning null (because that can often lead to a cascaded error, where it's no longer easy to determine the root cause of the error).

PR #380, opened 2018-02-20
Modularized the bitcoin operations in network.js

kantai:
The plan for supporting bitcoin lookups for client transaction generation is: 

1. In mainnet, that bitcoin requests will go to an insight-api mimicking service being built by @jackzampolin 
2. In regtest, we interact with bitcoind directly
3. Until (1) is implemented, bitcoin interactions query blockchain.info (as is currently the case in `feature/blockstack-operations`)

I _know_ we didn't want to pursue a driver model, but in order to support the above arrangement, we need three different ways of querying the bitcoin network. In working on writing the API for (1), modularizing and separating these parts out into different "BitcoinNetwork" objects was easier than the alternative.

Will add test cases for the InsightClient -- all of the other code changes are covered by the existing unit tests -- after that, I'll request review.

### DIFF
--- a/dist/blockstack.js
+++ b/dist/blockstack.js
@@ -1639,7 +1639,7 @@ function getPublicKeyFromPrivate(privateKey) {
 Object.defineProperty(exports, "__esModule", {
   value: true
 });
-exports.network = undefined;
+exports.network = exports.BlockchainInfoApi = exports.InsightClient = exports.BitcoindAPI = exports.BitcoinNetwork = undefined;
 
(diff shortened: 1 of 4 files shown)


kantai:
Added test cases --- this is ready for review, or a swift rejection, in the case we don't want to try supporting these different bitcoin api endpoints.

jackzampolin:
I'm working on the features we need out of this service right now. I think having an interface here would be 👍

larrysalibra:
> I know we didn't want to pursue a driver model, but in order to support the above arrangement, we need three different ways of querying the bitcoin network. 

Sounds good to me. It's rather late here and I'm not going to get a chance to review this.

yknl (approved):
Thanks @kantai, looks good to me. I tested this with client-side transaction generation in the browser. 
Just had to update the response for `getNamesOwned`, which is already updated in `blockstack-operations`.
```
  getNamesOwned(address: string) {
    const networkAddress = this.coerceAddress(address)
    return fetch(`${this.blockstackAPIUrl}/v1/addresses/bitcoin/${networkAddress}`)
      .then(resp => resp.json())
+     .then(obj => obj.names)
  }
```

kantai:
Thanks @yknl for the review! I'll merge these changes now.

PR #382, opened 2018-02-23
Add blockstack transaction generation support to blockstack.js

kantai:
This PR adds support for transaction generation in blockstack.js:

Unit tests are in:
```tests/unitTests/src/unitTestsOperations.js```

integration tests are in:
```tests/operationsTests/src/operationsTests.js```

There are three main modules of concern here: 
1. `network.js`: this interacts with APIs for blockstack lookups and bitcoin transaction lookups and broadcasts
2. `txbuild.js`: these functions will actually construct blockstack transactions
3. `safety.js`: these are safety checks which will allow a client to ensure that certain conditions are met before generating or broadcasting a transaction (see how they are deployed in, e.g., blockstack-browser: https://github.com/blockstack/blockstack-browser/blob/feature/js-transaction-generation/app/js/profiles/store/registration/actions.js#L99)

Network configuration and selection is done via the `config` key in the `blockstack.js` module. For example, to select a REGTEST default, and then modify the RPC port:

```javascript
  bskConfig.network = bskNetwork.defaults.LOCAL_REGTEST
  bskConfig.network.blockstackAPIUrl = `http://localhost:${port}`
```

This can be tested by running the integration tests in this branch:
```bash
$ npm run integration-test-ops
```
It can be tested against the blockstack-browser by linking this branch into `blockstack-browser`, and following the instructions on the PR here: https://github.com/blockstack/blockstack-browser/pull/1218


package.json
```diff
@@ -7,15 +7,16 @@
...
     "compile": "rm -rf lib; babel src -d lib",
-    "compile-tests": "rm -rf tests/unitTests/lib; rm -rf tests/authTests/lib; babel tests/unitTests/src -d tests/unitTests/lib; babel tests/authTests/src -d tests/authTests/lib;",
+    "compile-tests": "rm -rf tests/unitTests/lib; rm -rf tests/authTests/lib; rm -rf test/operationsTests/lib; babel tests/unitTests/src -d tests/unitTests/lib; babel tests/authTests/src -d tests/authTests/lib; babel tests/operationsTests/src -d tests/operationsTests/lib;",
     "prepublish": "npm run build",
     "dev-auth": "npm run compile; npm run browserify; node ./tests/browserTests/auth-server.js",
     "dev-proofs": "npm run compile; npm run browserify; node ./tests/browserTests/proofs-server.js",
     "unit-test": "npm run lint && npm run flow && npm run compile && npm run compile-tests && npm run browserify && node ./tests/unitTests/lib/index.js",
     "integration-test-auth": "npm run compile; npm run compile-tests; npm run browserify; node ./tests/authTests/lib/index.js",
+    "integration-test-ops": "npm run lint && npm run compile && npm run compile-tests && npm run browserify && node ./tests/operationsTests/lib/index.js",
```
larrysalibra: If integration tests are also going to check for lint, they should also run the type checker.

src/operations/index.js
```diff
@@ -0,0 +1,9 @@
+export {
```
larrysalibra, line 3: These new files are missing the comment to enable flow on them `/* @flow */`

We should eventually enable for all files, but at the moment we need to add this manually.

src/operations/skeletons.js
```diff
@@ -0,0 +1,167 @@
...
+import { config } from '../config'
+
+// todo : add name length / character verification
+
+
+export function makePreorderSkeleton(
+  fullyQualifiedName: string, consensusHash : string, preorderAddress: string,
+  burnAddress : string, burnAmount: number,
+  registerAddress: string = null) {
+  // Returns a preorder tx skeleton.
```
larrysalibra: yay documentation by code!

src/operations/skeletons.js
```diff
@@ -0,0 +1,167 @@
...
+  tx.addOutput(nullOutput, 0)
+  tx.addOutput(preorderAddress, DUST_MINIMUM)
+  tx.addOutput(burnAddress, burnAmount)
+
+  return tx.buildIncomplete()
+}
+
+export function makeRegisterSkeleton(
+  fullyQualifiedName: string, ownerAddress: string,
+  valueHash: string = null) {
```
larrysalibra: `string` can't be null - change to `?string`

src/operations/skeletons.js
```diff
@@ -0,0 +1,167 @@
...
+
+  tx.addOutput(nullOutput, 0)
+  tx.addOutput(ownerAddress, DUST_MINIMUM)
+
+  return tx.buildIncomplete()
+}
+
+export function makeRenewalSkeleton(
+  fullyQualifiedName: string, nextOwnerAddress: string, lastOwnerAddress: string,
+  burnAddress: string, burnAmount: string, valueHash: string = null) {
```
larrysalibra: `string` -> `?string`

src/operations/utils.js
```diff
@@ -0,0 +1,111 @@
...
+
+function transactionBytes(inputs, outputs) {
+  return TX_EMPTY_SIZE +
+    inputs.reduce((a, x) => (a + inputBytes(x)), 0) +
+    outputs.reduce((a, x) => (a + outputBytes(x)), 0)
+}
+
+//
+
+export function estimateTXBytes(txIn : bitcoinjs.Transaction | bitcoinjs.TransactionBuilder,
```
larrysalibra: `txIn` type is listed as `bitcoinjs.Transaction` but then later on the code checks if `txIn` is a `bitcoinjs.TransactionBuilder`. does this function take both types?

src/operations/utils.js
```diff
@@ -0,0 +1,111 @@
...
+    innerTx = txIn.tx
+  }
+
+  return innerTx.outs.reduce((agg, x) => agg + x.value, 0)
+}
+
+export function decodeB40(input: string) {
+  const characters = '0123456789abcdefghijklmnopqrstuvwxyz-_.+'
+  const base = bigi.valueOf(40)
+  return input.split('').reverse()
```
larrysalibra: This return statement is pretty large - making it more verbose would probably improve readability.

larrysalibra (changes requested):
I've tested this both with the built in test suite and the integration tests and it works well. I've also tested this in regtest mode with the browser and have been able to register names with it.

After the changes I mentioned in the comments are made, we'll be able to ship this! Awesome work @kantai !

src/operations/utils.js
larrysalibra: This is more a style thing.

The type checking stuff is important though because some of the type annotations in the new code don't match the behavior of the code.

kantai: I appreciate this -- will tease this apart a little bit and add some comments here. I let my `.reduce()` magic get a little away from me on this one :smile:

src/operations/skeletons.js
kantai: I'm glad you appreciate it!

src/operations/skeletons.js
kantai: Yep -- enabling flow in all these files and tracking down the errors now.

kantai:
Thanks for the review @larrysalibra! I'm working through the flow enforcements now.

src/operations/utils.js
kantai: yep -- it takes both types, I think that the `|` there indicates that it can take either type.

kantai:
These comments are addressed in the latest commit

yknl (approved):

larrysalibra (approved):
I tested this with the blockstack browser and successfully registered both regtest and mainnet names with it. Great work everyone!

kantai:
Thanks for the reviews @yknl and @larrysalibra! Merging...

Issue #369, opened 2018-02-15
Create documentation on Blockstack profiles


larrysalibra:

yknl:
Should this documentation go in a `readme.md` under `/docs`? cc @larrysalibra

larrysalibra:
it can go in a file in `/docs` for now and we can figure out where it lives long term in the future.

PR #385, opened 2018-02-26
Improve profile documentation

yknl:
Improves documentation on profiles #369


larrysalibra (approved):
We probably want to expand this with a bit more info, but it's a good start! Thanks @yknl

PR #386, opened 2018-02-27
Document undocumented auth methods & improve auth documentation

larrysalibra:
This pull request addresses https://github.com/blockstack/blockstack.js/issues/376 and https://github.com/blockstack/blockstack.js/issues/370.

To check out the public docs, run `npm run build-documentation` and open `/docs/index.html` in your browser.

View the updated readme for auth here: https://github.com/blockstack/blockstack.js/blob/feature/auth-doc-improvements/src/auth/README.md


src/auth/README.md
```diff
@@ -1,10 +1,21 @@
...
+* [Manifest file](#manifest-file)
+* [Key pairs](#key-pairs)
+* [Scopes](#scopes)
+* [Authentication tokens](#authentication-tokens)
+* [`blockstack:` custom protocol handler](#blockstack-custom-protocol-handler)
+* [Adding Blockstack Authentication to your app](#adding-blockstack-authentication-to-your-app)
+
+## Overview
+
+Blockstack Authentication is a bearer token-based authentication system. From an app user's perspective, it functions similar to legacy third-party authentication techniques that they're familiar with. From an app developers' perspective it's a bit different than a typical client-server based OAuth flow that you've seen with centralized sign in services in that authentication happens entirely on the client side.
```
kantai: "From an app developers' perspective it's a bit different than a typical client-server based OAuth flow that you've seen with centralized sign in services in that authentication happens entirely on the client side." 

That sentence has a lot of clauses to parse. Maybe reword like:

"For an app developer, the flow is a bit different from the typical client-server flow of centralized sign in services (e.g., OAuth). Rather, with Blockstack, the authentication flow happens entirely client-side."

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
 
-## Operations
+## User flow
 
-### Request a user to sign in
+What follows is a walk through of the experience of a user, Alice, signing in to your app with Blockstack. 
 
-```es6
-import { requestSignIn } from 'blockstack'
+First, Alice clicks the "Sign in with Blockstack" button on your app. She is redirected to her copy of the Blockstack Browser. The Blockstack Browser shows Alice an approval dialog with information about your app including:
```
kantai: This section could be nice as a list:

1. Alice clicks the "Sign in with Blockstack" button on your app, AwesomeApp
2. Alice is redirected to her copy of the Blockstack Browser, showing Alice an approval dialog with information about AwesomeApp including:
    * The origin your app was served from
    * Your app's name
    * Your app's logo
    * The types of permissions and data your app is requesting

3. Alice chooses to authenticate with one of her Blockstack IDs by selecting the ID and clicking Approve.
4. Alice is redirected back to AwesomeApp
5. AwesomeApp receives:
    * a cryptographic proof of Alice's identity
    * public information in Alice's profile
    * access to a dedicated bucket in Alice's Gaia storage hub for application reads and writes

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
+  "icons": [{
+    "src": "http://blockstack-todos.appartisan.com/logo.png",
+    "sizes": "400x400",
+    "type": "image/png"
+  }]
+}
 ```
 
-### Verify an auth request
+The manifest file *MUST* have [Cross-origin resource sharing (CORS) headers](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) that allow the manifest file to be fetched from any arbitrary source. This usually means returning
```
kantai: This usually means returning:

```
Access-Control-Allow-Origin: *
```

kantai: (Was missing the example)

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
+## Key pairs
 
-const verified = verifyAuthRequest(authRequest)
-```
+Blockstack Authentication makes extensive use of public key cryptography. As mentioned above, we use ECDSA with the secp256k1 curve. What follows is a description of the various public-private key pairs used in the authentication process including how they're generated, where they're used and to whom the private key is disclosed.
+
+### Transit private key
 
-### Create an auth response
+The transit private is an ephemeral key that is used to encrypt secrets that need to be passed from the Blockstack Browser to the app during the authentication process. It is randomly generate by the app at the beginning of the authentication response. The public key that corresponds to the transit private key is stored in a single element array in the `public_keys` key of the authentication request token. The Blockstack Browser encrypts secret data such as the app private using this public key and sends it back to the app when the user signs in to the app. The transit private key signs the app authentication request.
```
kantai, line 138: ->

The transit private key is an ephemeral key used to sign the authentication request token and to encrypt secrets passed from the Blockstack Browser to the application during the authentication process. The app randomly generates this key at the start of the authentication process (while calling [`makeAuthRequest`](./src/auth/authMessages.js)). The public key associated with this private key is passed in the authentication request token, in the single element array `authRequestToken.public_keys`. The Blockstack Browser uses this public key to encrypt data in the subsequent authentication response (e.g., the app private key).

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
 
-```es6
-import { makeAuthResponse, makeECPrivateKey } from 'blockstack'
+### Blockstack ID Identity address private key
```
kantai, line 140: Should this just be "Blockstack Identity Key" ?

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
+The identity address private key is derived from the user's keychain phrase and is the private key of the Blockstack ID that the user chooses to use to sign in to the app. It is a secret owned by the user and never leaves the user's instance of the Blockstack browser. This private key signs the authentication response token for an app to indicate that the user approves sign in to that app.
```
kantai: ->
Each Blockstack identity is associated with a private key derived from the user's keychain phrase. The Blockstack ID that signs into an app is derived from a hash of the identity key's corresponding public key (in particular, we use a Base58 check-encoding). This private key is owned by the user and never leaves the user's Blockstack Browser. The identity key signs the authentication response, proving that the user approved the application sign in, however, the identity key _is never_ passed to the application as part of the authentication flow.

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
-const privateKey = makeECPrivateKey()
-const profile = { "name": "Naval Ravikant" }
-const username = "naval.id"
-const authResponse = makeAuthResponse(privateKey, profile, username)
+### App private key
+
+The app private key is an app-specific private key that is generated from the user's identity address private key using the `domain_name` as input. It is deterministic in that for a given Blockstack ID and `domain_name`, the same private key will be generated each time. The app private key is securely shared with the app on each authentication, encrypted by the Blockstack browser with the transit public key.  
+
+The app private key serves three functions. 
+* It is used to create the credentials that give an app access to the gaia hub storage bucket for that specific app.
```
kantai: "It is used to create the credentials that give an app access to the gaia hub storage bucket for that specific app." ->
"It is used to create credentials for accessing the user's app-specific gaia storage bucket."

Also, this list can be numbers instead of bullets.

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
+### App private key
+
+The app private key is an app-specific private key that is generated from the user's identity address private key using the `domain_name` as input. It is deterministic in that for a given Blockstack ID and `domain_name`, the same private key will be generated each time. The app private key is securely shared with the app on each authentication, encrypted by the Blockstack browser with the transit public key.  
+
+The app private key serves three functions. 
+* It is used to create the credentials that give an app access to the gaia hub storage bucket for that specific app. 
+* It is used in the end-to-end encryption of files stored for the app on the user's gaia hub. 
+* It serves as a cryptographic secret that apps can use to perform other cryptographic functions.
+
+## Scopes
```
kantai: Would it make more sense for this subsection to be called "Permission/Scopes"?

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
+Scopes define the information and permissions an app requests from the user during authentication. If no `scopes` array is provided to the `redirectToSignIn` or `makeAuthRequest` functions.
```
kantai: "If no `scopes` array is provided to the `redirectToSignIn` or `makeAuthRequest` functions.": missing second part of the sentence

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
+- `store_write` - read and write data to the user's Gaia hub in an app-specific storage bucket
+- `publish_data` - publish data so that other users of the app can discover and interact with the user
+- `email` - requests the user's email if available
+
+## Authentication tokens
+
+Communication between the app requesting authentication and the Blockstack Browser, the component that authenticates the user, happens via a protocol based on passing JSON web tokens via the URL query string. There are two types of authentication tokens, the authentication request token which is generated by an app requesting authentication of a user and sent to the user's Blockstack Browser and the authentication response token which is sent from the Blockstack Browser to the app when the user approved sign in to the app.
```
kantai: The app and the Blockstack Browser communicate during the authentication flow by passing back and forth two tokens:

1. The requesting application sends the Blockstack Browser an `authRequest` token.
2. Once a user approves a sign in, the Blockstack Browser responds to the application with an `authResponse` token.

These tokens are [JSON Web Tokens](https://jwt.io/), and they are passed via URL query strings.

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
+
+### JSON Web Token signatures
+
+Blockstack's authentication tokens are based on the [ RFC 7519 OAuth JSON Web Token (JWT)](https://tools.ietf.org/html/rfc7519) with additional support for the secp256k1 curve used by bitcoin and many other cryptocurrencies.
+
+This support is indicated by specifying `ES256K` in the `alg` key to indicated that the JWT signature uses ECDSA using the secp256k1 curve.  We provide both [JavaScript](https://github.com/blockstack/jsontokens-js) and [Ruby](https://github.com/blockstack/ruby-jwt-blockstack/tree/ruby-jwt-blockstack) JWT libraries with support for this curve.
```
kantai: This signature algorithm is indicated by specifying `ES256K` in the token's `alg` key, specifying that the JWT signature uses ECDSA with the secp256k1 curve.  We provide both [JavaScript](https://github.com/blockstack/jsontokens-js) and [Ruby](https://github.com/blockstack/ruby-jwt-blockstack/tree/ruby-jwt-blockstack) JWT libraries with support for this signing algorithm.

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
+    email, // email if email scope is requested & email available
+	  profile_url, // url to signed profile token
+    hubUrl, // url pointing to user's gaia hub
+    version // version tuple
+  }
+```
+  
+## `blockstack:` custom protocol handler
+
+The `blockstack:` custom protocol handler is how Blockstack apps send their authentication requests to the Blockstack Browser. When the Blockstack Browser is installed on a user's computer, it registers itself as the handler for the `blockstack:` customer protocol.
```
kantai: Blockstack apps send their authentication requests to the Blockstack Browser via the `blockstack:` custom protocol handler. This protocol handler is registered when Blockstack Browser is installed.

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
+
+When apps call [`redirectToSignIn`](http://blockstack.github.io/blockstack.js/index.html#redirecttosignin) or [`redirectToSignInWithAuthRequest`](http://blockstack.github.io/blockstack.js/index.html#redirecttosigninwithauthrequest), the code tries to check if a handler for the `blockstack:` custom protocol is installed and, if so, redirects the user to `blockstack:<authRequestToken>`. This causes the authentication request token to be passed from the app to the Blockstack Browser which in turn validates the request, displays information about the app to the user and asks the user if she wants to approve sign in to the app.
```
kantai: When an application calls [`redirectToSignIn`](http://blockstack.github.io/blockstack.js/index.html#redirecttosignin) or [`redirectToSignInWithAuthRequest`](http://blockstack.github.io/blockstack.js/index.html#redirecttosigninwithauthrequest), `blockstack.js` checks if a `blockstack:` protocol handler is installed and, if so, redirects the user to `blockstack:<authRequestToken>`. This passes the authentication request token from the app to the Blockstack Browser, which will in turn validate the request and display an authentication dialog.

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
 
-### Verify an auth response
+## Adding Blockstack Authentication to your app
+
+The way you can add Blockstack Authentication to you app depends on whether your app is a modern decentralized Blockstack App where code runs client-side without trusted servers or a legacy client-server app where a server is trusted.
```
kantai: Integrating Blockstack Authentication will depend on the architecture of your application. The integration will be different depending on whether your app is a modern, decentralized, Blockstack App, where all code runs client-side without trusted servers, or your app is a legacy client-server app, where there may be trusted servers.

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
 
-### Verify an auth response
+## Adding Blockstack Authentication to your app
+
+The way you can add Blockstack Authentication to you app depends on whether your app is a modern decentralized Blockstack App where code runs client-side without trusted servers or a legacy client-server app where a server is trusted. 
+
+### Authentication in Client-side apps
+This method is appropriate for decentralized client-side apps where the user's zone of trust - the parts of the app that the user is trusting - begins and ends with the code running on their own computer. In apps like these, any code the app interacts with that's not on their own computer such as external servers does not need to know who she is.
```
kantai: "In apps like these, any code the app interacts with that's not on their own computer such as external servers does not need to know who she is." ->
"In apps like these, any code the app interacts with _not on the user's computer_ should not need to know who the user is."

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
+In this process you'll use these four functions:
+
+- `redirectToSignIn`
+- `isSignInPending`
+- `handlePendingSignIn`
+- `loadUserData`
+
+##### Starting the sign in process
+
+When your app wants to start the sign in process, typically when the user clicks a "Sign in with Blockstack" button, your app will call the [`redirectToSignIn`](http://blockstack.github.io/blockstack.js/index.html#redirecttosignin) method of [blockstack.js](https://github.com/blockstack/blockstack.js).
```
kantai: -> To start the sign in process (typically when your user clicks a "Sign in with Blockstack" button) call the [`redirectToSignIn`](http://blockstack.github.io/blockstack.js/index.html#redirecttosignin) method of [blockstack.js](https://github.com/blockstack/blockstack.js).

src/auth/README.md
```diff
@@ -72,69 +83,196 @@ if (blockstack.isUserSignedIn()) {
...
+
+##### Starting the sign in process
+
+When your app wants to start the sign in process, typically when the user clicks a "Sign in with Blockstack" button, your app will call the [`redirectToSignIn`](http://blockstack.github.io/blockstack.js/index.html#redirecttosignin) method of [blockstack.js](https://github.com/blockstack/blockstack.js). 
+
+This creates an ephemeral transit key, stores it in the web browser's `localStorage`, uses it to create an authentication request token and finally redirects the user to the Blockstack browser to approve the sign in request.
+
+##### Handling an authentication response
+
+When a user approves a sign in request, the Blockstack Browser will return the signed authentication response token to the `redirectURI` specified in `redirectToSignIn`. To check for the presence of this token, your app calls `isSignInPending` and if this returns `true` you then call the `handlePendingSignIn` which decodes the token and returns the signed in user's data and simultaneously storing it to `localStorage` so that it can be retrieved later with `loadUserData`.
```
kantai: ->

When a user approves a sign in request, the Blockstack Browser will return the signed authentication response token to the `redirectURI` specified in `redirectToSignIn`.

To check for the presence of this token, your app should call `isSignInPending`. If this returns `true`, the app should then call `handlePendingSignIn`. This decodes the token, returns the signed in user's data, and simultaneously stores it to `localStorage` so that it can be retrieved later with `loadUserData`.

kantai (changes requested):
I left some comments throughout the doc -- these are mostly stylistic changes -- the content here is great, @larrysalibra. Once you address these comments (or shout at me until I abandon my opinions on style), this documentation looks great (the generated docs look good to me).

src/auth/README.md
larrysalibra: Like this.

src/auth/README.md
larrysalibra: thanks

src/auth/README.md
larrysalibra: Much better wording. Thanks 🥇 !

larrysalibra:
Thanks for the feedback @kantai  - I think I addressed all of your comments. Very helpful!

kantai (approved):

kantai:
Looks good @larrysalibra !

PR #391, opened 2018-03-13
Expose identity address and decentralized ID on user object

### LINKED ISSUE #271: Expose identity address in user data object
We don't store the identity address in the user data object. It exists in the authentication response token, however apps shouldn't have to decode json web tokens to get it at.

It should be added in this function: https://github.com/blockstack/blockstack.js/blob/b6889ec2a6fe45fe29818cd59089a97030109195/src/auth/authApp.js#L132

We could use the key `identityAddress`.

The identity address is stored in the `iss` key of the authentication response token in the following format:
```
"iss": "did:btc-addr:1GBLCrGLATRcLXS8V81RbzyCQPaSv4ATby"
```

We should only store the address value in that key.

We should also expose the raw `did` in a key `did`.

bitpshr:
This PR exposes the identity address and decentralized ID on the user object for convenience.

Resolves #271

### DIFF
--- a/src/auth/authApp.js
+++ b/src/auth/authApp.js
@@ -4,7 +4,7 @@ import { decodeToken } from 'jsontokens'
 import { makeAuthRequest, verifyAuthResponse } from './index'
 import protocolCheck from 'custom-protocol-detection-blockstack'
 import { BLOCKSTACK_HANDLER, isLaterVersion } from '../utils'
-import { makeECPrivateKey } from '../index'
+import { getAddressFromDID, makeECPrivateKey } from '../index'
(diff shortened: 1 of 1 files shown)


bitpshr:
Circling back on this, as it'd prevent a few apps I'm developing from having to manually decode tokens. Is this still a relevant change?

larrysalibra (approved):
This looks great to me me @bitpshr - thanks the for the pull request. It will make it into the next release.

vsund:
This should resolve/supersede also https://github.com/blockstack/blockstack.js/pull/273 I guess?

larrysalibra:
thanks for the heads up @vsund

PR #395, opened 2018-03-16
Update Buffer creation (as this way is deprecated)

vsund:
Quickly searched through the code, this was the only deprecated call.


vsund:
Though the TypeScript compiler complains about this (using also Buffer as input), might be an error in the types (to reference the error here: TS wants to have a `any[]` as input).

kantai:
Thanks for this @vsund -- I'll confirm that this works for me in node v6 then merge this.

kantai:
This works great @vsund, merging now.

Issue #398, opened 2018-03-22
Enable encryption by default


larrysalibra:
Let’s enable file encryption by default so that developers have to specifically disable encryption and that we don’t run into situations where devs mistakenly think encryption is enabled by default.

https://github.com/blockstack/blockstack.js/blob/719865539f651ca08e94d668ac1125f52080c67d/src/storage/index.js#L120

larrysalibra:
Shipping this in v0.17.0

PR #399, opened 2018-03-22
enable encryption by default in storage #398

larrysalibra:
This pull request addresses #398 by changing the behavior of the default options object in the `getFile` and `putFile` storage API calls so that encryption is enabled by default. Previously, it was disabled by default.

Tests have been refactored to work with this API change.

I tested this by updating the Blockstack Todos app to use this new API and verified that it encrypts and decrypts files by default.

I also incremented the version number to 0.17 because this is a breaking API change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "blockstack",
-  "version": "0.16.0",
+  "version": "0.17.0",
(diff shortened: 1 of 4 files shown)


kantai (approved):
This looks good and works for me.

larrysalibra:
Thanks @kantai

PR #402, opened 2018-03-28
Enable Custom Gaia Hubs

kantai:
This PR enables reading a Gaia hub from the `authResponse` token.

I tested this by:

- [x] linking the change into blockstack-todos, logging in with current browser
- [x] linking the change into browser, unlinking change in todos, logging in
- [x] linking the change in both, logging in

The reason we had left this version unbumped is kind of a blockstack source code history lesson. Version numbers were added to the auth protocol using a hacky equality check on " === 1.1.0". This meant that a "1.2.0" request would break the protocol. So we decided to fix the version check, wait a version or two, and then bump the protocol version. Then we forgot about it.

### DIFF
--- a/src/auth/authMessages.js
+++ b/src/auth/authMessages.js
@@ -17,7 +17,7 @@ import {
 
 import { encryptECIES, decryptECIES } from '../encryption'
 
-const VERSION = '1.1.0'
+const VERSION = '1.2.0'
 
 type AuthMetadata = {
   email: ?string,


larrysalibra:
@cwackerfuss can you take a look at this when you have a moment?

blockpac:
Hi @kantai is this required to make the Heroku GAIA-hub work with blockstack.js ?

kantai:
@blockpac -- yes, this is required for apps to be able to use those gaia hubs.

kantai:
Hey @yknl -- do you think you'll be able to look at this during this coming sprint?

yknl:
Yes, I'll look at this during this sprint. 👍

yknl (approved):
Tested via instructions and works

kantai:
Thanks for the review @yknl  -- merging now.

PR #404, opened 2018-04-01
Fix #403: Change default name lookup URL to use core.blockstack.org

### LINKED ISSUE #403: Change default name look up URLs
We need to change the default name lookup URLs from localhost:6270 to core.blockstack.org before shipping the browser without core.

These defaults should come from the network setting in blockstack.js.

brandonparee:
## Fix #403 

Change default name lookup URL by using default stored in `network.js`

## Commit message
* Change default name lookup URL in `storage/index.js` by using the default `blockstackAPIURL` from the `MAINNET_DEFAULT` provided by `network.js`
* Add relevant changes to documentation
* Add relevant changes to tests


kantai (commented):
Thanks for the PR @brandonparee --

This works for me, and the approach is okay, however, it may make more sense to try and read the configured API url.

If you make the default value `null` and then check if the passed URL is false-y, then you can do something like

```javascript
import config from '../config'

...

export function getUserAppFileUrl(path: string, username: string, appOrigin: string,
  zoneFileLookupURL: string? = null) {
  if (!zoneFileLookupURL) {
    zoneFileLookupURL = config.network.blockstackAPIUrl
  }
...
```

That way, if an app reconfigures the network object, that reconfiguration will propagate through the defaults here as well. Do you think that approach would make sense? (We could also update the `src/profiles/profileLookup.js` that way)

kantai:
Hey @brandonparee -- these changes look great, I'm going to move this over to a `blockstack/feature/change-default-name-lookup-url` branch to quickly change some of the interaction between defaults of `lookupProfile` and the other functions. Then we should double check that the defaults all work seamlessly with existing apps.

PR #408, opened 2018-04-07
File Signing (Feature #381)

### LINKED ISSUE #381: Option to sign but not encrypt files
Files written to Gaia are not encrypted or signed by default.  However, blockstack.js already has facilities to sign data, since it already has to do so in order to encrypt it.  We should support an option to sign but not encrypt files, in order to share authentic public data.

We could achieve this in the API by adding a `.sign = true` option in the `options` object passed to `putFile()`.

brandonparee:
## Details

Enable file signing in `blockstack.js` using ECDSA. This will resolve #381. 

* Implement `signECDSA` and `verifyECDSA`
  * Followed example from [Elliptic library](https://github.com/indutny/elliptic#ecdsa)
* `signECDSA` has `privateKey` and `content` as parameters and returns an object containing the original `content` as well as the `signature` and `publicKey`. This object is what gets stored on Gaia.
* `verifyECDSA` has `signatureObject` as a parameter and returns a `Boolean`.
* `putFile` was modified to add an additional `sign` parameter.
  * When `sign` is `true` the file will not be encrypted.
  * Calls `signECDSA`
* `getFile` was modified to add an additional `verify` parameter.
  * Calls `verifyECDSA`
  * Ensures the given `publicKey` matches the address of the Gaia bucket
  * Returns `content` if both checks pass
  * An error is thrown if the checks fail

## Todo

This is just a first attempt at solving this problem, and I'd love all of the feedback I can get!

- [x] Update documentation to reflect changes
- [x] Add to JSDoc comments
- [x] Add tests


brandonparee:
I've added some unit tests for the new new functions I added `signECDSA` and `verifyECDSA`. 

I also fixed some merge conflicts that came up due to the addition of `encryptContent` (great idea by the way!)

This should be ready to merge in, however if there are any additional changes that need to be made I can make those and update the PR.

kantai:
@brandonparee -- this is awesome, thanks!

We're going to include this in a coming patch for enabling signatures in a couple different ways -- so this PR will end up being merged in another PR as well. Will keep you updated.

PR #410, opened 2018-04-09
Remove dependency on hasprop

kantai:
This removes our dependency on `hasprop` which is a stale package, with dependencies that emit npm warnings. I also removed the `hasprop` export from utils, which was previously `undefined`.

I added a regression test for the naval-legacy profile parsing, that ensures that the response from `getPersonFromLegacyFormat` is still the same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,6 @@
     "ecurve": "^1.0.6",
     "elliptic": "^6.4.0",
     "es6-promise": "^4.2.4",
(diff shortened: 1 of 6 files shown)


larrysalibra (approved):
This looks good to me. Integration tests work fine and warnings related to hasprop are gone! thanks for cleaning up this technical debt!

kantai:
Thanks @larrysalibra!

PR #412, opened 2018-04-09
Feature/stacks transactions

jcnelson:
This branch makes the following changes in support of the upcoming iteration of the Stacks token:

## Changelog

* It adds support for *all* Blockstack transaction types to blockstack.js.  This is necessary because we're not going to upgrade the Python CLI to handle Stacks transactions, which means that our node.js CLI needs to be able to generate all Blockstack transaction types.  I have added the relevant methods to `src/operations`, and now test them regularly via the integration test framework (which uses the budding node.js CLI to generate and send these transactions).

* It expands the set of safety checks.  These are consumed primarily by the node.js CLI at this point, but the Browser could use them as well if its transaction support gets expanded.

* It expands the API in the `network.js` file to include querying name and namespace prices in Stacks, as well as querying Stacks account balances and getting raw zone files. 

## Breaking Changes

* The pricing API has changed in Blockstack Core, since it must indicate to clients what units a name or namespace is priced in (BTC, Stacks, etc.).  The price API endpoint here includes the units, and (for Stacks) the price must be a BigInteger since it can easily exceed 2**53 - 1.  All consumers of this API in blockstack.js have been patched to be compatible with it.  However, I do not know (and would need help determining) which other API consumers outside of blockstack.js are affected.

I do not expect this PR to be accepted as-is until you all are happy with the price API.


src/operations/txbuild.js
```diff
@@ -283,9 +448,17 @@ function makePreorder(fullyQualifiedName: string,
...
                     ownerKeyHex: string,
                     paymentKeyHex: string,
-                    zonefile: string) {
+                    zonefile: string,
+                    valueHash: string = '') {
   const network = config.network
-  const valueHash = hash160(Buffer.from(zonefile)).toString('hex')
+  if (valueHash.length === 0) {
+    if (!zonefile) {
+      throw new Error('Need zonefile or valueHash arguments')
```
kantai: Better to check `valueHash` like: `if (!valueHash) `, since that will detect `null` as well. Then, error throws should be: 
```javascript
return Promise.reject(new Error(..
```

src/operations/txbuild.js
```diff
@@ -478,6 +702,177 @@ function makeRenewal(fullyQualifiedName: string,
...
+
+  const paymentKey = hexStringToECPair(paymentKeyHex)
+  const preorderAddress = paymentKey.getAddress()
+
+  const preorderPromise = Promise.all([network.getConsensusHash(),
+                                       network.getNamespacePrice(namespaceID)])
+        .then(([consensusHash, namespacePrice]) =>
+          makeNamespacePreorderSkeleton(
+            namespaceID, consensusHash,
+            network.coerceAddress(preorderAddress),
```
kantai: We decided at some point (actually found it, search on discourse forums is amazing! https://forum.blockstack.org/t/2018-01-18-engineering-meeting/3979/1) that we didn't want to coerce network addresses, because _long-term_ we want clients to have to know the difference between when they're operating on mainnet versus regtest.

kantai:
Awesome @jcnelson -- I'll look around to see if anything downstream is depending on `network.getNamePrice`, but I think the answer right now is that nothing else currently depends directly on it. The browser consumes the `/v1/prices/names/foo.id` endpoint directly, which is bad and definitely qualifies as technical debt, but the changes to getNamePrice in this PR shouldn't affect that.

I left a couple of style comments on the PR, but haven't tested it out yet.

src/operations/txbuild.js
jcnelson: Fixed!

src/operations/txbuild.js
jcnelson: Fixed both

jcnelson:
The consensus from this was to break this PR into smaller PRs---ones without any breaking changes, and ones with breaking changes.  Will go ahead and do that.

jcnelson:
I'm going to go ahead and drop this PR in favor of my `feature/all-transactions` branch, which only supports the current BTC-based transactions.  I'll resubmit `feature/stacks-transactions` when we're ready to support Stacks across the stack (no pun intended).

PR #416, opened 2018-04-10
Allows for non-lowercase usernames to be verifiable in Hacker News

BrunoBernardino:
Fixes https://github.com/blockstack/blockstack.js/issues/415

### DIFF
--- a/dist/blockstack.js
+++ b/dist/blockstack.js
@@ -5345,7 +5345,7 @@ var HackerNews = function (_Service) {
       proofUrl = _get(HackerNews.__proto__ || Object.getPrototypeOf(HackerNews), 'prefixScheme', this).call(this, proofUrl);
 
       for (var i = 0; i < baseUrls.length; i++) {
-        if (proofUrl === '' + baseUrls[i] + proof.identifier) {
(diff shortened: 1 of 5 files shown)


dist/blockstack.js
```diff
@@ -9734,7 +9734,7 @@ module.exports={
   "_args": [
     [
       "bigi@1.4.2",
-      "/Users/larry/git/blockstack.js"
+      "/Users/brunobernardino/Projects/blockstack.js"
```
BrunoBernardino: How can I get rid of these? I `git checkout --`d `docs.json` and `package-lock.json` but apparently that still leaked here. 😬

BrunoBernardino: Ah. It was this: https://github.com/blockstack/blockstack.js/issues/368

BrunoBernardino:
Just noticed this won’t work. The URL can’t be lowercased, because it’s what’s used for the fetching of the content.

I’ll fix it tomorrow.

BrunoBernardino:
Ok, it's ready for a review!

larrysalibra:
Thanks for this @BrunoBernardino - trying this out now!

larrysalibra (approved):
This works well for me! Thanks for the pull request!

PR #419, opened 2018-04-11
Updating auth README example

poffdeluxe:
Spotted this small issue with an example in the `auth/README.md` while working on an app. `handlePendingSignIn` doesn't take a callback anymore and instead returns a promise.


kantai:
Great -- thanks for finding this @poffdeluxe

Issue #378, opened 2018-02-19
Mark deleteFile as NotImplemented in the documentation


kantai:
deleteFile isn't currently implemented, so we should indicate as such in the documentation

vilsbole:
Indeed it is confusing since the [PR](https://github.com/blockstack/blockstack.js/pull/290) is marked as merged but both the implementation in both `master` and `develop` simply throws an `[Error]`(https://github.com/blockstack/blockstack.js/blob/286df82401c5be8d89e352054e2192e0f600ea98/src/storage/index.js#L163-L165)

kantai:
Documentation is fixed.

PR #424, opened 2018-04-12
Delete is not implemented

kantai:
Delete file is not implemented, and should not be marked as working. See issue: #378


kantai:
This is just a documentation hotfix, so will merge now

PR #425, opened 2018-04-12
Add namespace transactions and name revoke

jcnelson:
## ChangeLog

This pull request adds support for all currently-available Blockstack transactions.  It uses only the `/v1/*` endpoints in the Blockstack RESTful API.  It does not support Stacks, and can work against the `master` branch of Blockstack Core today.

This PR was generated by stripping out the Stacks-related logic from `feature/stacks-transactions`.

## Breaking Changes

None.


kantai (changes requested):
I added back in the integration test with operations on the fake `id` namespace -- it's good to leave those in as regression tests.

Otherwise, I ran into a couple issues on this PR.

1. The included integration tests don't work for me. I needed to make a couple changes to fix some type errors (`hash160` wasn't correctly imported, the namespace object needed a `new` operator, the `hash160` function needed to be converted to a hexstring). However, even with those changes, the test doesn't pass--- I got a funding error and then stopped trying to fix the test.

2. safety.js needs more code coverage. In `develop` right now, it has:
```
  safety.js                  |    60.32 |    47.06 |     87.5 |     87.5 | 41,50,51,53,72 |
```
In this branch, it has:
```
  safety.js                  |    44.71 |     38.1 |    48.28 |    56.45 |... 104,106,125 |
```

The test coverage level in `develop` for that file isn't great, but this PR has a bunch of new code here, and it'd be good to see coverage for that.

jcnelson:
Thanks for your feedback!  Will fix these.

jcnelson:
I improved the test coverage for `safety.js` and got the operations integration test to pass.

kantai:
Great, thanks @jcnelson -- I'll take another look at it this evening (or tomorrow morning).

kantai (approved):
These changes are great, thanks @jcnelson -- if you can add some entries to CHANGELOG.md in an [Unreleased] section that'd be awesome, then you can merge this into `develop` whenever.

PR #426, opened 2018-04-13
Refactor promise handling in the proof validation

kantai:
This eliminates a problem with uncaught promise rejections in the proof validation code. Before, cases such as 500 errors or parsing failures in the proof validation would cause a promise to reject, which was ultimately not propagated to the returned (top-level) promise. This led to "Uncaught promise rejection" errors, and would lead to _all_ proofs to be marked as invalid.

I added a "bad proof" test for this, and tested this change by running the proof validation unit tests, and additionally testing some live proofs on the commandline -- a good example of a currently "bad proof" is the following:

```javascript
var bsk = require('blockstack')
var account = [{ '@type': 'Account',
  placeholder: false,
  service: 'instagram',
  identifier: 'xanditt',
  proofType: 'http',
  proofUrl: 'https://www.instagram.com/p/BZZLSOxhM8i/' }]
bsk.validateProofs({account}, '1BiG7hjHukZFd2sZiJFhs8W93pgaUbVeYp', 'xan.id').then((result) => validated = result).catch(console.log)
```
In the current version of blockstack.js, that will yield an uncaught exception warning:
```
> (node:26277) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: Cannot read property '1' of null
```
_and_ validated will never be set.

In this version in this branch, that same code won't yield that warning, nor will it reject. Instead, validated will be set to:

```javascript
[ { service: 'instagram',
    proof_url: 'https://www.instagram.com/p/BZZLSOxhM8i/',
    identifier: 'xanditt',
    valid: false } ]
```

Which is the correct behavior.


kantai:
Pinging @yknl for review -- or is this a better match for @larrysalibra ?

larrysalibra:
I can take this off of @yknl 's plate - I'll go through these PRs have after I've tested out testnet

yknl:
thanks @larrysalibra !

larrysalibra (approved):
I tested this as follows:

I checked out this branch and then ran the following in the top level directory of the repository.

```
 var bsk = require ('./')
var account = [{ '@type': 'Account',
  placeholder: false,
  service: 'facebook',
  identifier: 'xan.ditkoff',
  proofType: 'http',
  proofUrl: 'https://www.facebook.com/xan.ditkoff/posts/10210399256322262' }]

bsk.validateProofs({account}, '1BiG7hjHukZFd2sZiJFhs8W93pgaUbVeYp', 'xan.id').then((result) => validated = result).catch(console.log)
```
```
> validated
[ { service: 'facebook',
    proof_url: 'https://www.facebook.com/xan.ditkoff/posts/10210399256322262',
    identifier: 'xan.ditkoff',
    valid: true } ]
```
I also tried the instagram proof above but this was now true.

To test out invalid proofs, I tested my own hacker news proof, modifying the proof in my hacker news profile so it would be false and verifying that `valid` was `false` and that no errors were thrown.

I also reviewed and ran the additional tests.

Thanks for this @kantai !

Issue #417, opened 2018-04-11
Instagram proofs update


yknl:
Update Instagram social proof checking as Instagram has changed the format of the description meta tag from

```
<meta content="2 Likes, 1 Comments - username (@username) on Instagram: “Verifying my Blockstack ID is secured with the address 1KlNv27j3kk1YfbyykdJ8H3npRejs8d...”" name="description" />
```

to

```
<meta content="2 Likes, 1 Comments - @username on Instagram: “Verifying my Blockstack ID is secured with the address 1KlNv27j3kk1YfbyykdJ8H3npRejs8d…”" name="description" />
```

The required code changes are here:
https://github.com/blockstack/blockstack.js/blob/master/src/profiles/services/instagram.js#L42
and possibly
https://github.com/blockstack/blockstack.js/blob/master/src/profiles/services/instagram.js#L54

Test case needs an update as well:
https://github.com/blockstack/blockstack.js/blob/master/tests/testData/profiles/ken.verification.instagram.html#L57

kantai:
A fix was merged into develop (thanks @cponeill!)

cponeill:
Glad I could help out.

kantai:
This was merged into master in release 17.2.0

PR #428, opened 2018-04-15
Resolving Instagram Proofs Update #417

cponeill:
This fix updates Instagram social proof checking #417 since the format of the meta tag has been changed.

Fixes include:
- Changing regular expression to quickly find "username" in meta tag content.
- Updating test cases.


kantai:
This looks great @cponeill -- I tested this both with a new instagram post and with the old posts. I added a regression test as well, and will merge into `develop` now.

PR #429, opened 2018-04-15
Allow encryption of files with provided public key

bodymindarts:
When using multi-reader storage to share files with other users of an app it can make sense to encrypt the files with the public key of another user. This PR lets callers pass a public key to the `putFile` function which will be used to encrypt a file instead of the appPrivateKey from the user data.


ntheile:
@bodymindarts  Can you give an example usage of this? Also, when the user on the other end tries to decrypt the file does getFile just work normally without passing in any special params?

larrysalibra (approved):
This works well for me! Thank you!

bodymindarts:
Sorry for not responding. @ntheile yes a receiving user can use `getFile` and it _just works_.

larrysalibra:
This pull request will be merged when #461 is approved and merged!

PR #430, opened 2018-04-16
Drop leading zero in version number

larrysalibra:
As we discussed in [last week's engineering meeting ](https://forum.blockstack.org/t/2018-04-12-engineering-meeting/4820/4?u=larry), this pull request drops the leading 0 from the version number. Going forward we will adhere to semantic versioning.

It also updates the package description.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 All notable changes to the project will be documented in this file.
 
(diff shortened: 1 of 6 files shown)


larrysalibra:
I've also added an empty change log which we can start filling in going forward.

kantai:
This looks great to me @larrysalibra -- thanks for going ahead and doing this. I changed the http links to https. Feel free to merge.

PR #437, opened 2018-04-24
Hotfix/safety name count fix

jcnelson:
This fixes a bug in `safety.js` where sponsored names are also counted towards the number of names an address owns.  The consensus core does not count sponsored names towards an address's name quota, so this safety check should only count on-chain names as well.


kantai (approved):
This looks great to me, thanks @jcnelson.

kantai:
Merging this fix as well.

PR #438, opened 2018-04-24
Hotfix/namespace burn address

jcnelson:
The burn address for a namespace is chosen by its `version` field.  `version === 1` means "burn to the default burn address", and `version === 2` means "burn to the namespace's `address` value".  This PR fixes this behavior in `network.js` and fixes the operations unit test to test for this behavior.

### DIFF
--- a/src/network.js
+++ b/src/network.js
@@ -90,23 +90,28 @@ export class BlockstackNetwork {
   }
 
   getNamespaceBurnAddress(namespace: string) {
-    return fetch(`${this.blockstackAPIUrl}/v1/namespaces/${namespace}`)
-      .then(resp => {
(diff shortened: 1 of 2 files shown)


kantai:
This looks clear enough to me -- isn't there an additional condition here though? If the namespace is `v2` and older than a year, the burn address switches back to the default, right?

jcnelson:
Right, thanks for clarifying.  Will make that change too.

jcnelson:
Pushed a fix to ensure that the right burn address gets returned at <1, =1, and >1 years after a pay-to-namespace-creator namespace is revealed.

kantai (approved):
Looks great, thanks @jcnelson

Issue #418, opened 2018-04-11
Add link to bug reporting in console


kantai:
A few different libraries improve the rate of error reporting in their projects by including links to a bug reporting system _in their stack traces_.

This can be really great for getting detailed error reports -- the idea being that we'd just prefill values using URL strings on the form: blockstack.org/bugs

A good place to start would be to look at how other frameworks have implemented this (particularly in javascript), and then try to emulate.

PR #439, opened 2018-04-25
Remove console.log in resolveZoneFileToProfile

kantai:
I find these console.log'd errors pretty annoying, and printing things with console.log as a library is generally frowned upon.

Opening a PR to see if others agree with me :grinning:

### DIFF
--- a/src/profiles/profileZoneFiles.js
+++ b/src/profiles/profileZoneFiles.js
@@ -97,11 +97,9 @@ export function resolveZoneFileToProfile(zoneFile, publicKeyOrAddress) {
           return
         })
         .catch((error) => {
-          console.log(`resolveZoneFileToProfile: error fetching token file ${tokenFileUrl}`, error)
           reject(error)
         })
     } else {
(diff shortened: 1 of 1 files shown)


vsund:
There are quite some `console.log` calls in the code. For my [fork](https://github.com/ntzwrk/blockstack.ts) I went with a global logger class (with different log levels).
Developers can probably help themselves better (independently from where the error was made) with proper logging in place. So I think we even should move to consistent logging for cases that might confuse developers (e.g. in the `Service` classes there are some cases that might not be an error but in case of an error it might be helpful to know which path was taken).

But though I think that this could be off by default :)

hstove:
Agreed that probably the best solution is to have different logging levels, with the ability for developers to specify which level they want to display.

kantai:
Cool -- @hstove, want to start an issue for coming up with a logging solution? (Probably also related to #418).

Until we come up with a logging approach, do folks think we should just remove the console.logs ?

vsund:
Would be harder to then add in logging later without all the `console.log` calls ;)

Issue #440, opened 2018-04-25
Add formal logging levels


hstove:
Coming from the conversation around #439, we should have introduce a formal logger with different levels (i.e. error, warn, info, debug). Then, developers can specify what level they'd want to display, with different configuration for development and production.

vsund:
Just to throw this in here, I have a `Logger` class in the TypeScript fork:
https://github.com/ntzwrk/blockstack.ts/blob/develop/src/debug.ts#L10

hstove:
Thanks for the heads up, we'll likely just want to copy that API and use it wherever you use `Logger` in that project.

vsund:
I also added error levels like I thought they would be, would be cool if someone would have a second eye on this.

Also the code probably isn't anymore this close to the original. I refactored the structure in big parts (and [more to come](https://github.com/ntzwrk/blockstack.ts/issues/15) ;)).

kantai:
This was merged into master in release 17.2.0, thanks @hstove

PR #441, opened 2018-04-25
Adds a logger

hstove:
In response to #440 and #439 , adds a `Logger` class to clean up the console logs. To use, set `blockstack.config.logLevel` to one of

- debug
- info
- warn
- error
- none

Set to `none` for nothing to be logged.


kantai:
This works great for me, thanks @hstove!

kantai (approved):

PR #443, opened 2018-04-26
Add TransactionSigner Interface to makeTransaction functions

kantai:
This adds an interface definition and support for using it to the makeTransaction functions.

Background on the interface is here:

https://forum.blockstack.org/t/proposal-transaction-signer-interface-for-blockstack-js/5043/4

This does not add any new functionality to `blockstack.js`, but someone using the library would be able to define their own transaction signer class to enable support for things like multi-sig addresses, segwit, or hardware wallets.

This is tested via the existing unit tests and integration testing (`npm run operations-tests`)

To test on mainnet, you can link this branch into blockstack-browser and perform a registration.


larrysalibra:
This worked fine with the operations integrations tests `npm run integration-test-ops`

Trying a mainnet transaction now.

larrysalibra (approved):
I also tested this by linking this to the develop branch of the browser and sending a mainnet bitcoin transaction to transfer funds out of my browser wallet to another address. The transaction was successfully signed. Thanks @kantai !

kantai:
Awesome, thanks @larrysalibra -- will add an entry to the changelog.md and then merge.

PR #447, opened 2018-04-26
Hotfix/safety expire block

jcnelson:
Core returns `expire_block`, not `expires_block` on `GET /v1/names/{name}`.


kantai (approved):
Thanks for catching this!

PR #448, opened 2018-04-26
Minor corrections to auth README

nikkolasg:
Some minors things I found while reading the README :)

The first change is lost in the whole paragraph, but it's simply "generate" => "generated"


kantai:
Great, thanks for this @nikkolasg -- will merge this now.

PR #450, opened 2018-04-27
force UTXO values to be integers

jcnelson:
This forces the values of UTXOs to be integers.  It's very hard to test this--I discovered it by accident in the integration test framework when I noticed a test failing because one of the values was `999251499.9999999`.

### DIFF
--- a/src/network.js
+++ b/src/network.js
@@ -617,7 +617,7 @@ export class BitcoindAPI extends BitcoinNetwork {
       .then(resp => resp.json())
       .then(x => x.result)
       .then(utxos => utxos.map(
-        x => Object({ value: x.amount * SATOSHIS_PER_BTC,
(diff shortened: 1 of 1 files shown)


src/network.js
```diff
@@ -617,7 +617,7 @@ export class BitcoindAPI extends BitcoinNetwork {
       .then(resp => resp.json())
       .then(x => x.result)
       .then(utxos => utxos.map(
-        x => Object({ value: x.amount * SATOSHIS_PER_BTC,
+        x => Object({ value: parseInt(Math.round(x.amount * SATOSHIS_PER_BTC)),
```
kantai: This can just be `Math.round`, the `parseInt` is uneccesary.

kantai:
Cool -- thanks for catching this @jcnelson. I switched it to just calling Math.round (which also lets it pass the linter/flow). Merging now.

PR #452, opened 2018-04-27
[Storage] encrypt and decrypt strings

### LINKED ISSUE #269: Allow encryption/decryption without files
In some use cases, some data may need encrypting/decrypting without involving a complete file. An API endpoint which will run en/decryption on the Blockstack core (avoiding the many issues with client-side encryption) would be very helpful.

nikkolasg:
This PR fixes #269 by allowing to encrypt and decrypt strings directly. It's a take over from the work in https://github.com/blockstack/blockstack.js/pull/274 from https://github.com/nivas8292 .

During my code exploration I've noted that the `getFile` and `putFile` never gets tested with encryption turned on. As a further step, I think it might be good to add that test coverage. What do you think ? 
I could open an issue and work on this later on if you think that makes sense.

### DIFF
--- a/src/storage/index.js
+++ b/src/storage/index.js
@@ -45,6 +45,30 @@ export function getUserAppFileUrl(path: string, username: string, appOrigin: str
 }
 
 /**
+ * Encrypts the data provided with the transit public key.
+ * @param {String|Buffer} content - data to encrypt
+ * @return {String} Stringified ciphertext object
+ */
+export function encryptContent(content: string | Buffer) {
+  const privateKey = loadUserData().appPrivateKey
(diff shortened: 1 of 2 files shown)


kantai:
Awesome, thanks for this @nikkolasg -- taking a look at this now.

kantai:
This looks great and worked seamlessly for me. Thanks for adding test cases. Merging #274 into this, adding a changelog entry and then merging into `develop`.

> As a further step, I think it might be good to add that test coverage. What do you think ? I could open an issue and work on this later on if you think that makes sense.

That'd be awesome!

PR #454, opened 2018-04-29
[storage] test coverage for put/get file with encryption turned on

### LINKED ISSUE #453: [storage] Test coverage for `putFile` and `getFile` using encryption
Add a test to use `getFile` and `putFile` with the encryption option turned on.
To put back in context, this has been discussed in a previous PR #452 .

nikkolasg:
This PR fixes #453 by adding a test calling `putFile` and `getFile` with the encryption option turned on. 

For making this work I had to find out a couple of things which may need more attention:

+ `putFile` with an `Object` works seamlessly in other tests while it should **not** a-priori ? Many tests use `{ test: 'test'}` as an input. With encryption turned on, it forces us to use a string otherwise it does not work (thanks to `encryptECIES` and `decryptECIES`) and from what I understand here (i.e. looking at API documentation), this should be the **only allowed behavior**: to accept a string for the file content. If this assumption is correct, then the other tests that write `object`-typed messages should be adapted, so tests only test what the API claims to support. (Btw, I'm not a TS expert, but shouldn't it detect this kind of mismatching types ?)

+ The ability to test both function `putFile` and `getFile` in full conjunction is a bit impaired via the fact that there are **two** different mocking strategy when "writing" the file and when "reading" it. The former functionality uses a stub promise from the `sinon` library and the latter uses the `FetchMock` library to catch the HTTP call to get the data. That means there could be legitimately-good-looking tests that writes a certain file and that retrieve another file using the same URL.
While this is probably OK for now, it would be much better to have a **single interface** that accepts to read or write objects to. We would have the "real" implementation using "Gaia & co." and a mocked implementation of this interface that stores files content in memory (similar to the `LocalStorage` class defined at the top of the test file for the storage API). It would provide a much better end-to-end testing of the functionalities with a simpler codebase (no 2 different testing frameworks for related tasks,etc) and potentially allows for easier refactoring of the storage codebase later on.

I'm quite new to the whole blockstack codebase so it may well be that my suggestions are not well founded for a reason I don't know yet, so any comments are most welcomed !

### DIFF
--- a/tests/unitTests/src/unitTestsStorage.js
+++ b/tests/unitTests/src/unitTestsStorage.js
@@ -27,6 +27,8 @@ const localStorage = new LocalStorage()
 global.localStorage = localStorage
 global.window = {}
 global.window.localStorage = localStorage
+global.window.location = {}
+global.window.location.origin = 'https://myApp.blockstack.org'
 
 export function runStorageTests() {
   test('getFile unencrypted', (t) => {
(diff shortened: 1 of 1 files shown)


kantai:
This is great, @nikkolasg, and it works for me -- merging this now!

PR #461, opened 2018-05-11
Encrypt files with provided key & change encrypt content to take public key

larrysalibra:
This pull request changes the api of `encryptContent` to take a public key as an option instead of a private key to allow us to encrypt content for other parties.

It also includes @bodymindarts's pull request #429 which adds the ability to specify a public key for which to encrypt a file in the `putFile` options object.

I've also added unit test coverage for these changes.

You can test this by running the unit tests and linking this into an app and specifying a 3rd party public key.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,7 @@
 {
   "extends": "airbnb",
   "parser": "babel-eslint",
-  "plugins": [],
+  "plugins": ["flowtype"],
   "ecmaFeatures": {
     "jsx": false,
     "modules": true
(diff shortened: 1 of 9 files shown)


larrysalibra:
@bodymindarts Would love if you'd take a look at this too if you have a chance!

bodymindarts:
It certainly lgtm.  +1 for adding a test case!

larrysalibra:
@yknl Going to go ahead and merge this since @bodymindarts was so kind as to take a look. Thanks @bodymindarts

PR #465, opened 2018-05-21
Add option to pass auth response token to handlePendingSignIn

### LINKED ISSUE #464: Option to pass auth response token to handlePendingSignIn
Current implementation of `handlePendingSignIn` uses `getAuthResponseToken` to get the auth response token from query string. It would be helpful for users implementing their own authentication flow to have an option to pass the auth response token to `handlePendingSignIn`. 

For example, in an electron app (with custom protocol handler `blockstack-electron` and appDomain  'http://localhost:8080' ), the following would be a possible auth flow:
1. In a BrowserWindow, `makeAuthRequest` with 'http://localhost:8080/redirect' as the redirect URL
2. Call `redirectToSignInWithAuthRequest` with auth request created in step 1
3. On successful authentication, Blockstack browser will call the redirect URL 'http://localhost:8080/redirect',  which in turn will redirect to `blockstack-electron://redirect?authResponse=xxxxxxx`
4. In electron app, handle the redirect `blockstack-electron://redirect?authResponse=xxxxxxx` in the main process and pass the auth response token to the BrowserWindow that called `redirectToSignInWithAuthRequest`
5. In BrowserWindow, pass the auth response token using `handlePendingSignIn` to blockstack.js

muneebm:
Changes to resolve #464.
Added a second parameter `authResponseToken` with default value `getAuthResponseToken()` to `handlePendingSignIn` .
New syntax:
`export function handlePendingSignIn(nameLookupURL: string = 'https://core.blockstack.org/v1/names/', authResponseToken: string = getAuthResponseToken())`
Added  unit test 'handlePendingSignIn with authResponseToken' in `unitTestsAuth.js` to validate the changes.

### DIFF
--- a/src/auth/authApp.js
+++ b/src/auth/authApp.js
@@ -122,9 +122,9 @@ export function redirectToSignIn(redirectURI: string = `${window.location.origin
  * Retrieve the authentication token from the URL query
  * @return {String} the authentication token if it exists otherwise `null`
  */
-export function getAuthResponseToken() {
+export function getAuthResponseToken(): string {
   const queryDict = queryString.parse(location.search)
-  return queryDict.authResponse ? queryDict.authResponse : null
+  return queryDict.authResponse ? queryDict.authResponse : ''
 }
 
 /**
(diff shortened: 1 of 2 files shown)


larrysalibra:
Thank you for the pull request @muneebm - I'll take a look at this in the next couple days!

larrysalibra (approved):
This looks great @muneebm and thank you for the test! I'm going to merge this into the existing branch that had similar functionality - I like your parameter order better than what's in the existing branch.

PR #469, opened 2018-05-25
New Gaia Authentication

kantai:
This enables authentication using the new Gaia hub authentication scheme (using JWTs). This solves the issue https://github.com/blockstack/gaia/issues/114

Details on the new authentication scheme are here:

https://github.com/blockstack/gaia/blob/master/README.md#v1-authentication-scheme

If the gaia hub that a user is trying to connect to _doesn't_ support the new authentication scheme, this version will only use the legacy authentication scheme if the challenge text fulfills the following criteria:

1. Is a JSON array
2. Element 1 of the array is "gaiahub"
3. Element 4 of the array is "blockstack_storage_please_sign"

This will make it so that a properly behaving legacy gaia hub should continue to work, but will prevent the client from signing arbitrary challenge texts.

This PR includes unit tests for these changes, and it can be tested with the new version of the gaia hub:

```
var bsk = require('.')
var btc = require('bitcoinjs-lib')
var key = btc.ECPair.makeRandom()
bsk.connectToGaiaHub('https://hub.blockstack, key.d.toHex()).then(x => z = x)
bsk.uploadToGaiaHub('foo.json', '"hello world"', z, 'application/json').then(console.log)
```

If you deploy a legacy gaiaHub, you can test with it as well. You can also test this by linking it into `blockstack-todos`.


src/storage/hub.js
```diff
@@ -34,33 +38,61 @@ export function getFullReadUrl(filename: string,
...
+    const publickey = getPublicKeyFromPrivate(signerKeyHex)
+    const token = Buffer.from(JSON.stringify(
+      { publickey, signature })).toString('base64')
+    return token
+  } else {
+    throw new Error('Failed to connect to legacy gaia hub. If you operate this hub, please update.')
+  }
+}
+
+function makeV1GaiaAuthToken(hubInfo: Object, signerKeyHex: string): string {
```
larrysalibra, line 65: it's not obvious what type of object this takes. is it `GaiaHubConfig`?

This files doesn't have flow type checking enabled. 

We should do that by adding `// @flow` so that it catches these types of issues.

larrysalibra (changes requested):
I tried this out by:
* Running the unit tests
* storing a file using the node command line
* linking it to blockstack-todos

It work perfectly for each instance.

Awesome work @kantai - one thing i'd like to see is enabling type checking on any of the files modified in this pull request that don't already have it enabled. I think this will add some clarity as to the expected type of some function arguments.

larrysalibra (approved):
Thanks for the changes @kantai !

larrysalibra:
@kantai i'm going to go ahead and merge these

kantai:
Thanks @larrysalibra!

Issue #468, opened 2018-05-23
InvalidDIDError uses "param" instead of "parameter"


MichaelFedora:
Everything else, including the superclass, uses `parameter`.

https://github.com/blockstack/blockstack.js/blob/e8b0c2afaaa432283b47726a89b6388699b38125/src/errors.js#L57

^ this should be `super({ code: 'invalid_did_error', message, parameter: '' }) `

vsund:
Can second this, occurred (alongside some other flaws like this) while porting to a type-safe language: https://github.com/ntzwrk/blockstack.ts/commit/3116488faf12b10ee9a200c060e1de0ba736511f

larrysalibra:
Fixed in the next release!

PR #471, opened 2018-05-28
No need to specify parameter in InvalidDIDError

larrysalibra:
We incorrectly specified "param" in the object passed to the call to super when we don't need to specify any parameter at all.

Addresses #468 raised by @michaelfedora and @vsund.

If either of you have a second, let me know if this addresses the issue.

I'm planning on merging this into develop since I don't think we need to actually ship a hotfix release for this change. Will include it in the next release.


MichaelFedora:
Looks great - thanks!

larrysalibra:
Thanks @MichaelFedora !

PR #475, opened 2018-05-28
Remove uri-js so create-react-app will build

### LINKED ISSUE #472: won't work with create-react-app
Whenever I try to build an app I've created with `create-react-app` I get the following error.  I also opened an issue with the `uri-js` library.  I'm also making a quick PR to remove `uri-js` and replace it with `url` from node's api.

```
Creating an optimized production build...
Failed to compile.

Failed to minify the code from this file:

        ./node_modules/blockstack/node_modules/uri-js/dist/esnext/uri.js:42

Read more here: http://bit.ly/2tRViJ9
```

DanielRHayes:
Removes `url-js` in favor of only using `url` from the node api.

fixes #472


larrysalibra (approved):
looks good to me and works well here. Thanks!

PR #477, opened 2018-06-08
Support manually providing transit private key during sign in

larrysalibra:
This pull request adds support for specifying the transit private key as a 3rd parameter to `handlePendingSignIn'. This allows the developer to generate the transit key manually without relying on localStorage for storing the transit key between authentication request and response.

It also includes an additional unit test to test use of this parameter with no transit key stored in localstorage.

You can test this by running `npm run test` or linking it to the todos app.

Thanks to @bodymindarts for bringing up the need for this.


kantai (approved):
This looks great, and works when I tested logging into an application that links it in. Apologies for taking a while to review this.

larrysalibra:
Thanks @kantai - apologies as well for taking so long to merge this.

PR #484, opened 2018-07-02
Add intended hubUrl to the Gaia authentication JWT

kantai:
This adds a `.hubUrl` field to the Gaia authentication JWT.

This will allow (future) versions of Gaia hubs to require that authentication tokens contain the correct `.hubUrl` field, which will prevent a malicious hub from replaying authentication tokens on another hub.

### DIFF
--- a/src/storage/hub.js
+++ b/src/storage/hub.js
@@ -62,7 +62,7 @@ function makeLegacyAuthToken(challengeText: string, signerKeyHex: string): strin
   }
 }
 
-function makeV1GaiaAuthToken(hubInfo: Object, signerKeyHex: string): string {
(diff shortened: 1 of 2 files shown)


larrysalibra (changes requested):
I tested this by linking it to the develop branch of the browser and signing in to graphite docs and storing files.

The code looks fine to me as well.

I couldn't find a test for this function. Is it tested elsewhere?

kantai:
There was an existing test for the V1 gaia connection here: https://github.com/blockstack/blockstack.js/blob/1355d54e8c78cbe802e9d04d2733d4c5546fdff3/tests/unitTests/src/unitTestsStorage.js#L695

I added a check to ensure that the JWT payload contains the expected `hubUrl`.

larrysalibra (approved):
Thanks @kantai - test looks good to me.

PR #489, opened 2018-07-05
Fix a padding issue in ecPairToHexString

kantai:
This fixes an issue in `ecPairToHexString` which will result in generating an _incorrect_ hex string encoding of the private key. This will lead to outputting an address which won't correspond to the original ECPair. I don't believe this function is used elsewhere in our codebase, so this shouldn't have created any issues so far, but this should be patched.

Included a test which uses a private key which previously would have been incorrectly encoded to ensure that the behavior is corrected.


larrysalibra:
Thanks @kantai - will review this before next release.

larrysalibra (approved):
This looks good to me - it needs a change log update. I'll send a commit with it.

Issue #487, opened 2018-07-03
Run npm audit and clean up packages


larrysalibra:

larrysalibra:
The results of this shipped in 18.0.0.

Issue #486, opened 2018-07-03
(Mis)use of stale `isomorphic-fetch` package


kantai:
`isomorphic-fetch` is a stale `npm` package, that seems to no longer be maintained.

This is an issue for `node` clients (like @jcnelson 's `node-cli`), because the `node-fetch` version included in that `isomorphic-fetch` is out of date, and doesn't have the necessary features for things like `storage` lookups of non-text data.

Further, we just shouldn't be using unmaintained packages.

@wbobeirne suggested https://github.com/lquixada/cross-fetch
Another viable alternative is to fork `isomorphic-fetch` and just update `node-fetch`.

wbobeirne:
If it's in a purely node environment (e.g. the cli) we should probably just cut out the middleman and not use any isomorphic fetch package, and use node-fetch directly.

jcnelson:
Either way, the node.js environment will have to monkey-patch in a different `fetch` if `isomorphic-fetch` is present.

kantai:
This library is used in both pure node and browser environments, which means that _some_ patching library has to be used for the same `fetch` calling code to work in both environments (unless we want to use some other system of passing the `fetch` function into `blockstack.js` calls, which we shouldn't want to do). I believe simply replacing `isomorphic-fetch` with `cross-fetch` will resolve this issue.

kantai:
This is fixed in `18.0.2`

PR #492, opened 2018-07-09
Switch to cross-fetch and address vulnerabilities by updating dependencies  

larrysalibra:
This pull request addresses #486 and #487 by switching from isomorphic-fetch to cross-fetch and updating a number of dependencies that were stale or had known vulnerabilities.

As part of switching fetch packages, I refactored some tests that were using the fetch mocking package `nock` to use `fetch-mock` like the majority of our tests. 

Updating `eslint` to address vulnerabilities also required cleaning up some new lint.

You can test these changes by running `npm run test` and also by linking it into Blockstack browser.


.eslintrc
```diff
@@ -2,10 +2,6 @@
   "extends": "airbnb",
   "parser": "babel-eslint",
   "plugins": ["flowtype"],
```
kantai, line 4: I think we should add the following:

```javascript
    "indent": [2, 2, {
     "FunctionDeclaration" : { "parameters": "first" },
     "FunctionExpression" : { "parameters": "first" },
     "ObjectExpression": "first",
     "ArrayExpression": "first",
     "ImportDeclaration": "first",
     "CallExpression": { "arguments": "first" }            
     }], 
```

That tells eslint to align object literals, function definitions, imports, etc., roughly the way that we used to.

kantai: On further checking, I think this relates to the latest version of eslint breaking the version of airbnb's style config we use. Is updating to the latest airbnb style config and matching eslint versions, etc. to that an option? Otherwise, we probably shouldn't extend the airbnb config at all.

kantai:
Apart from the linting issues, this looks great to me, and works when I linked it into an application and performed file operations, user lookups, etc.

.eslintrc
larrysalibra: I was able to do that by moving eslint down to the latest point release of the previous major version. Long term we probably shouldn't extend airbnb but for now this will work.

Issue #491, opened 2018-07-08
Implement listFiles


jcnelson:
Implement `listFiles()` to enumerate the files in an application's Gaia hub.  Depends on https://github.com/blockstack/gaia/pull/128.

jcnelson:
Pull request sent.

larrysalibra:
This will be included in the next release! Thanks @jcnelson !

PR #493, opened 2018-07-09
Feature/listfiles

jcnelson:
This PR adds `listFiles(callback: (name: string) => boolean)` to the storage API.  The implementation takes a callback as an argument, which in turn takes the name of the file as input and outputs a boolean.  The callback's return value indicates whether or not `listFiles` should continue fetching files.  The method returns a Promise that resolves to the total number of files listed.  The method is hardcoded to stop listing after 65536 pages in order to avoid endless data attacks from a faulty Gaia hub.

This patch depends on the Gaia `list-files` endpoint, implemented as a PR [here](https://github.com/blockstack/gaia/pull/128)


src/storage/index.js
```diff
@@ -452,4 +452,82 @@ export function deleteFile(path: string) {
   Promise.reject(new Error(`Delete of ${path} not supported by gaia hubs`))
 }
 
+/**
+ * Loop over the list of files in a Gaia hub, and run a callback on each entry.
+ * Not meant to be called by external clients.
```
larrysalibra, line 464: We should mark this as `@private` so that it doesn't show up in the public documentation.

src/storage/index.js
```diff
@@ -452,4 +452,82 @@ export function deleteFile(path: string) {
...
+      }
+    })
+}
+
+/**
+ * List the set of files in this application's Gaia storage bucket.
+ * @param {function} callback - a callback to invoke on each named file
+ * @return {Promise} that resolves to the number of files listed
+ */
+export function listFiles(callback: (name: string) => boolean) : Promise<number> {
```
larrysalibra: I'm not entirely clear on the purpose of the callback to invoke on each file.

As a developer using this API, I would expect it to return a list of files instead of taking a callback that performs an operation on the files and returning the number of files.

It seems like this is actually couple api calls in one:

`listFiles` -> this could optionally take: a number of files to list, a page number and/or potentially a query string and resolve to an array of the file paths. Each of 

`totalFiles` or `countFiles` -> this could resolve to the number of files and is just a convenience method implemented by calling listFiles and returning the length of the array

`forEach`- applying a function to each element of the array from `listFiles` - this already exists in JavaScript https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

What about changing the `listFiles` function signature to be something like the following:

`function listFiles(query: string = '*', maxResults: number = 1000, page: number = 0): Promise<Array<string>>`

This would allow developers who want to do something like simple, like list all the files in a bucket that doesn't have that many files to do:

`blockstack.listFiles().then((files) => { // do something with files here } `

larrysalibra (changes requested):
Thanks for this @jcnelson - I've added a couple comments about the API. I wasn't involved in the earlier discussions so apologies if these comments have already been discussed.

Another question that comes to mind is whether or not this method is intended to be used on other buckets besides the current app-user bucket?  You could imagine situations where you'd want to list all of the files another user has in a bucket and since they're world-readable by default, providing this functionality seems like it would make sense.  If this is the case, we'd need to update the APIs to support this use case as well.

src/storage/index.js
jcnelson: > As a developer using this API, I would expect it to return a list of files instead of taking a callback that performs an operation on the files and returning the number of files.

The problem is that this list can be huge, and/or can take a while to assemble if either the Gaia hub, the backing store, or the user's network connection is slow.  This API gives developers the ability to work with partial results while the next pages of the listing are in flight.

> this could optionally take: a number of files to list, a page number and/or potentially a query string and resolve to an array of the file paths. Each of

A few things:
* Pages aren't numbers.  Usually, they're HMAC'ed database cursors of some kind, and are opaque to us.  The page identifier is specific to the Gaia driver, and page identifiers cannot be assumed to be reusable across multiple requests.  Also, most backing stores we support do not have the notion of an "offset" at which to start listing---you have to start at the beginning, and iterate through the list of files.
* If you want to list a certain number of files, this can already be serviced with the existing API---simply have your callback keep count of the number of files listed so far.
* Different backing stores will have different query strings (they may not even support regular expression filtering), so it doesn't make sense to allow the application to pass this information in.  We don't want applications to get coupled to particular backing stores.

> totalFiles or countFiles -> this could resolve to the number of files and is just a convenience method implemented by calling listFiles and returning the length of the array

Not all backing stores support a way to efficiently query the number of files.  Also, we'd need to add a separate API route in Gaia to support this.  I think that if we want to have something like `countFiles`, we should discuss it first on Gaia.

> This would allow developers who want to do something like simple, like list all the files in a bucket that doesn't have that many files to do:

This is also achievable with the given API.  Simply have the callback accumulate all file names.

jcnelson:
> Another question that comes to mind is whether or not this method is intended to be used on other buckets besides the current app-user bucket? You could imagine situations where you'd want to list all of the files another user has in a bucket and since they're world-readable by default, providing this functionality seems like it would make sense. If this is the case, we'd need to update the APIs to support this use case as well.

Right now, it's only intended to allow the bucket owner to list files.  It's a privacy concern and a performance concern.  Regarding privacy, we don't want it to be possible for users to list each other's files by default, since it could be used to monitor other users' activities and expose sensitive information.  However, I'm in favor of making it possible for users to somehow mark their buckets as world-readable, as long as it's opt-in.

Regarding performance, Gaia's `/list-files/` endpoint is on the critical path, and the Gaia node is likely going to be the limiting factor on how fast the system can service listings.  We should think very carefully about exposing `/list-files/` to users besides the bucket owner, because we can't move the Gaia hub off of the critical path.

src/storage/index.js
jcnelson: Done

larrysalibra:
This is for #491

jcnelson:
Hey Larry, have I addressed your review to your satisfaction?  What more needs to be done on my end?

src/storage/index.js
larrysalibra: Thanks!

src/storage/index.js
larrysalibra: Okay. This makes sense to me. Let's go with this API and if developers find it confusing we revisit it or create convenience methods for various specific use cases.

larrysalibra (approved):
Thanks for explaining this @jcnelson and apologies for taking so long on this review!

larrysalibra:
I'm going to add a change log entry and will then merge this.

larrysalibra:
I also added an explanation to the comment to clarify what the return value of the callback does.

PR #494, opened 2018-07-09
Feature/pass name lookup url

jcnelson:
This PR makes it so `handlePendingSign()` will take the `nameLookupUrl` from the `authResponse` token, if none is given by the caller.  This is useful because the Browser (or whatever process is handling the sign-in) will then be able to direct the application to the name lookup API endpoint of the user's choice.  We should not be setting a hard-coded default to `core.blockstack.org`.


kantai:
The Core API  endpoint is used in a couple of places in `blockstack.js`, so it's probably better to pass the API endpoint itself, rather than the name lookup URL, e.g., (`https://core.blockstack.org`, `http://localhost:16270`).

Also, we should be able to set the `config.network.blockstackAPIUrl` global, and then have that be the default everywhere that it's used (in verifying auth, doing multi-player reads, profile lookups).

jcnelson:
> The Core API endpoint is used in a couple of places in blockstack.js, so it's probably better to pass the API endpoint itself, rather than the name lookup URL, e.g., (https://core.blockstack.org, http://localhost:16270).

Where is it used?  I don't see `core.blockstack.org` hard-coded anywhere besides `blockstack.config.network`.

Also, isn't `config.network.blockstackAPIUrl` already global?

```
> require('blockstack').config.network.blockstackAPIUrl
'https://core.blockstack.org'
```

kantai:
The current PR adds the `nameLookupURL` to the auth response, and just uses that for the verification step of the auth response. I'm suggesting using the endpoint root instead, so that the authresponse object contains, for example `http://localhost:16270`, rather than `http://localhost:16270/v1/names/`, and also using that to set the `config.network.blockstackAPIUrl`

jcnelson:
I updated the PR to do that.  Now, `config.network.blockstackAPIUrl` is overridden by an API URL from the `authResponse` if present.

jcnelson:
Superseded by https://github.com/blockstack/blockstack.js/pull/503

PR #496, opened 2018-07-16
Update host url to point to webapp

aulneau:
This PR changes the default host constant to point to `browser.blockstack.org` versus the website. This should remove the need for this page -> blockstack.org/auth

cc @larrysalibra


larrysalibra (approved):
Works well for me. Thanks @aulneau !

Issue #462, opened 2018-05-12
Sign in button redirects to wrong URL


ubarsaiyan:
Whenever I press sign in I am redirected to a URL http://localhost:8888/auth?authRequest=/eyJ0e.... which opens a blank page. Removing `/` between `authRequest=` and `eyJ` redirects me to the correct URL which asks for authorisation of the app.
Ubuntu 17.10
Latest version of blockstack 
The problem exists with all apps.

yknl:
hey @ubarsaiyan what browser are you using?

ubarsaiyan:
@yknl Latest Chrome.

andrerfneves:
I have the same issue, but my URL does not have the `/` mentioned.
> http://localhost:8888/auth?authRequest=eyJ0eXAiOiJKV1Qi...

Nonetheless, I end up with a blank screen.
<img width="1440" alt="screen shot 2018-06-14 at 8 57 41 pm" src="https://user-images.githubusercontent.com/2574011/41445160-a57c5262-7015-11e8-8d06-69789496f829.png">

Chrome: Version 67.0.3396.87
Blockstack: Version 17.2.0

lnbc1QWFyb24:
I had this same issue, but fixed it by making sure my manifest.json was correctly configured with webpack and you need to make sure your webpack config has access-control-allow-origin headers. If you use the preconfigured setup used in the tutorials you shouldn't have this problem. Hope that helps!

x5engine:
@aaronbarnardsound already done and all well configured but this gets the wrong profile with no image or full name

wbobeirne:
Handled in #497 & https://github.com/blockstack/blockstack-browser/pull/1528

PR #497, opened 2018-07-17
Expose manifest.json fetch errors

wbobeirne:
Helps address #462 along with [this blockstack-browser PR](https://github.com/blockstack/blockstack-browser/pull/1528). Right now requests to manifest.json that fail (Most commonly from not enabling CORS) are simply swallowed. This exposes the error, and alters logging a bit to be more descriptive.


yknl (approved):
This works well with the browser and mobile SDKs

PR #502, opened 2018-07-26
Redirect iOS Auth Requests to HTTPS Browser

### LINKED ISSUE #481: On iOS safari, always redirect to browser.blockstack.org
Until we ship our native iOS app, we should always redirect to browser.blockstack.org on mobile Safari. 

This will prevent "unsupported protocol handler" messages on iOS devices.

Thanks to @bodymindarts for raising this issue.

Tagging @shreyasthiagaraj so he's aware of this when working on the future iOS app.

wbobeirne:
Closes #481. Redirects iOS devices to `https://browser.blockstack.org/...` instead of `blockstack://...`. Tested on iOS simulator to make sure it works, and macOS Safari to make sure it didn't change that behavior.

### DIFF
--- a/src/auth/authApp.js
+++ b/src/auth/authApp.js
@@ -83,9 +83,17 @@ export function redirectToSignInWithAuthRequest(authRequest: string = makeAuthRe
     window.location = httpsURI
   }
 
-  function unsupportedBrowserCallback() { // Safari is unsupported by protocolCheck
-    Logger.warn('can not detect custom protocols on this browser')
-    window.location = protocolURI
+  function unsupportedBrowserCallback() {
+    if (/iPad|iPhone|iPod/.test(navigator.platform)) {
+      // iOS doesn’t do protocols (yet)
+      Logger.warn('platform does not support custom protocols, sending to https')
+      window.location = httpsURI
+    }
+    else {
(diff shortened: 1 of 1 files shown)


larrysalibra (changes requested):
@wbobeirne this looks good to me except for the failing test that appears to be only because of some lint.

I know we don't have great test coverage at the moment - in the interests of eventually having great test coverage, I'd like to see test coverage on this behavior change.

wbobeirne:
I feel that this particular function is inherently untestable due to its reliance on `custom-protocol-detection-blockstack`, which is a module that relies _heavily_ on browser APIs, e.g. creating iframes, adding them to the document etc. Perhaps we could do some headless browser testing, but I don't think we have a precedent set for that.

I've fixed the lint issue, let me know what you think would be best for testing.

hstove:
Yes, agreed that testing this particular function is tricky. In my opinion, this seems like such a small but important change that we can get by with manual testing and QA for now on this, and then merge, so we can get it to developers as soon as possible. Just my 2 cents.

larrysalibra (approved):
Thanks for the lint fixes @wbobeirne! I think your comments re testing make sense for this.

PR #505, opened 2018-07-27
Update bitcoinjs-lib to v4 (includes interfaces changes)

bodymindarts:
Recently [bitcoinjs v4](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/CHANGELOG.md#401) was released and I updated the library in misthos.
As there are many breaking changes to the interfaces this process took quite a long time and I often had to dig into the internals to see what has changed.
I'm really happy I did do this however since the improvements are tremendous. A run of our entire test suite went from >10minutes to <40 seconds(!!!!).

While the details of how to upgrade are still fresh I thought I'd go ahead and update it in blockstack.js too to save you some of the pain.

I'd love to see it merged soon as currently we have 2 versions of bitcoinjs in our bundle (and also because of the performance gains).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,7 +11,8 @@ export {
 } from './keys'
 export {
   nextYear, nextMonth, nextHour, makeUUID4, updateQueryStringParameter,
-  isLaterVersion, isSameOriginAbsoluteUrl, hexStringToECPair, ecPairToHexString
+  isLaterVersion, isSameOriginAbsoluteUrl, hexStringToECPair, ecPairToHexString,
+  ecPairToAddress
 } from './utils'
 
 export { transactions, safety } from './operations'
--- a/src/keys.js
+++ b/src/keys.js
@@ -1,7 +1,6 @@
 /* @flow */
 import { randomBytes } from 'crypto'
 import { ECPair, address as baddress, crypto as bcrypto } from 'bitcoinjs-lib'
-import bigi from 'bigi'
 
 export function getEntropy(numberOfBytes: number) {
   if (!numberOfBytes) {
@@ -12,7 +11,7 @@ export function getEntropy(numberOfBytes: number) {
 
 export function makeECPrivateKey() {
   const keyPair = new ECPair.makeRandom({ rng: getEntropy })
-  return keyPair.d.toBuffer(32).toString('hex')
+  return keyPair.privateKey.toString('hex')
 }
 
 export function publicKeyToAddress(publicKey: string) {
@@ -23,6 +22,6 @@ export function publicKeyToAddress(publicKey: string) {
 }
 
 export function getPublicKeyFromPrivate(privateKey: string) {
-  const keyPair = new ECPair(bigi.fromHex(privateKey))
-  return keyPair.getPublicKeyBuffer().toString('hex')
+  const keyPair = ECPair.fromPrivateKey(new Buffer(privateKey, 'hex'))
+  return keyPair.publicKey.toString('hex')
 }
--- a/src/network.js
+++ b/src/network.js
@@ -480,7 +480,7 @@ export class BlockstackNetwork {
       reverseHash.reverse()
       excludeSet.push({
         tx_hash: reverseHash.toString('hex'),
-        tx_output_n: utxoUsed.index 
+        tx_output_n: utxoUsed.index
       })
     })
 
@@ -488,7 +488,15 @@ export class BlockstackNetwork {
 
     const txHash = tx.getHash().reverse().toString('hex')
     tx.outs.forEach((utxoCreated, txOutputN) => {
-      if (bitcoinjs.script.classifyOutput(utxoCreated.script) === 'nulldata') {
+      const isNullData = function isNullData(script) {
+        try {
+          bitcoinjs.payments.embed({ output: script }, { validate: true })
+          return true
+        } catch (_) {
+          return false
+        }
+      }
+      if (isNullData(utxoCreated.script)) {
         return
       }
       const address = bitcoinjs.address.fromOutputScript(
@@ -504,7 +512,7 @@ export class BlockstackNetwork {
         tx_hash: txHash,
         confirmations: 0,
         value: utxoCreated.value,
-        tx_output_n: txOutputN 
+        tx_output_n: txOutputN
       })
       this.includeUtxoMap[address] = includeSet
     })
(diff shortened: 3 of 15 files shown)


src/index.js
```diff
@@ -11,7 +11,8 @@ export {
 } from './keys'
 export {
   nextYear, nextMonth, nextHour, makeUUID4, updateQueryStringParameter,
-  isLaterVersion, isSameOriginAbsoluteUrl, hexStringToECPair, ecPairToHexString
+  isLaterVersion, isSameOriginAbsoluteUrl, hexStringToECPair, ecPairToHexString,
+  ecPairToAddress
```
bodymindarts, line 15: `Utils.ecPairToAddress` replaces `ecKeyPair.getAddress()` as it was dropped from the interface of `ECPair`

src/keys.js
```diff
@@ -12,7 +11,7 @@ export function getEntropy(numberOfBytes: number) {
 
 export function makeECPrivateKey() {
   const keyPair = new ECPair.makeRandom({ rng: getEntropy })
-  return keyPair.d.toBuffer(32).toString('hex')
+  return keyPair.privateKey.toString('hex')
```
bodymindarts, line 14: fields have changed on `ECPair`

src/keys.js
```diff
@@ -23,6 +22,6 @@ export function publicKeyToAddress(publicKey: string) {
 }
 
 export function getPublicKeyFromPrivate(privateKey: string) {
-  const keyPair = new ECPair(bigi.fromHex(privateKey))
-  return keyPair.getPublicKeyBuffer().toString('hex')
+  const keyPair = ECPair.fromPrivateKey(new Buffer(privateKey, 'hex'))
```
bodymindarts, line 25: constructor is not exposed anymore need to use `ECPair.from***` methods. Also `fromPrivateKey` takes a regular `Buffer` not a `bigi.fromHex` result.

src/network.js
```diff
@@ -480,15 +480,23 @@ export class BlockstackNetwork {
...
-        tx_output_n: utxoUsed.index 
+        tx_output_n: utxoUsed.index
       })
     })
 
     this.excludeUtxoSet = excludeSet
 
     const txHash = tx.getHash().reverse().toString('hex')
     tx.outs.forEach((utxoCreated, txOutputN) => {
-      if (bitcoinjs.script.classifyOutput(utxoCreated.script) === 'nulldata') {
```
bodymindarts, line 491: classifyOutput is not exposed anymore so we need another way to figure out if its nulldata

src/network.js
```diff
@@ -480,15 +480,23 @@ export class BlockstackNetwork {
...
+      const isNullData = function isNullData(script) {
+        try {
+          bitcoinjs.payments.embed({ output: script }, { validate: true })
```
bodymindarts, line 493: the `bitcoin.payments.{embed,p2sh,p2pkh,...}` functions exist now instead of the `bitcoin.script.***.encode` functions.
So here we are creating a nullData output. If that fails it falls through into the `catch {..}` clause.

src/operations/skeletons.js
```diff
@@ -180,7 +180,8 @@ export function makePreorderSkeleton(
   hashed.copy(opReturnBuffer, 3)
   opReturnBuffer.write(consensusHash, 23, 16, 'hex')
 
-  const nullOutput = bitcoin.script.nullData.output.encode(opReturnBuffer)
+  const nullOutput = bitcoin.payments.embed({ data: [opReturnBuffer] }).output
```
bodymindarts, line 183: again: `bitcoin.script` is gone. `bitcoin.payments` replaces it but with different calling conventions.

src/operations/txbuild.js
```diff
@@ -50,7 +50,7 @@ function fundTransaction(txB: bitcoinjs.TransactionBuilder, paymentAddress: stri
   const txFee = estimateTXBytes(txB, 0, 0) * feeRate
   const outAmounts = sumOutputValues(txB)
   const change = addUTXOsToFund(txB, utxos, txFee + outAmounts - inAmounts, feeRate)
-  txB.tx.outs[changeIndex].value += change
+  txB.__tx.outs[changeIndex].value += change
```
bodymindarts, line 53: The internal field `TransactionBuilder.tx` has been prefixed with `__tx` to indicate that it is private and not to be accessed by clients. I'm sure there are workarounds to achieve the same result without accessing this field directly but I didn't spend the time to figure that out as the tests pass.

src/operations/txbuild.js
```diff
@@ -429,12 +429,13 @@ function makePreorder(fullyQualifiedName: string,
     return Promise.all([network.getUTXOs(preorderAddress), network.getFeeRate(), preorderPromise])
       .then(([utxos, feeRate, preorderSkeleton]) => {
         const txB = bitcoinjs.TransactionBuilder.fromTransaction(preorderSkeleton, network.layer1)
+        txB.setVersion(1)
```
bodymindarts, line 432: By default transactions now all have `version = 2` when they are created. Functionally this has no effect on what blockstack.js actually does, but it does make some tests fail due to different signatures and different transaction sizes resulting from the different signatures. There are tests that make assumptions about the transaction sizes and they fail without setting `version = 1`.

To resolve this you can either change the assumption in the test or add this line. I opted to add this line to demonstrate that all tests pass without modification. Perhaps a reviewer would feel more confident in changing the tests and removing this from the production code.

bodymindarts (commented):
This review documents the first occurrence of each type of change that was necessary with an explanation as to what changed in bitcoin.js that made it necessary.

larrysalibra:
Thanks so much for this @bodymindarts - reading through the changes everything looks good to me. I'll try out the code when I'm in front of my computer in a bit. We should be able to ship this pretty quickly!

wbobeirne:
One thing to update (and my favorite change from this) is we no longer need to warn users about uglify settings, yay! We can remove that bit from [the readme](https://github.com/blockstack/blockstack.js/blob/cd72e61d0789e7d60689219403b7ffb7122ca5d8/README.md#production-usage).

bodymindarts:
@wbobeirne have you actually tested that or is it an assumption at this point? That would be awesome if true!

larrysalibra (approved):
This works well for me @bodymindarts - thank you so much for taking the time to do this. I'm going to merge this into a feature branch so that i can add change log info and investigate whether or not we can update readme about uglification. Will try to get this shipped quickly.

Issue #334, opened 2017-12-27
Decide on Node version support policy


larrysalibra:
We've had an informal policy of supporting the most recent Node LTS.

@kantai brings up a good point that we should have a more formal policy.

Whatever policy we came up with, we should apply it consistently across all of our Javascript projects.

I briefly looked into this it seems like something that requires more than a couple minutes to decide. 

This is one of their two charts:

<img width="914" alt="screen shot 2017-12-27 at 5 58 00 pm" src="https://user-images.githubusercontent.com/597182/34378153-8345425a-eb2f-11e7-98ff-c0c6a1b7880a.png">

I'm thinking supporting "Active" LTS releases might make sense. 

What do you think @kantai @yknl and others in the community?

larrysalibra:
Thoughts @kantai @yknl ?

kantai:
Supporting 'active'  sounds like the right policy to me.

wackenvoss:
I'll third that -- "Active" sounds right for us as an ongoing rule.

larrysalibra:
Updated CircleCI to use node v8.11.3 and our readme to indicate this policy.

PR #506, opened 2018-07-27
Update circleci to test on node 8 the current "active LTS" version #334

larrysalibra:
This pull request changes the CircleCI test environment to test on node 8.11.3 which is the current active LTS version per our discussion on #334.

I had to add a package to flowconfig's ignore section that was causing flow to break on CircleCI.


kantai:
This looks great to me @larrysalibra -- excited to not be testing with node 6 anymore :1st_place_medal:

kantai (approved):

PR #507, opened 2018-07-27
Use polyfill with cross-fetch

kantai:
When we switched from `isomorphic-fetch` to `cross-fetch`, we didn't use the polyfill imports for `cross-fetch` (`isomorphic-fetch` on the other hand, was polyfilling). This leads to errors in any node client of the current version of `blockstack.js` when trying to do `fetching` operations:

```
ReferenceError: fetch is not defined
    at BlockstackNetwork.getNameInfo (/home/aaron/devel/blockstack.js/lib/network.js:190:7)
    at lookupProfile (/home/aaron/devel/blockstack.js/lib/profiles/profileLookup.js:34:44)
    at getUserAppFileUrl (/home/aaron/devel/blockstack.js/lib/storage/index.js:51:38)
    at /home/aaron/devel/blockstack.js/lib/storage/index.js:149:14
```

Switching the imports to `polyfill` resolves that.


kantai:
This should perhaps be a hotfix and ship as a patch version, but I'll leave that decision to @larrysalibra

larrysalibra (approved):
I tried this out locally in my `node` environment and was able to both replicate the bug and verify that this PR fixes it. 

I'll add a change log entry for it and ship a point release.

PR #509, opened 2018-07-27
Feature/update bitcoinjs 4

larrysalibra:


larrysalibra:
I confirmed that bitcoinjs-lib removed the instructions asking users to reserve certain keywords from uglifications: https://github.com/bitcoinjs/bitcoinjs-lib/commit/4838b6b3c26fa25c1c439505e5469290fdd469aa#diff-04c6e90faac2675aa89e2176d2eec7d8

I also tested this assumption out by removed the reserved words from publik's webpack configuration , linking it to this branch and testing it out.

PR #512, opened 2018-07-30
Fixing Inconsistent White Spaces Between Codes

cheezeburger:
Fixed for the following files:
dids.js
index.js
utils.js


larrysalibra (approved):
Thanks!

PR #516, opened 2018-08-02
Fixing mention of the private app

moxiegirl:
@larrysalibra There was a mention of the private `generateAndStoreTransitKey` method in the public docs.  I've fixed that and regenerated the docs.

Signed-off-by: moxiegirl <mary@blockstack.com>

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -266,7 +266,7 @@
             },
             {
               "type": "inlineCode",
-              "value": "generateAndStoreTransitKey",
+              "value": "makeAuthRequest",
               "position": {
                 "start": {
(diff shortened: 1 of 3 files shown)


larrysalibra:
Thanks @moxiegirl that's a great catch! 

It looks like `generateAndStoreTransitKey` should actually be marked as a public method because someone who wants to use `makeAuthRequest` would need to use it to construct the request.

We made a mistake and marked `generateAndStoreTransitKey` as `@private`

I'll merge this into a branch and adjust

moxiegirl:
Thanks @larrysalibra

Issue #514, opened 2018-08-02
Use of isLaterVersion in handlePendingIntent breaks for legacy auth response


friedger:
After generating a legacy auth response, I can't use `handlePendingIntent` because of 
`isLaterVersion` method.

Legacy auth responses do not have a version property. Hence, I get a 
    
    TypeError: Cannot read property 'split' of undefined"

To fix, add an empty version property to legacy auth responses.

PR #519, opened 2018-08-03
Feature/fix islaterversion

larrysalibra:
@friedger I merged #515 into a this feature branch.

I felt a bit uncomfortable about changing the structure of legacy authentication responses because I don't know what old software that break.

I changed it to treat an `undefined` version string as `"0.0.0"`. I also added some unit tests of this function to make sure we're all on the same page as expected behavior. 

@kantai was the one that added versioning to authentication formats, so it would be great to get his input as well to make sure these changes don't break any assumptions.

This is addressing #514


kantai:
This looks good to me, and worked when I tried logging in to an app after linking in browser.

kantai (approved):

friedger:
Much better than my solution :-)

PR #520, opened 2018-08-03
Make sure only public api methods are in generated docs

larrysalibra:
@moxiegirl This pull request exposes the method regarding transit key mentioned in #516 and also marks as private a number of blockchain methods that aren't usable at the moment by people making apps as we discussed.

If you have a moment, would love if you could take a look at this see if it makes sense.


src/auth/authApp.js
```diff
@@ -27,25 +27,26 @@ const DEFAULT_PROFILE = {
...
-  return transitKey
+export function getTransitKey() : string {
+  const transitKey = localStorage.getItem(BLOCKSTACK_APP_PRIVATE_KEY_LABEL)
+  return ((transitKey: any): string)
 }
 
 /**
- * Fetches the hex value of the transit private key from local storage.
+ * Generates a ECDSA keypair to
+ * use as the ephemeral app transit private key
```
moxiegirl, line 41: Maybe just "transit private key" which is the phrase you used for me. We can explain that a transit private key is ephemeral elsewhere.

moxiegirl (approved):
I left a change but it was a nit. If you want to ignore and merge, I can fix it later.

x5engine:
what? a crow coding? @moxiegirl ?

moxiegirl:
@meteorplus Yep.  We do lots of things humans don't expect.

PR #522, opened 2018-08-03
Redirect all Mobile OS's to HTTPS Browser

### LINKED ISSUE #521: FF Android issues with custom protocol handlers
> probably something not supported by firefox here: https://github.com/blockstack/blockstack.js/blob/master/src/auth/authApp.js#L71
> probably our custom protocol detection module https://github.com/larrysalibra/custom-protocol-detection
> 
> Will O'Beirne [2:31 PM]
> Yeah, I agree it's probably in there. I just can't get this damn remote debugger working argh.
> 
> Will O'Beirne [2:49 PM]
> @kenliao Looks like Firefox falsely reports that the handler is supported:
> ```blockstack.js: generating v1.1.0 auth request
> openUriUsingFirefox
> protocol handler detected```
> 
> kenliao [2:54 PM]
> ahh
> this makes a great case for not using protocol handlers, but instead go directly to browser.blockstack.org and use web extensions to intercept requests
> 
> Will O'Beirne [2:55 PM]
> Yeah, this is like the third browser I've run into this month that has some jank with protocol detection

wbobeirne:
Closes #521. A more extreme version of #502, simply redirects _all_ mobile (and tablet) operating systems to browser.blockstack.org. We have no native apps yet that can handle `blockstack://`, and when we do, they can just intercept `browser.blockstack.org`.

As an aside, we'll probably want to eventually redirect to `browser.blockstack.org` always, so this is a good step in that direction.

### Tested that the following...
Now redirect to `browser.blockstack.org`
* Android 8 Chrome - Pixel 2
* Android 8 Firefox - Pixel 2
* iOS 11 Safari - iPhone X Simulator

Still redirect to `blockstack://` when detected
* macOS 10.13.4 Safari
* macOS 10.13.4 Chrome
* macOS 10.13.4 Firefox

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [18.0.3] - 2018-08-06
+## [unreleased]
 
 ### Changed
 - Resolve unsupported protocol error by redirecting
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to the project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [18.0.3] - 2018-08-06
+
+### Changed
+- Resolve unsupported protocol error by redirecting
+  to hosted authenticator on all mobile devices
+
 ## [18.0.3] - 2018-08-01
 
 ### Changed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "blockstack",
-  "version": "18.0.3",
+  "version": "18.0.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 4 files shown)


larrysalibra (approved):
this looks great to me @wbobeirne

moxiegirl:
Discussion from Slack:

JustinHunter [Yesterday at 5:58 AM]
In this pull request (https://github.com/blockstack/blockstack.js/pull/522)

```Still redirect to blockstack:// when detected

macOS 10.13.4 Safari
macOS 10.13.4 Chrome
macOS 10.13.4 Firefox```


What does “when detected” mean? Does that mean when the local browser is detected as being installed on the machine?


6 replies
Yeah -- if it detects an application that can handle the `blockstack://` URI protocol, it will still redirect with that method


Also sent to the channel
JustinHunter [1 day ago]
Doesn’t seem to be working properly then. I do not have the local instance of the browser downloaded, but in Firefox and Safari, it tries to open it. In chrome and brave, the handler takes me to browser.blockstack.org as expected.


Also sent to the channel
Hank [1 day ago]
yeah. if you look at the code, we only redirect to the hosted version in the `failCallback`, which i think only works in chromium browsers. we still have the `unsupportedBrowserCallback`, which just tries to force a redirect to the native app. i was discussing this with @larry in https://github.com/blockstack/blockstack.js/pull/524 , and if we made that code change, it would break for users who use safari but have the native app installed, because there isn't a good way to do a `protocolCheck` for that browser


JustinHunter [1 day ago]
Ah I could have swore this had already been solved for. Guess not. Thanks Hank!


Hank [1 day ago]
i'm not sure, but think we can add a timeout or something in `unsupportedBrowserCallback` to see if the location actually changed, and if not (aka if the timeout runs), then we redirect to the hosted version (edited)


Also sent to the channel
larry [1 day ago]
the way it "detects" the protocol handler in chromium is also rather questionable - i believe it tries to detect a UI element appearing

PR #524, opened 2018-08-06
Send browsers that can't protocol check to HTTPS

wbobeirne:
This is a small PR that changes the behavior of where users are redirected if their browser doesn't support checking if they can handle the `blockstack://` protocol URL. Currently, users of those browsers (namely Safari and Opera) get the following when trying to auth, if they don't have the desktop app installed:

screen shot 2018-08-06 at 11 19 46 am
screen shot 2018-08-06 at 11 20 10 am

<sup>Opera on the left, Safari on the right.</sup>

This PR changes it to default to redirecting browsers that can't do detection to `browser.blockstack.org` instead of `blockstack://` (Much like #522.) This makes the new user experience much better, but effectively makes the native app unusable in these browsers, since the user will never be directed to `blockstack://` and subsequently `localhost:8888`.

### Open Questions

* Do we have a large existing Opera / Safari userbase that this will affect?
* Can we communicate this change to users in a way that won't make them think they've had their app data erased?
* Is this tradeoff worthwhile? Does our community of developers agree?
* Should we provide a way for them to use the native app once they hit `browser.blockstack.org`?

### DIFF
--- a/src/auth/authApp.js
+++ b/src/auth/authApp.js
@@ -92,9 +92,9 @@ export function redirectToSignInWithAuthRequest(authRequest: string = makeAuthRe
   }
 
   function unsupportedBrowserCallback() {
-    // Safari is unsupported by protocolCheck
-    Logger.warn('can not detect custom protocols on this browser')
-    window.location = protocolURI
+    // Safari, Opera & others are unsupported by protocolCheck
+    Logger.warn('can not detect custom protocols on this browser, sending to https')
+    window.location = httpsURI
   }
 
   protocolCheck(protocolURI, failCallback, successCallback, unsupportedBrowserCallback)


hstove:
@larrysalibra is this still relevant? Would love to merge this, I think it would help issues on Chromebooks, etc.

larrysalibra:
It breaks many existing users that have installed the blockstack browser locally.

hstove:
Ok, good info. I think there may be workarounds still, but we'll have to think about it.

larrysalibra:
@hstove for the moment, I think we should add special cases for devices like chromebooks that have never had a locally installable version of the blockstack browser (similar to what we did on mobile) to always redirect those platforms to browser.blockstack.org

markmhendrickson:
@zone117x Assigning this issue to you for review since you may want to close it out (or integrate a solution for it) in light of your protocol handling work

PR #538, opened 2018-08-27
Improvements for multi-sig support

kantai:
This includes improvements to the transaction generation library for use with multi-sig.

1. `coerceAddress` now works correctly for script-hash addresses that it can identity the version of (currently can identify mainnet and testnet bitcoin)
2. The `transaction.makeFoo` functions now have an optional argument `buildIncomplete`, which, if passed, will allow the transaction building library to construct an unsigned transaction hex. This is helpful for passing around unfinished transactions between co-signers.
3. Default transaction version to version 1. The upgrade to bitcoinjs-lib changed the default behavior of the `TransactionBuilder` object to use version 2 transactions. This passes the version 1 explicitly, as version 1 transactions generally have wider support with other libraries we may interact with (in particular, both `trezor.js` and `ledgerjs` are much harder to use with version 2 transactions).

This also includes some code refactoring which eliminated a bunch of duplication in `txbuild.js`.

Tests are added for (1) and (2), and the refactors have full coverage with the existing unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   code. First, coerce address now correctly coerces P2SH
   addresses. Second, bitcoinjs-lib recently switched to defaulting to
   version 2 transactions, which breaks our interoperability with a handful
-  of other libraries. Finally, with comes a little bit of refactoring, to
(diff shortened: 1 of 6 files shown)


larrysalibra (approved):
Looked through this and tried out the tests which worked for me. Thanks @kantai - sorry for the slow turnaround

PR #542, opened 2018-09-10
UserSession API + Storage Strategies

larrysalibra:
This pull request moves user-session related calls into the instance of an object `UserSession`. Data related to a given user-session is encapsulated in this instance and persisted by an instance of a object that extends `SessionDataStore`.

Defaults such as app domain, etc continue to be auto-detected when run in a browser environment.

This means the library can continue to be used with minimal configuration in a web browser environment:

```JavaScript
import { UserSession } from 'blockstack'

userSession.redirectToSignIn()

userSession.getFile('file.json')
```

In a web browser, by default `UserSession` will store session data in the browser's `localStorage` through an instance of `LocalStorageStore`.

This means that app developers can leave management of session state to users.

In a non-web browser environment, it is necessary to pass in an instance of `AppConfig` which defines the parameters of the current app.

```JavaScript
  import { AppConfig, UserSession } from 'blockstack'
  const appConfig = new AppConfig('http://localhost:3000')
  const userSession = new UserSession({ appConfig })
```

In non-browser environments (ie. node), session state is stored in the instance using the `InstanceDataStore`. If a developer wishes to store this elsewhere, he should can extend `SessionDataStore` and implement his own storage strategy.

You can test this by linking it to this branch of the blockstack todos app: https://github.com/blockstack/blockstack-todos/tree/feature/blockstack-19

In the directory of this checked out repo run:

```JavaScript
npm link
```

In the blockstack-todos directory run:

```JavaScript
npm link blockstack
npm run start
```

This pull request requires some additional documentation and test coverage before it is ready to be merged. I wanted to get this in front everyone earlier for review of the proposed API which has been modified slightly from the [original proposal](https://forum.blockstack.org/t/proposal-giving-control-of-session-state-to-app-developers/6006?u=larry).


larrysalibra:
I pushed this branch as blockstack 19.0.0-alpha.3 on npm to help with testing: 
https://www.npmjs.com/package/blockstack/v/19.0.0-alpha.3.

larrysalibra:
You can also test this using the animal-kingdom demo app: https://github.com/larrysalibra/animal-kingdom

kantai:
Thanks for this @larrysalibra -- I'll put this on my review and testing plate for this week.

yknl:
Thanks Larry, the changes look good to me. 

One small thing is that the initialization of the `UserSession` is a little bit confusing.

Using the animal kingdom app as an example, in [`App.js`](https://github.com/blockstack/animal-kingdom/blob/master/src/App.js#L12), a user session is initiated with 
```js
this.userSession = new UserSession()
```

In [`Landing.js`](https://github.com/blockstack/animal-kingdom/blob/master/src/Landing.js#L10) and [`SignedIn.js`](https://github.com/blockstack/animal-kingdom/blob/master/src/SignedIn.js#L17) it is initiated with an app config object 
```js
this.userSession = new UserSession({ appConfig })
```

It's not immediately clear to me why they are different. 

Additionally, would it be better to simplify the config parameter to a generic object? This removes the need for an `AppConfig` class.

For example: 
```js
this.userSession = new UserSession({ 
  appConfig: {
    scopes: ['store_write', 'publish_data']
  },
  sessionStore: sessionDataStore
})
```

src/network.js
```diff
@@ -310,7 +310,9 @@ export class BlockstackNetwork {
    * @return {Promise} a promise that resolves to the WHOIS-like information 
    */
   getNameInfo(fullyQualifiedName: string) {
-    return fetch(`${this.blockstackAPIUrl}/v1/names/${fullyQualifiedName}`)
+    console.log(this.blockstackAPIUrl)
```
kantai: this should be a debug statement

src/network.js
```diff
@@ -321,6 +323,7 @@ export class BlockstackNetwork {
         }
       })
       .then((nameInfo) => {
+        console.log(`nameInfo: ${JSON.stringify(nameInfo)}`)
```
kantai: this should also be a debug statement

src/auth/appConfig.js
```diff
@@ -0,0 +1,105 @@
...
+    this.authenticatorURL = authenticatorURL
+  }
+
+  /**
+   * The location to which the authenticator should
+   * redirect the user.
+   * @returns {string} - URI
+   */
+  redirectURI() : string {
+    return `${this.appDomain}${this.redirectPath}`
```
aulneau, line 95: Would there ever be a case where we might not want to hardcode the `redirectURI` as a combo `appDomain` and `redirectPath`, eg for native apps or electron apps, I might want to share the data bucket for `coinsapp.co` but use `coins://authRequest` as the redirect for the native implementations. cc @kantai @yknl @hstove

kantai (changes requested):
These changes look okay to me. There's two questions I have, though:

1. Why not try to make this somewhat backwards compatible by doing something like setting the UserSession storage strategy on import? That way, we could still have the blockstack.js functions (getFile, putFile, etc.) defined at the top-level, rather than being object methods.

2. If we're going to make these breaking changes, we should include some kind of documentation like "How do I upgrade my application to blockstack.js v19?" Even with that, there will be a lot of out-of-date documentation following this release: @moxiegirl, are you tracking this?

moxiegirl:
@kantai I'm tracking this convo yes. A few points:

- perhaps we should time this merge? We have a lot releasing pre-holidays and not a lot of folks will be around to help if things go wobbly. What do you think?

-  I don't have scope for coming up with an upgrade steps myself...so someone needs to take a rough pass at those.

- Do we need to include upgrading all our samples such as hello-world, ios, android, in concert with the merge.

- Need a formal forum post and something we can use a blog post might be good too

kantai:
> Do we need to include upgrading all our samples such as hello-world, ios, android, in concert with the merge.

Probably not for ios and android, but, eventually, yes for the three sample apps on blockstack.org, and the yeoman app-generator, because those will all be instructions that apply to old versions of the library. 

> Need a formal forum post and something we can use a blog post might be good too

Agreed. A blog post would be ideal, because this is a very nice new functionality, but also a breaking change that our app developers will need to make.

> I don't have scope for coming up with an upgrade steps myself...so someone needs to take a rough pass at those.

Yep -- we'll make sure that someone makes a draft of this document, and I think such a draft would be a requirement before merging these changes.

> perhaps we should time this merge? We have a lot releasing pre-holidays and not a lot of folks will be around to help if things go wobbly. What do you think?

Maybe -- I think we should enumerate what all would need to be updated, and then see whether or not that's achievable before the holidays.

hstove:
I still think this is a pretty heavy breaking change that could be mitigated by my suggestion in the forum:

https://forum.blockstack.org/t/proposal-giving-control-of-session-state-to-app-developers/6006/24

If others (mainly Larry) are ok with this, I can take a stab at the implementation.

Then no one would have to change any code (including our tutorials) and still be able to upgrade.

yknl:
> I still think this is a pretty heavy breaking change that could be mitigated by my suggestion in the forum

I would be supportive of keeping the existing API for a period of time, logging a deprecation warning and eventually deprecate them in a later version.

kantai:
> I still think this is a pretty heavy breaking change that could be mitigated by my suggestion in the forum

I agree @hstove.

I am open to breaking changes, I just want to hear an argument against trying to make this non-breaking first.

src/auth/appConfig.js
friedger: @aulneau custom schemes should not be abused for deep links. What is the specification of coins:// scheme? If another app handles this scheme the authentication will fail because it does not know the transit key.

friedger:
Developers need to do the following changes when adopting this update:
* create a `UserSession` varable, say `blockstackSession`
* replace prefix `blockstack` with the `blockstackSession` for
  - redirectToSignIn
  - isSignInPending
  - handlePendingSignIn
  - loadUserData
  - isUserSignedIn
  - signUserOut
  - makeAuthRequest
  - generateAndStoreTransitKey
  - redirectToSignInWithAuthRequest
  - getAuthResponseToken
  - verifyAuthResponse 
  - getFile
  - putFile
  - deleteFile
  - encryptContent
  - decryptContent

That should be all.

moxiegirl:
@kantai How about updating the `app-generator` code? Or is that ok as is?

Also, we likely need a little project to track all the work to update the sample apps etc.  right? Better than a checklist here.

Like who is going to update the sample application code? I can certainly update the docs but someone needs to check the code in completed projects.

hstove:
> How about updating the app-generator code? Or is that ok as is?

We'd have to update the app generator(s) too.

src/network.js
larrysalibra: Thanks for catching this! I pushed a fixed.

src/network.js
larrysalibra: Thanks for catching this! I pushed a fixed.

aulneau:
Additionally I think we can create a codemod to help with migration: https://github.com/facebook/codemod

A real life example -> https://github.com/styled-components/styled-components-codemods

larrysalibra:
Thanks for all the great feedback everyone! ❤️pull request reviews!

@yknl wrote about `UserSession` optionally taking an object with an `AppConfig`

> It's not immediately clear to me why they are different.

> Additionally, would it be better to simplify the config parameter to a generic object? This removes the need for an AppConfig class.

I agree that the Animal Kingdom example is confusing - the `App.js` instance should probably take an `AppConfig` instance to be consistent.

We could certainly use a "generic" object instead of an instance of a class, however, that object would in and of itself be an implicit type that should be defined in the type checker.

@hstove wrote:

> I still think this is a pretty heavy breaking change that could be mitigated by my suggestion in the forum:

> https://forum.blockstack.org/t/proposal-giving-control-of-session-state-to-app-developers/6006/24

> If others (mainly Larry) are ok with this, I can take a stab at the implementation.

I think this is a really great way to approach maintaining backwards compatibility. @zone117x (i think it was Matt) in our engineering meeting also suggested publishing these changes with the API @hstove suggests along with deprecation warnings in a new 18 release.

zone117x:
I believe for [jsdoc deprecation warnings](http://usejsdoc.org/tags-deprecated.html) to be visible to developers in their editors (code completion / intellisense / IDE warnings) we may need to change the npm packaging. I've outlined the issue a bit on the [forum](https://forum.blockstack.org/t/proposal-giving-control-of-session-state-to-app-developers/6006/22?u=zone117x). To summarize: many editors/IDEs can't parse the babel shims / modules in `lib/*.js`. 

One solution is to perform only the Flow transform from `src/*.js` to `lib/*.js`, then use babel for `lib/*.js` to `dist/blockstack.js`.

This also has the benefit of exposing all the nice jsdoc while coding against blockstack.js.

larrysalibra:
@zone117x I really like [your suggestion](
https://forum.blockstack.org/t/proposal-giving-control-of-session-state-to-app-developers/6006/22?u=zone117x) on improving editor code completion and documentation compatibility. That would be a huge improvement in developer experience.

This is probably something that makes sense as a separate issue/pull request - we can certainly decide to not ship a new release until such a PR makes it into develop.

zone117x:
> This is probably something that makes sense as a separate issue/pull request - we can certainly decide to not ship a new release until such a PR makes it into develop.

I have a lot of this done but its in a Flow to typescript port I experimented with a few weeks ago. I'll check if my non-typescript related package.json and babel changes work. If so I'll PR for further discussion.

zone117x:
Created a [PR](https://github.com/blockstack/blockstack.js/pull/581) that allows support for the deprecation warnings in jsdoc/editor/IDE. The PR may take some time for testing and review so if we don't care very much about the deprecation visibility then might make sense to go ahead with this PR without it.

larrysalibra:
Glad to see this pull request is coming back to life!

yknl:
This PR could use another round of reviews now that the backward compatibility changes are in. 
@hstove @kantai @zone117x

src/auth/authMessages.js
```diff
@@ -47,14 +55,15 @@ type AuthMetadata = {
  * Use this to pass options that aren't part of the Blockstack auth spec, but might be supported
  * by special authenticators.
  * @return {String} the authentication request
+ * @private
  */
-export function makeAuthRequest(transitPrivateKey: string = generateAndStoreTransitKey(),
```
hstove, line 51: This is a breaking change - some apps use `makeAuthRequest`, and it's in the public API. For example, we use this function when telling people how to use the custom Gaia onboarding flow. I'd suggest keeping the method name the same.

src/auth/authApp.js
```diff
@@ -405,51 +494,44 @@ export function handlePendingSignIn(nameLookupURL: string = '',
...
  */
-export function signUserOut(redirectURL: ?string = null) {
-  window.localStorage.removeItem(BLOCKSTACK_STORAGE_LABEL)
-  window.localStorage.removeItem(BLOCKSTACK_GAIA_HUB_LABEL)
-
-  if (redirectURL !== null) {
-    window.location = redirectURL
+export function loadUserDataImpl(caller: UserSession) {
+  const userData = caller.store.getSessionData().userData
+  if (!userData) {
```
hstove, line 550: This is a breaking change in expected functionality. The previous `loadUserData` implementation returns `null` if the user is signed out, but now we are throwing an error.

src/auth/userSession.js
```diff
@@ -0,0 +1,332 @@
...
+  constructor(options: {appConfig?: AppConfig,
+    sessionStore?: SessionDataStore,
+    sessionOptions?: SessionOptions }) {
+    let runningInBrowser = true
+
+    if (typeof window === 'undefined') {
+      runningInBrowser = false
+    }
+
+    Logger.debug(`UserSession: runningInBrowser: ${runningInBrowser ? 'yes' : 'no'}`)
```
hstove: This is pretty verbose, as most apps will have this called on every page. I'd suggest only logging this if `runningInBrowser` is `false`.

hstove (changes requested):
Overall looking good, but I'd like to see a little more backwards compatibility with method behavior.

Origin:
Has anybody tried out these changes with a node app?

Origin:
I'm having some trouble using this branch in a node app with babel, is this where I should ask questions or should I create a post on the forum?

JacobHearst:
I'm having issues getting this running in a node app. When I try and compile and run it with babel, the program crashes on the initialization of `AppConfig` saying that `window` is not defined despite passing in `http://localhost:3000` to the constructor. Is this a babel specific issue?

hstove:
> I'm having issues getting this running in a node app. When I try and compile and run it with babel, the program crashes on the initialization of `AppConfig` saying that `window` is not defined despite passing in `http://localhost:3000` to the constructor. Is this a babel specific issue?

Can you post a code sample and a backtrace? Also note that we haven't built explicit support for node.js environments yet. If you're just compiling with Babel, it should work (I've done it). I suspect you're actually defining your `UserSession` during compilation (as opposed to after the app starts running).

JacobHearst:
> Can you post a code sample and a backtrace? Also note that we haven't built explicit support for node.js environments yet. If you're just compiling with Babel, it should work (I've done it). I suspect you're actually defining your `UserSession` during compilation (as opposed to after the app starts running).

Relevant code:

    const repos = new Server(path.resolve(__dirname, 'tmp'), {
        autoCreate: true
    });
    const port = process.env.PORT || 7005;

    let appConfig;
    let userSession;

    repos.listen(port, () => {
        console.log(`node-git-server running at http://localhost:${port}`);

        appConfig = AppConfig('http://localhost:3200');
        userSession = new UserSession({ appConfig });
    });

Backtrace:

    node-git-server running at http://localhost:7005
    lib\auth\appConfig.js:56
        var appDomain = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : window.location.origin;
                                                                                        ^

    ReferenceError: window is not defined
        at AppConfig (C:\Users\uc252129\Documents\tmp\lib\auth\appConfig.js:56:89)
        at Server.<anonymous> (C:\Users\uc252129\Documents\saved\dist\index.js:36:43)
        at Object.onceWrapper (events.js:313:30)
        at emitNone (events.js:106:13)
        at Server.emit (events.js:208:7)
        at emitListeningNT (net.js:1387:10)
        at _combinedTickCallback (internal/process/next_tick.js:136:11)
        at process._tickCallback (internal/process/next_tick.js:181:9)
        at Function.Module.runMain (module.js:696:11)
        at startup (bootstrap_node.js:204:16)

For a little context, I'm also using a library called `node-git-server` which can be found [here](https://github.com/gabrielcsapo/node-git-server). The server starts and then crashes when I try and define `appConfig`

hstove:
@Origin thanks for that!

So, using `UserSession` in a node.js context isn't supported yet. I've outlined the requirements in #611, but we will merge this PR without it and then add support for node.js environments in a later pull request.

Your code is also not quite correct. When construction an `AppConfig`, the first argument is your scopes, then the second argument is the app URL.

~~~javascript
appConfig = AppConfig(['store_write'], 'http://localhost:3200');
~~~

JacobHearst:
> @Origin thanks for that!
> 
> So, using `UserSession` in a node.js context isn't supported yet. I've outlined the requirements in #611, but we will merge this PR without it and then add support for node.js environments in a later pull request.
> 

Thanks!

> Your code is also not quite correct. When construction an `AppConfig`, the first argument is your scopes, then the second argument is the app URL.
> 
> ```js
> appConfig = AppConfig(['store_write'], 'http://localhost:3200');
> ```

A little too much copy/pasta, thanks for catching that

src/auth/authApp.js
yknl: Although it's changing the behaviour, I would say that throwing an error would be more correct in this case. Can we just make a note of this change in the change logs?

src/auth/authMessages.js
yknl: This is addressed in https://github.com/blockstack/blockstack.js/pull/542/commits/db940b247d0ef7bef24c618bb75ad6c6f04726b5

dist/blockstack.js
```diff
@@ -307,6 +308,39 @@ function signUserOut() {
...
+ * Use this to pass options that aren't part of the Blockstack auth spec, but might be supported
+ * by special authenticators.
+ * @return {String} the authentication request
+ */
+function makeAuthRequest(transitPrivateKey, redirectURI, manifestURI, scopes) {
+  var appDomain = arguments.length > 4 && arguments[4] !== undefined ? arguments[4] : window.location.origin;
+  var expiresAt = arguments[5];
+  var extraParams = arguments.length > 6 && arguments[6] !== undefined ? arguments[6] : {};
+
+  console.warn('DEPRECATION WARNING: The makeAuthRequest() function will be deprecated in the ' + 'next major release of blockstack.js. Use UserSession to configure your auth request.');
```
zone117x: Do we want to have JSDoc `@deprecated` on these as well? They won't show up in editors until we make some improvements to the build / npm packaging, but I think they will show up in the generated docs?

yknl: I had them initially but for some reason documentation.js simply ignores the `@deprecated` tags.

src/auth/authApp.js
hstove: Well we already did the work to make the upgrade easier for folks who want to use v19 without a `UserSession`. Ultimately we want to encourage adoption of this as much as possible. With your backwards compatibility changes, you've made it so that almost every developer can upgrade without changing their code. This would change that. I know I've personally a lot of written code that checks for `loadUserData()` without first calling `isUserSignedIn()`

yknl: I would argue that the point of the backward compatibility changes were to make it easier to migrate to v19. This change in behaviour is more of a bug fix. I think it should be okay as long as we make it clear that there will be some changes required to upgrade. This is after all a major new version.

hstove: Yeah that was my point, we want to make it easier to upgrade. This would actually be the hardest part of upgrading, because so many people use it. It's not like we've documented that it's supposed to throw an error.

hstove: But, you're the maintainer, and I will happily defer to what you think is best!

zone117x (approved):
Ran this against e2e tests with several more API calls and looks good
https://github.com/blockstack/blockstack-browser/pull/1846

src/auth/authApp.js
yknl: Let's keep it this way for now, if it turns out to be a big problem during the beta we can change it back.

PR #546, opened 2018-09-26
Changes for Stacks Transactions from `develop`

kantai:
This does two things:

1. Merges changes and code improvements to the txbuild from `develop` (a lot of code de-duplication in txbuild from an added function `signInputs()` utility function, addition of a `buildIncomplete` option for returning skeletal TXs.)

2. Enables support for a separate bitcoin funder to the stacks transaction build.


lamontb4u:
I would like to unsubscribe. No more emails please.

src/network.js
```diff
@@ -62,8 +62,16 @@ export class BlockstackNetwork {
   }
 
   coerceAddress(address: string) {
-    const addressHash = bitcoinjs.address.fromBase58Check(address).hash
-    return bitcoinjs.address.toBase58Check(addressHash, this.layer1.pubKeyHash)
+    const { hash, version } = bitcoinjs.address.fromBase58Check(address)
```
jcnelson, line 65: Lol, I have a similar snippet of code in `cli-blockstack`.  Glad this is getting added to blockstack.js!

Do you want to throw an exception of the `version` isn't either `scriptHash` or `pubKeyHash`?

src/operations/txbuild.js
```diff
@@ -1071,84 +968,93 @@ function makeNameImport(name: string,
...
-        const signingTxB = fundTransaction(txB, senderAddress, utxos, feeRate, 0)
-        let signingPromise = Promise.resolve()
-        for (let i = 0; i < signingTxB.__tx.ins.length; i++) {
-          signingPromise = signingPromise.then(
-            () => senderKey.signTransaction(signingTxB, i))
-        }
-        return signingPromise.then(() => signingTxB)
-      }))
-    .then(signingTxB => signingTxB.build().toHex())
+  return Promise.all([senderKey.getAddress(), btcKey.getAddress()])
```
jcnelson, line 1037: This is a pretty significant change to this function, but I see no unit test code for it.  Can you add a unit test to check that a `TOKEN_TRANSFER` transaction can be successfully created with a different `btcFunderKeyIn` than the `senderKeyIn`?

jcnelson (changes requested):
Looks good overall.  Thanks for consolidating all the transaction-signing boilerplate!  Just left two comments, if you could address those first before merging.

src/operations/txbuild.js
kantai: Yep -- added unit test for this now.

src/network.js
kantai: Sure, it now throws if it cannot recognize the version number

kantai:
Thanks for the speedy review @jcnelson -- added a unit test for the token transaction funding change (was good it was added, too, as there was a bug in the funding calculation, where the funder would try to fund the owner's output, even when the owner's input can do that).

jcnelson (approved):

jcnelson:
Thanks!

jcnelson:
Feel free to merge whenever you're ready

kantai:
Merged! Thanks

PR #547, opened 2018-10-12
Feature/stacks transactions

jcnelson:
This feature branch adds support for generating Stacks transactions, and it adds support for querying the Blockstack Core node for account information.  It also adds the relevant unit tests.  Note that as with other operations, the Blockstack Core integration test framework tests for transaction format validity.


src/network.js
```diff
@@ -14,6 +16,12 @@ const TX_BROADCAST_SERVICE_ZONE_FILE_ENDPOINT = 'zone-file'
 const TX_BROADCAST_SERVICE_REGISTRATION_ENDPOINT = 'registration'
 const TX_BROADCAST_SERVICE_TX_ENDPOINT = 'transaction'
 
+export function hash160(buff: Buffer) {
```
kantai: This function is already defined in `operations/utils.js` -- we should probably just import it here.

larrysalibra:
@jcnelson what's the best way to go about testing this pull request?

src/network.js
jcnelson: Fixed and added a unit test.

jcnelson:
@larrysalibra you can do one of the following:

* Use it to issue transactions against the testnet (via the CLI, via a script you write, or via the node REPL)
* Run any/all of the Stacks-related integration tests in Blockstack Core -- they use the CLI and blockstack.js to generate and send transactions and query the state of accounts

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "blockstack",
-  "version": "18.0.4",
+  "version": "19.0.0",
```
larrysalibra, line 3: Do these changes break backwards compatibility or simply add additional functionality?

larrysalibra (commented):
Looked through the code and it looks good!

I also tested this with the CLI. 

I was able set up a wallet successfully. I was unable to send transactions because of testnet node issues:

`FetchError: request to http://testnet.blockstack.org:16269/v1/broadcast/registration failed, reason: connect ECONNREFUSED 13.65.207.163:16269`

I'll give it another try tomorrow to see if the node has been fixed.

larrysalibra (changes requested):
I'm still unable to broadcast transactions on testnet:

```
Vanquish:cli-blockstack larry$ blockstack-cli -t register larry123.id2 "$OWNER" "$PAYMENT" https://hub.blockstack.org
FetchError: request to http://testnet.blockstack.org:16269/v1/broadcast/registration failed, reason: connect ECONNREFUSED 13.65.207.163:16269
    at ClientRequest.<anonymous> (/Users/larry/git/blockstack.js/node_modules/cross-fetch/node_modules/node-fetch/lib/index.js:1393:11)
    at emitOne (events.js:116:13)
    at ClientRequest.emit (events.js:211:7)
    at Socket.socketErrorListener (_http_client.js:387:9)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at emitErrorNT (internal/streams/destroy.js:64:8)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
request to http://testnet.blockstack.org:16269/v1/broadcast/registration failed, reason: connect ECONNREFUSED 13.65.207.163:16269
Vanquish:cli-blockstack larry$
```

jcnelson:
The transaction broadcaster was silently failing on the last testnet update from a couple days ago due to running an outdated version of blockstack.js.  It should be running now.

kantai:
For help in testing other applications, I'm going to publish a pre-release of this on `npm` as:

```
18.1.0-alpha.2
```

Only incrementing the minor version, as this is a backwards compatible change.

Note:

Went with `.2` because I unpublished `alpha.1` -- accidently published it with the `latest` tag, meaning it would install by default. Cool `npm` feature/behavior to keep in mind for publishing a pre-release:

https://medium.com/@mbostock/prereleases-and-npm-e778fc5e2420

Basically, you can publish with `npm publish --tag _something_` and it won't be tagged as `latest`. Pretty neat.

larrysalibra (commented):
I tried out a number of transactions - all behaved as expected except for registering two names quickly in succession.   More below:

---

I tried to register two names in quick succession - the 2nd name registered successfully. The 1st did not. These are all using the master version of blockstack-cli updated to use blockstack.js 18.1.0-alpha.2.

I suspect that what's happening is that both name registrations use the same UTXOs, so only one gets confirmed. 

I would expect behavior to be the same as a bitcoin wallet: quick subsequent transactions use unconfirmed UXTOs as inputs. Users shouldn't have to wait until 1 name registration is confirmed before registering a new name.  I'm not sure if this behavior is possible and if it is, if the change needs to happen in blockstack.js or the CLI.  If this behavior is intended, then this branch is good to merge. 

Below are the steps I took (Verified behavior is reproducible with two rounds of these step:


1. blockstack-cli -t make_keychain

2. Send 10000000 stacks from testnet faucet:

```
Vanquish:cli-blockstack larry$ blockstack-cli -t balance ST6MT4JVC0DSFHKQ9KCA66MF1QEAVM130EATRZD9
{
  "STACKS": "10000000",
  "BTC": "5005500"
}
```

3. submit registration for larry123.id2 followed by registration for larry888.id2:
```
Vanquish:cli-blockstack larry$ blockstack-cli -t register larry123.id2 "$OWNER" "$PAYMENT" https://hub.blockstack.org 
{
  "profileUrls": {
    "error": null,
    "dataUrls": [
      "https://gaia.blockstack.org/hub/17QvCiPfWjDQahAHWTJ4EWf4qtQPdCyoiU/profile.json"
    ]
  },
  "txInfo": {
    "status": true,
    "message": "Preorder and Register queued for broadcast.",
    "preorderTxHash": "a154a4d906d54b35bdcbfe89f558d6923f42acdb6fc0f1fc932d32c3b96ce7d1",
    "registerTxHash": "b72b0ba7cd46c88950ac4208c7670a460d432f63bcf7a0c2ce13f2b21dbca86f"
  }
}
Vanquish:cli-blockstack larry$ blockstack-cli -t register larry888.id2 "$OWNER" "$PAYMENT" https://hub.blockstack.org 
{
  "profileUrls": {
    "error": null,
    "dataUrls": [
      "https://gaia.blockstack.org/hub/17QvCiPfWjDQahAHWTJ4EWf4qtQPdCyoiU/profile.json"
    ]
  },
  "txInfo": {
    "status": true,
    "message": "Preorder and Register queued for broadcast.",
    "preorderTxHash": "7a9de1a355195a2b9401406f80665f53cedb07329409edf6807369081950d3c7",
    "registerTxHash": "9f7a9415baaca20f3f6904237dee2f6b689425c3f088cd377f3028e694de412c"
  }
}
```

4. Make a new keychain and try to transfer larry123.id2 to a new owning `ID-` address: (note, `$OWNER` and `$PAYMENT` are still the same values as the first keychain since `$PAYMENT` has money and `$OWNER` is presumably the owner of larry123.id2.


```
Vanquish:cli-blockstack larry$ blockstack-cli -t transfer larry123.id2 ID-n47V4F9hFAxQDDrDz2X21EaKJdXU84VWAY true "$OWNER" "$PAYMENT"
(node:32543) UnhandledPromiseRejectionWarning: Error: Owner has no UTXOs for UPDATE.
    at addOwnerInput (/Users/larry/git/cli-blockstack/node_modules/blockstack/lib/operations/txbuild.js:38:11)
    at /Users/larry/git/cli-blockstack/node_modules/blockstack/lib/operations/txbuild.js:667:24
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
(node:32543) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:32543) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
{
  "status": false,
  "error": "Name cannot be safely transferred",
  "isNameValid": true,
  "ownsName": false,
  "addressCanReceiveName": true,
  "isInGracePeriod": false,
  "estimateCostBTC": 460500,
  "paymentBalanceBTC": 4388500
}
```

5. Checked testnet web page and found that only `larry888.id2` appears registered.


6. Verify with cli:

```
Vanquish:cli-blockstack larry$ blockstack-cli -t lookup larry888.id2
{
  "zonefile": null,
  "profile": null,
  "profileUrl": null
}
Vanquish:cli-blockstack larry$ blockstack-cli -t lookup larry123.id2
{
  "error": "Name not found"
}
Vanquish:cli-blockstack larry$
```

7. Try to register larry123.id2 again:

```
Vanquish:cli-blockstack larry$ blockstack-cli -t register larry123.id2 "$OWNER" "$PAYMENT" https://hub.blockstack.org 
{
  "profileUrls": {
    "error": null,
    "dataUrls": [
      "https://gaia.blockstack.org/hub/17QvCiPfWjDQahAHWTJ4EWf4qtQPdCyoiU/profile.json"
    ]
  },
  "txInfo": {
    "status": true,
    "message": "Preorder and Register queued for broadcast.",
    "preorderTxHash": "2c0577280d67715c5e7a15e450089167368ba2e7cd88f4c43b661806d94b48cc",
    "registerTxHash": "716fefdd7a3d328bba6207ea038d83ed22732d9075bd90cf7bc61edb9a986e75"
  }
}
```

8. Registration works this time:

```
Vanquish:cli-blockstack larry$ blockstack-cli -t lookup larry123.id2
{
  "zonefile": null,
  "profile": null,
  "profileUrl": null
}
```

larrysalibra (approved):

larrysalibra (changes requested):
Wanted to highlight something from the long comment above:

1. the errors shown when trying to transfer a name that doesn't exist are confusing 
`Error: Owner has no UTXOs for UPDATE.`  `"error": "Name cannot be safely transferred"` - this should say something like "you're trying to transfer a name that doesn't exist" (if that's the underlying problem)

jcnelson:
> I would expect behavior to be the same as a bitcoin wallet: quick subsequent transactions use unconfirmed UXTOs as inputs. Users shouldn't have to wait until 1 name registration is confirmed before registering a new name. I'm not sure if this behavior is possible and if it is, if the change needs to happen in blockstack.js or the CLI. If this behavior is intended, then this branch is good to merge.

This is the expected behavior since you were using the same OWNER and PAYMENT keys in both larry123.id2 and larry888.id2.  Taking a look at the bitcoin chain state on the testnet, I can see that the NAME_REGISTRATION transaction for larry123.id2 was never accepted by bitcoind.  Without having the actual transaction on-hand, the most likely explanation is that it double-spent an output consumed by an already-confirmed transaction from the same OWNER or PAYMENT key.

I can tell from the testnet log that the NAME_PREORDER for larry123.id2 was confirmed in the same block just before larry888.id.  The NAME_REGISTRATION for larry888.id2 was confirmed a few blocks later.  This suggests to me that the NAME_REGISTRATION for larry123.id2 consumed the same output as the NAME_REGISTRATION for larry888.id2.

The error you see on trying to transfer larry123.id2 is cryptic, and should have told you that larry123.id2 is not registered.

Users can register many names in rapid succession in the CLI, but they need to use different key pairs to do this.  The Browser naturally enforces this at the UI level by binding at most one name to a given ID-address.  The CLI *could* enforce this, but it does not.

kantai:
> This is the expected behavior since you were using the same OWNER and PAYMENT keys in both larry123.id2 and larry888.id2. Taking a look at the bitcoin chain state on the testnet, I can see that the NAME_REGISTRATION transaction for larry123.id2 was never accepted by bitcoind. Without having the actual transaction on-hand, the most likely explanation is that it double-spent an output consumed by an already-confirmed transaction from the same OWNER or PAYMENT key.

Yeah -- basically this happens because the REGISTER transaction will spend the PREORDER's output, however, we wait a little bit before broadcasting the REGISTER. This means that a following transaction could spend that ouput first. This can be remedied in the blockstack.js transaction library by manually tracking spent UTXOs (and this is done using the `network.modifyUTXOSetFrom()` function). This is called in the Browser to try and handle this correctly, but really, we should figure out a better way to handle this situation. I think it's pretty out-of-scope of this PR (as its already the behavior in the transaction library).

larrysalibra:
Cool - thanks for the info @jcnelson and @kantai - makes sense to me!

jcnelson:
We good to merge then?

larrysalibra (approved):

larrysalibra:
Yes! Good from my side.

PR #549, opened 2018-10-16
Two transaction related fixes

kantai:
1. Per #538, we try to set the transaction version to version 1. This was _missed_ in bitcoin spends. This PR fixes that.
2. For the bitcoind network interface (used in dev environments, mostly), we call `importaddress` on every UTXO fetch (this is because bitcoind only tracks imported address' UTXOs). However, we can be a bit more intelligent (and get better behavior out of the endpoint) by only doing this on the first such request. This PR implements that behavior.

A unit test was added for the above bitcoind behavior.


larrysalibra (approved):
Works for me!

PR #552, opened 2018-10-19
Add option to pass Content-Type to putFile API

### LINKED ISSUE #551: Option to pass Content-Type to putFile API
The current implementation sets Content-Type header to either 'text/plain' or 'application/octet-stream' based on the type of the input parameter `content`.
For example, if you want to store a publicly accessible html file, the current implementation stores it with 'text/plain' header. And if you access the public read URL from a web browser, the HTML will be interpreted as plain text.
So it would be helpful to have another option in  `options` parameter to pass the Content-Type header.

muneebm:
Changes to resolve #551 
putFile API: added a third option `contentType` in `options` parameter.
Added unit test: 'putFile & getFile unencrypted, not signed, with contentType'.

### DIFF
--- a/src/storage/index.js
+++ b/src/storage/index.js
@@ -362,23 +362,26 @@ export function getFile(path: string, options?: {
  *                                                  or the provided public key
  * @param {Boolean} [options.sign=false] - sign the data using ECDSA on SHA256 hashes with
  *                                         the app private key
+ * @param {String} [options.contentType=''] - set a Content-Type header for unencrypted data
  * @return {Promise} that resolves if the operation succeed and rejects
  * if it failed
  */
 export function putFile(path: string, content: string | Buffer, options?: {
   encrypt?: boolean | string,
-  sign?: boolean
+  sign?: boolean,
+  contentType?: string
   }) {
   const defaults = {
     encrypt: true,
-    sign: false
+    sign: false,
+    contentType: ''
   }
 
   const opt = Object.assign({}, defaults, options)
 
-  let contentType = 'text/plain'
-  if (typeof (content) !== 'string') {
-    contentType = 'application/octet-stream'
+  let contentType = opt.contentType
+  if (!contentType) {
+    contentType = (typeof (content) === 'string') ? 'text/plain' : 'application/octet-stream'
   }
 
   // First, let's figure out if we need to get public/private keys,
--- a/src/storage/index.js
+++ b/src/storage/index.js
@@ -379,7 +379,7 @@ export function putFile(path: string, content: string | Buffer, options?: {
 
   const opt = Object.assign({}, defaults, options)
 
-  let contentType = opt.contentType
+  let { contentType } = opt
   if (!contentType) {
     contentType = (typeof (content) === 'string') ? 'text/plain' : 'application/octet-stream'
   }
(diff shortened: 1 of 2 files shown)


tests/unitTests/src/unitTestsStorage.js
```diff
@@ -281,6 +281,50 @@ export function runStorageTests() {
...
+    FetchMock.get(fullReadUrl, config)
+
+    const options = { encrypt: false, contentType: 'text/html' }
+    putFile(path, fileContent, options)
+      .then((publicURL) => {
+        t.ok(publicURL, fullReadUrl)
+      })
+      .then(() => {
+        const decryptOptions = { decrypt: false }
+        getFile(path, decryptOptions).then((readContent) => {
```
jcnelson, line 323: Can you confirm in this test that the `content-type` header is honored correctly on getFile?  You can check by looking at the type of object returned -- it will be a string for `text/*` and `application/json`, or an `ArrayBuffer` for anything else.

jcnelson:
Thanks for sending this PR @muneebm!  I just had one comment.

src/storage/index.js
```diff
@@ -362,23 +362,26 @@ export function getFile(path: string, options?: {
...
+    sign: false,
+    contentType: ''
   }
 
   const opt = Object.assign({}, defaults, options)
 
-  let contentType = 'text/plain'
-  if (typeof (content) !== 'string') {
-    contentType = 'application/octet-stream'
+  let contentType = opt.contentType
```
hstove: Small nitpick, but we can use object destructuring to clean this up, so it would be:

~~~javascript
let { contentType } = opt;
~~~

muneebm: Thanks Hank. Good catch!

tests/unitTests/src/unitTestsStorage.js
muneebm: Thanks for the review, Jude. The tape assert `t.equal(readContent, fileContent)` was already doing this check. But I added another assert to make it more obvious.

hstove:
Looks good to me! Thanks for this.

larrysalibra (approved):
I tried this by linking it to blockstack-todos and changing the `putFile` code to the following:
```
return blockstack.putFile('test.html', '<html><head></head><body>hello</body></html>', { contentType: 'text/html', encrypt: false })
        .then((url) => console.log(url))
```

I added a todo item which triggered the put file and printed the read url to the console.

Checking the read url in curl, we can see that the `content-type` header is now set correctly.

```
curl: try 'curl --help' or 'curl --manual' for more information
Vanquish:blockstack-todos larry$ curl -I https://gaia.blockstack.org/hub/14DtA8xANeWbPKTiuHg5kkwoVhCw2Lok7x/test.html
HTTP/2 200 
accept-ranges: bytes
access-control-allow-origin: *
cache-control: public, max-age=1
content-md5: Th2N7pJ/hDNmdF5ciYScjQ==
content-type: text/html
date: Sat, 20 Oct 2018 14:08:48 GMT
etag: 0x8D6369584C09329
last-modified: Sat, 20 Oct 2018 14:08:32 GMT
server: Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0
vary: Origin
x-cache: HIT
x-ms-blob-type: BlockBlob
x-ms-lease-status: unlocked
x-ms-request-id: e519cea0-e01e-0068-657e-68e450000000
x-ms-version: 2009-09-19
content-length: 44

Vanquish:blockstack-todos larry$
```

Also verified that the web page written by the command displays properly in my browser:

screen shot 2018-10-20 at 10 12 01 pm


Thanks so much for this @muneebm!

I'll update the change log with a note once I merge this.

PR #554, opened 2018-10-22
Add missing new for error constructor

### LINKED ISSUE #553: [Storage Strategies] loadUserData invalid state
This should be 
`throw new InvalidStateError...`

https://github.com/blockstack/blockstack.js/blob/3a6249bcbcbaa153e892f5ef7ed8737d91c2a81e/src/auth/authApp.js#L221

friedger:
Fixes #553


larrysalibra (approved):
thanks for this @friedger

Issue #234, opened 2017-07-17
Add public API method that returns direct URL to file


larrysalibra:
Request from @icidasset in slack:

```getFileUrl("gaia_path_to_some_file.mp4")
-> http://some_url_for_that_file.mp4
```

This would enable applications to progressively read the file via [http range requests](https://tools.ietf.org/html/rfc7233).

You usually don't want to download very large files in one go, but instead download chunks of them. An example of this is a media playing app that can start play the music or video after the first portion has been received.

A challenge of this is that our current storage driver dropbox generates URLs without the appropriate CORS headers, so we need to proxy requests to these URLs.

larrysalibra:
This should be pretty straight forward to implement now that we've moved to gaia hubs. We already have a function that does this: `getFullReadUrl`: https://github.com/blockstack/blockstack.js/blob/3e4689685235a7ec36a617e5cc4a80ea88ad3f7c/src/storage/hub.js#L37

We could wrap it in a function with a more developer friendly API like `getFileUrl` and the same API options as `getFile`.

PR #556, opened 2018-10-25
Add `getFileUrl()`

kantai:
Got distracted by an old issue report #234 and decided to implement.

This adds a `getFileUrl()` function with roughly the same API as `getFile()` (minus the encryption, verification options).

```javascript
/**
 * Get the URL for reading a file from an app's data store.
 * @param {String} path - the path to the file to read
 * @param {Object} [options=null] - options object
 * @param {String} options.username - the Blockstack ID to lookup for multi-player storage
 * @param {String} options.app - the app to lookup for multi-player storage -
 * defaults to current origin
 * @param {String} [options.zoneFileLookupURL=null] - The URL
 * to use for zonefile lookup. If falsey, this will use the
 * blockstack.js's getNameInfo function instead.
 * @returns {Promise} that resolves to the URL or rejects with an error
 */
```

This _also_ refactors `getFileContents` (an internal function used by `getFile`) to use this function, which means that the tests for `getFile` now cover `getFileUrl` (so I didn't add tests).


jcnelson:
Cool!  Does this also mean that you don't need to use `window` at all anymore in node.js?  i.e. does this also address https://github.com/blockstack/blockstack.js/issues/500?

kantai:
Sort of -- only multi-player `getFile` will work without the `window` at all (everything else like `putFile` and single-player `getFile` tries to use userSession state, which would need to be refactored for non-browser contexts)

So this will work:
```javascript
const bsk = require('blockstack')
bsk.getFile('statuses.json', { username: 'collectivism.id', app: "https://app.travelstack.club", decrypt: false }).then(console.log).catch(console.log)
```

x5engine:
>  You don't need to use window at all anymore in node.js?

The dream

hstove:
This looks good to me - just need to resolve changelogs conflicts.

friedger:
@kantai Can we have this merged so I can add it to the Android SDK?

kantai:
Resolved changelog conflicts, and updated reviewer to the new maintainer of blockstack.js

yknl (approved):
This looks good to me 👍

zone117x (approved):

PR #557, opened 2018-10-25
Fix json post format in transaction broadcast for insight API.

yknl:
This fixes the POST request data format when broadcasting to insight API. (Not currently being used for any production Blockstack code.)


hstove (approved):
Since there are no tests for this (I understand why), I'm approving this based on reading the Insight API code here:

https://github.com/bitpay/insight-api/blob/master/lib/transactions.js#L331

zone117x (approved):
Approving with @hstove's reason

PR #559, opened 2018-10-30
Feature/tx magic variable

jcnelson:
This PR adds `network.getMagicBytes()` and `network.setMagicBytes()`, which lets the user override the magic bytes used to identify Blockstack transactions.  The purpose of this pull request is to enable the user to interact with one or more Blockstack testnets, realized as alternative virtualchains on top of Bitcoin mainnet.  This lets us test other pieces of Blockstack software against Bitcoin mainnet without having to either deal with either interacting with the (highly-prone-to-reorg) Bitcoin testnet or having to run a Bitcoin regtest node.  In particular, when used in conjunction with Blockstack Core and the CLI, this lets us instantiate an alternative Blockstack network so we can test hardware wallets with fake Stacks (but real Bitcoin).

### DIFF
--- a/src/network.js
+++ b/src/network.js
@@ -49,6 +49,8 @@ export class BlockstackNetwork {
 
   btc: BitcoinNetwork
 
+  MAGIC_BYTES: string
+
   constructor(apiUrl: string, broadcastServiceUrl: string,
               bitcoinAPI: BitcoinNetwork,
               network: Object = bitcoinjs.networks.bitcoin) {
@@ -60,6 +62,7 @@ export class BlockstackNetwork {
     this.DUST_MINIMUM = 5500
     this.includeUtxoMap = {}
     this.excludeUtxoSet = []
+    this.MAGIC_BYTES = 'id'
   }
 
   coerceAddress(address: string) {
@@ -81,6 +84,17 @@ export class BlockstackNetwork {
 
   getDefaultBurnAddress() {
     return this.coerceAddress('1111111111111111111114oLvT2')
+  }
+
+  getMagicBytes(): string {
+    return this.MAGIC_BYTES
+  }
+
+  setMagicBytes(newMagic: string) {
+    if (newMagic.length !== 2) {
+      throw new Error('Invalid magic bytes')
+    }
+    this.MAGIC_BYTES = newMagic
   }
 
   /**
--- a/src/network.js
+++ b/src/network.js
@@ -86,17 +86,6 @@ export class BlockstackNetwork {
     return this.coerceAddress('1111111111111111111114oLvT2')
   }
 
-  getMagicBytes(): string {
-    return this.MAGIC_BYTES
-  }
-
-  setMagicBytes(newMagic: string) {
-    if (newMagic.length !== 2) {
-      throw new Error('Invalid magic bytes')
-    }
-    this.MAGIC_BYTES = newMagic
-  }
-
   /**
    * Get the price of a name via the legacy /v1/prices API endpoint.
    * @param {String} fullyQualifiedName the name to query
--- a/src/operations/skeletons.js
+++ b/src/operations/skeletons.js
@@ -152,7 +152,7 @@ function makeTXbuilder() {
 
 function opEncode(opcode: string): string {
   // NOTE: must *always* a 3-character string
-  const res = `${config.network.getMagicBytes()}${opcode}`
+  const res = `${config.network.MAGIC_BYTES}${opcode}`
   if (res.length !== 3) {
     throw new Error('Runtime error: invalid MAGIC_BYTES')
   }
--- a/src/operations/skeletons.js
+++ b/src/operations/skeletons.js
@@ -150,6 +150,15 @@ function makeTXbuilder() {
   return txb
 }
 
+function opEncode(opcode: string): string {
+  // NOTE: must *always* a 3-character string
+  const res = `${config.network.getMagicBytes()}${opcode}`
+  if (res.length !== 3) {
+    throw new Error('Runtime error: invalid MAGIC_BYTES')
+  }
+  return res
+}
+
 export function makePreorderSkeleton(
   fullyQualifiedName: string, consensusHash : string, preorderAddress: string,
   burnAddress : string, burn: AmountType,
@@ -188,7 +197,7 @@ export function makePreorderSkeleton(
 
   const opReturnBufferLen = burnAmount.units === 'BTC' ? 39 : 66
   const opReturnBuffer = Buffer.alloc(opReturnBufferLen)
-  opReturnBuffer.write('id?', 0, 3, 'ascii')
+  opReturnBuffer.write(opEncode('?'), 0, 3, 'ascii')
   hashed.copy(opReturnBuffer, 3)
   opReturnBuffer.write(consensusHash, 23, 16, 'hex')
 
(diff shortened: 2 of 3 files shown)


src/network.js
```diff
@@ -83,6 +86,17 @@ export class BlockstackNetwork {
     return this.coerceAddress('1111111111111111111114oLvT2')
   }
 
+  getMagicBytes(): string {
```
kantai: We don't really use getters/setters elsewhere in this object -- so I think it'd be preferable to just use the MAGIC_BYTES field directly.

kantai (changes requested):
This all looks great to me, and worked when I tried it -- the only change I'd recommend is getting rid of the getter/setter for MAGIC_BYTES. We don't really use them elsewhere in `blockstack.js`, and it'd be good to have a uniformity.

kantai:
For those looking to also test via `blockstack.js` directly. This is what I did:

```javascript
var bsk = require('.')
var bigi = require('bigi')


bsk.config.network.MAGIC_BYTES = 'di'
// this URL might be different for you...
bsk.config.network.blockstackAPIUrl = 'http://172.17.0.2:16268'

var sk = bsk.makeECPrivateKey()

var sender = bsk.publicKeyToAddress(bsk.getPublicKeyFromPrivate(sk))

// add sender to your testnet genesis block.
//  also, send BTC to it (to fund TX fee)

var recipientSK = bsk.makeECPrivateKey()

var recipient = bsk.publicKeyToAddress(bsk.getPublicKeyFromPrivate(recipientSK))


bsk.transactions.makeTokenTransfer(recipient, 'STACKS', bigi.valueOf(1000), 'TEST!' , sk + '01')
.then(console.log)

// You should see a TX! Broadcast it and wait until your node processes it.
```

kantai:
Woot!

```
$ docker logs kind_clarke | grep ACCEPT
[2018-11-01 20:35:15,356] [DEBUG] [namedb:483] (7.140604945065792) ACCEPT TOKEN_TRANSFER at (548345, 780) data: address='17zg4aXiDA4ptBEXAAfHJnDqVqEjGLE6Rw', block_id='548345', consensus_hash='0efa29f955c6ae3bb5037039d89dba5e', op='$', recipient='76a914cfd25e09f2d33e1aec73bfcc5b608ec513bbe6c088ac', recipient_address='1KwrtHhRM3GHgeTQFcabbZ43aZc7judLUY', scratch_area='5445535421', sender='76a9144cb912533a6935880df7647fd5232e40aca07b8088ac', token_fee='1000', token_units='STACKS', txid='13bb0fdc939b85b4476f065dcac28abb6ce1f395f03243e1e23dd79904db6c8a', vtxindex='780'
```

jcnelson:
Thanks for testing @kantai!  I made your requested change.

kantai (approved):

jcnelson:
Let me know if I can merge it?  Still waiting for @larrysalibra

yknl:
Any chance we can get this merged in soon for testing before the wallet release? Have trouble running this via npm linking.

larrysalibra (commented):
Apologies for the delay in getting to this.

The looks good to me from reading the code. It's pretty clear what it does and how it does it.

I started going down the rabbit hole of testing this against https://testnet.blockstack.org, but had trouble generating an appropriate stacks address that the faucet would accept.

I modified these lines of @kantai 's test script to use the normal magic bytes and the appropriate api url:

```JavaScript
bsk.config.network.MAGIC_BYTES = 'id'
// this URL might be different for you...
bsk.config.network.blockstackAPIUrl = 'http://testnet.blockstack.org:16268'
```

I modified the [publicKeyToAddress function in blockstack JS](https://github.com/blockstack/blockstack.js/blob/c9d25ab78cd4f634372e7f0e9d35ebb6c083e7c7/src/keys.js#L17) to use the testnet version 111. 

I opened a related bug noting that we should make sure all address related functions handle different network versions: https://github.com/blockstack/blockstack.js/issues/563


I installed c32check using `npm install c32check` and tried converted the sender address to a stacks address.

```JavaScript
var c32check = require('c32check')
var sk = bsk.makeECPrivateKey()

var sender = bsk.publicKeyToAddress(bsk.getPublicKeyFromPrivate(sk))

var senderStacksAddress = c32check.b58ToC32(sender)

// fund senderStacksAddress from https://testnet.blockstack.org 
```

Entering the resulting address ( the b58check version of which began with an `m`) into the faucet box resulted in a "Failed to convert XXX to a Stacks address" error.

I spent some time digging through the CLI code to try to figure out if I was using an incorrect version or generating the address improperly, but ran out of time.

@jcnelson please let me know what I was doing wrong - either way I think this PR is ready to merge.

kantai:
> Entering the resulting address ( the b58check version of which began with an m) into the faucet box resulted in a "Failed to convert XXX to a Stacks address" error.

I believe the testnet faucet wants to be provided with a b58 address, not the c32 address. @jcnelson can correct me if I'm wrong there.

> I started going down the rabbit hole of testing this against https://testnet.blockstack.org, but had trouble generating an appropriate stacks address that the faucet would accept.

This PR implements a different kind of testnet -- the "testnet" at testnet.blockstack.org is a blockstack chain running on a regtest bitcoin chain. This PR allows you to submit transactions with a different magic header, which allows you to run a testnet on bitcoin _mainnet_ -- this, in turn, allows you to test things like Stack transfers with "testing Stacks" but with the real Bitcoin chain.

larrysalibra:
> I believe the testnet faucet wants to be provided with a b58 address, not the c32 address. @jcnelson can correct me if I'm wrong there.

The stacks testnet faucet accepts c32 addresses generated by the `blockstack-cli -t make_keychain` command.

> This PR implements a different kind of testnet -- the "testnet" at testnet.blockstack.org is a blockstack chain running on a regtest bitcoin chain. This PR allows you to submit transactions with a different magic header, which allows you to run a testnet on bitcoin mainnet -- this, in turn, allows you to test things like Stack transfers with "testing Stacks" but with the real Bitcoin chain.

I understand that - the "testnet on regtest" system at testnet.blockstack.org uses `id` as the magic header, so I was trying to set the magic header back to `id` and generate appropriate addresses and then a transaction that would be accepted on that network using this code.

In general, the config.network structure with regards to layer 1 mainnet vs testnet vs regtest and layer 2 mainnet vs testnet is a bit confusing.

We define two constants
LOCAL_REGTEST and MAINNET_DEFAULT that don't make it clear if they're referring to layer 1 or layer 2.

https://github.com/blockstack/blockstack.js/blob/28f44a0531fa315be7b7c00560fc52d7fed84d51/src/network.js#L1190

It seems the names refer to the layer 1 networks, but the settings contained within them also refer to layer 2 components as well.

jcnelson:
> Entering the resulting address ( the b58check version of which began with an m) into the faucet box resulted in a "Failed to convert XXX to a Stacks address" error.

I'm getting this too.  Will fix.

But nevertheless, you won't be able to test this code on testnet.blockstack.org since testnet.blockstack.org uses the `id` magic.  Its Core node won't pick up any transactions with other magic bytes.

> In general, the config.network structure with regards to layer 1 mainnet vs testnet vs regtest and layer 2 mainnet vs testnet is a bit confusing.

> We define two constants LOCAL_REGTEST and MAINNET_DEFAULT that don't make it clear if they're referring to layer 1 or layer 2.

They both refer to layer 1.  Changes to MAGIC_BYTES affect layer 2.  There are now four classes of network configurations -- `id` on BItcoin mainnet, non-`id` on Bitcoin mainnet, `id` on Bitcoin testnet/regtest, and non-`id` on Bitcoin testnet/regtest.

larrysalibra:
> But nevertheless, you won't be able to test this code on testnet.blockstack.org since testnet.blockstack.org uses the id magic. Its Core node won't pick up any transactions with other magic bytes.

I tried to set the magic bytes code to the default - `id`- shouldn't that work?

```
bsk.config.network.MAGIC_BYTES = 'id'
// this URL might be different for you...
bsk.config.network.blockstackAPIUrl = 'http://testnet.blockstack.org:16268'
```

jcnelson:
Yes, that should work.

jcnelson:
Going to go ahead and merge.  Thanks folks!

PR #561, opened 2018-11-02
Return correct file count

friedger:
This PR
* fixes a wrong file count in listFiles

Call listFiles for an account with a small number of files (e.g. 4). Then currently 0 is returned.


jcnelson:
Looks good to me.  Thanks @friedger !

PR #579, opened 2018-12-11
Fix LinkedIn proof

### LINKED ISSUE #550: validateProofs broken for LinkedIn
`LinkedIn.getProofIdentity` is broken.

Specifically, this line is no longer valid: `var profileLink = $('article').find('.post-meta__profile-link');`.
Seems like LinkedIn has updated.

shreyasthiagaraj:
Fixes #550. LinkedIn had updated their page format so our scraping logic was broken. Works now.

### DIFF
--- a/src/profiles/services/linkedIn.js
+++ b/src/profiles/services/linkedIn.js
@@ -32,13 +32,13 @@ class LinkedIn extends Service {
 
   static getProofIdentity(searchText: string) {
     const $ = cheerio.load(searchText)
-    const profileLink = $('article').find('.post-meta__profile-link')
+    const profileLink = $('.author-profile__view-profile')
 
     if (profileLink !== undefined) {
       if (profileLink.attr('href') === undefined) {
         return ''
       }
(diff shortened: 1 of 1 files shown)


zone117x:
Is there a reason we have to parse and query the DOM, as opposed to just returning the whole response string? Seems less likely for our code to break when the social websites update.

hstove:
> Is there a reason we have to parse and query the DOM, as opposed to just returning the whole response string? Seems less likely for our code to break when the social websites update.

Not parsing the DOM would probably serve the same result, while being less brittle. I'd support that 👍

zone117x:
@kantai just brought up that since other people can comment on posts it could cause problems. 
Maybe some of the services don't show/allow comments, or provide an embedded-share-style endpoint that only shows the OP.

yknl:
This seems to be outdated again due to LinkedIn changes. @zone117x's [proposed solution ](https://github.com/blockstack/blockstack-browser/issues/1857) might help fix this.

PR #583, opened 2018-12-14
make estimateTokenTransfer private

friedger:
estimateTokenTransfer currently appears in the js docs.

This PR add `@private` in the js docs for estimateTokenTransfer


kantai:
Thanks for catching this @friedger -- merging.

PR #591, opened 2019-01-06
[Review] Prototype for new protocol detection method

zone117x:
I think we should prioritize the auth problems due to the protocol handler. I'm requesting several reviewers so we can validate or invalidate this new technique. 

This technique uses a redirect "ping" of sorts to detect the protocol handler in a way not dependent on browser quirks and user-agent detection. Compared to the existing method, this implementation errs on the side of caution..
* False-positive detections are treated as unacceptable. Web app completely broken when user cannot login because we incorrectly assumed a protocol handler. 
* False-negatives should be rare, but are acceptable. The web app may be redirect to https auth when the protocol handler worked. These do **not** break the auth flow but rather slightly compromise the UX.

### How it works:
1. Web app `tab A` performs a protocol redirect for the propose of a ping (rather than auth). Example: `blockstack://echo?redirectURL=https://testapp.com/echoReply?requestID=12345`
2. Web app `tab A` starts a 1000ms timer waiting for the echo reply. Side note: a couple browser APIs related to DOM window focus & visibility are leveraged to detect *probable* protocol handler capability, and increase the timeout to 2-4 seconds. 
3. If the local auth app is installed, it is launched and immediately opens the ping-echo url for the web app that initiated it. As in, it opens web app `tab B` using the query param `echoReply=[requestID]`.
4. When `tab B` is loaded it immediately leverages [cross-window communication](http://www.dannyc.xyz/blog/cross-tab-and-window-communication.html) to tell `tab A` that the protocol handler was launched and is working. In most cases, this all occurs within the 1-4 second timeout, and the originating tab knows not to redirect itself to the https auth. 
5. `Tab B` then redirects *back* to the localhost/native auth flow. 
6. `Tab A` either: 
   1. Acknowledges `tab B`'s echo reply which indicates the protocol handler is working, and does not redirect to https web auth. 
   2. OR it does not get the echo reply within the timeout period (1000ms) and it redirects the page to the https web auth. 

### Testing out the prototype:
* Demo site for testing: https://zone117x.github.io/blockstack-protocol-ping-demo/public/ (NOTE: requires using with the change to the Blockstack Browser app, see below)
* Build of Blockstack Browser prototype for testing: https://github.com/blockstack/blockstack-browser/pull/1774 (ctrl+click the blockstack tray-icon and enable development mode, pull the [branch](https://github.com/blockstack/blockstack-browser/tree/feature/new-protocol-detection), run `npm run dev`

This approach requires a minor change to the auth browser (Blockstack Browser), see https://github.com/blockstack/blockstack-browser/commit/3cfb20426ea52997a88f978b78e35786d54683d7

### Improvements and fixes:
* Should work on all browsers and unlikely to be broken by any given browser.
* Prevents the browser prompt from Safari and others about unknown protocol when auth-browser is not installed by using an iframe technique rather than window.location to launch the protocol uri.
* Fixes the bug where the Blockstack Browser app is uninstalled but the browser / OS still has a protocol handler registered. See https://github.com/blockstack/blockstack.js/issues/571#issuecomment-451936372
* Lessens the severity of the bug where using your non-default web browser causes the default browser to be opened and signed in. With this approach the default browser will still be opened, but the original browser will notice the failure and redirect to https auth rather than do nothing and prevent auth. 
* In theory this should work on mobile as well. No more browser user-agent detection code. 

### Downsides:
* Requires a minor wait/timeout before redirecting to the http/web auth.
* The window redirection dance is a bit odd.

### Good & bad: 
If the native app is installed and opens, but something prevents the ping reply to the web app from occurring within the timeout (like a slow computer), then the original tab will end up with a false-negative (the protocol handler is not detected), and will redirect to the https auth. 

I think this could also be considered good behavior that indicates the resilience of this approach, because in the situation of some unknown failure condition, the user auth flow will still be able to progress through the https auth, as opposed to the web app doing nothing upon login-click. 

And in the case of this false-positive (where the redirect to https auth even though the protocol handler did actually launch), that original tab was useless anyway with the existing approach which leaves that web app open and unauthenticated. 

### Backwards compatibility / upgrade path

This approach does not require a blockstack.js major semver bump or any code usage changes from app devs. It should also not break auth flow when blockstack.js is updated and the auth-browser is outdated or vice versa. It does, however, require both an updated blockstack.js and auth-browser for users experience the improvements offered. 

### Reference issues:
* https://forum.blockstack.org/t/proposed-solutions-for-desktop-safari-firefox-sign-in/6713
* https://github.com/blockstack/blockstack.js/issues/558
* https://github.com/blockstack/blockstack.js/issues/571
* https://github.com/blockstack/blockstack.js/issues/488
* https://github.com/blockstack/blockstack.js/issues/298
* https://github.com/blockstack/blockstack-browser/issues/1759
* https://github.com/blockstack/blockstack-browser/issues/172
* https://github.com/blockstack/blockstack-browser/issues/1689
* https://github.com/blockstack/blockstack-browser/issues/1669
* https://github.com/blockstack/blockstack-browser/issues/1120
* https://github.com/blockstack/blockstack-browser/issues/1102
* https://github.com/blockstack/blockstack-browser/issues/314


yknl:
This is a great solution to the protocol handler problems we've been experiencing, but I think there are a few UX issues that we should resolve before moving forward.

When I tested this out on Safari, the open protocol handler prompt appeared, but if I don't click on "Allow" immediately, the app begins to load browser.blockstack.org in the background. I think we need a longer timeout. 

Regarding implementing [Larry's idea #2](https://forum.blockstack.org/t/proposed-solutions-for-desktop-safari-firefox-sign-in/6713?u=zone117x), it doesn't improve UX for 90% of the use cases. Because once a user is logged into an app, they're unlikely to have to log in again for a while. We would only need to trigger the auth flow for new apps that the user hasn't logged into before and which we don't have a preference for.

I think one potential enhancement would be to show a pop up indicating that the user is being redirected, and if it didn't work, they can click a button to go to the web app browser. However this means that we have to bundle UI into blockstack.js.

hstove:
> I think one potential enhancement would be to show a pop up indicating that the user is being redirected, and if it didn't work, they can click a button to go to the web app browser. However this means that we have to bundle UI into blockstack.js.

Matt discussed this in the forum, but we could inject an iframe with the UI. It'd be very little javascript required. Then we could also update it without apps having to update blockstack.js.

zone117x:
EDIT: The UX issue discussed below has been solved in this PR. 

-------

@yknl thanks for looking this over.

> I think there are a few UX issues that we should resolve before moving forward ... When I tested this out on Safari, the open protocol handler prompt appeared, but if I don't click on "Allow" immediately, the app begins to load browser.blockstack.org in the background. 

I just discovered this issue but worse. Safari creates some background page noise (minor inconvenience but not a deal breaker), however, all chromium-based browsers will auto-close the prompt if the page is redirected, which is definitely a deal-breaker. I outlined two solutions to this in the forum post I just made https://forum.blockstack.org/t/proposed-solutions-for-desktop-safari-firefox-sign-in/6713/9

> I think one potential enhancement would be to show a pop up indicating that the user is being redirected, and if it didn't work, they can click a button to go to the web app browser. However this means that we have to bundle UI into blockstack.js.

This is my favorite approach at the moment, which I described in the forum post - this new detection method combined the with the choice modal. 

I think the iframe for the https login modal (that @hstove just mentioned) is even better UX, but it comes at the cost of some minor security issues like user url src validation and click-jacking. However, at the moment I don't think we are doing anything on our https auth endpoint to prevent these types of attacks anyway, like [x-frame-options](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options), [Content-Security-Policy frame-ancestors](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors) or [iframe-busting](https://stackoverflow.com/a/959036/794962).

@yknl @hstove Do you think we can move forward with either of these fixes (choice modal or iframe)? If so I'll update the prototype for additional review. If we are not concerned about the potential iframe security issues then that would be quick and easy to implement.

zone117x:
Update: figured out a way to leverage a couple js dom APIs that behave consistently and reliably across browsers to fix the browser prompt auto-dismissal bug with this new protocol detection method. So far extensive testing on macOS with Chrome, Firefox, Safari, Opera, and Brave show it working really well. Opera shows an error prompt but works fine after dismissing. Working on testing Edge.

These methods are a continuation of the graceful-fallback / probability based approach - where all the techniques / APIs are currently working well in all tested browsers, but if something breaks from some future browser update or edge case, then the https auth redirect will *always* still be triggered.

hstove:
One behavior on Brave right now is that when I use the demo site, it redirects to localhost:8888, but in the original tab, it redirects to browser.blockstack.org. Is that expected behavior? I have no prompt for the protocol handle because it's set to always redirect.

zone117x:
> One behavior on Brave right now is that when I use the demo site, it redirects to localhost:8888, but in the original tab, it redirects to browser.blockstack.org. Is that expected behavior? I have no prompt for the protocol handle because it's set to always redirect.

Sounds like the auth-browser branch isn't being ran on your machine?

hstove:
> Sounds like the auth-browser branch isn't being ran on your machine?

Oh yeah, whoops!

zone117x:
Completed testing with Edge browser. This new method works in Edge but has a minor UX imperfection - when the native app is not installed, the browser will ask the user if they want to "look for an app in Microsoft app store". The web app **does** detect that a protocol handler is not installed, and **does** redirect them to the https auth. The user simply has to dismiss the prompt. 
When the native app is installed, it works without issue. This seems like an acceptable trade off.

src/auth/authApp.js
```diff
@@ -84,23 +82,123 @@ export function redirectToSignInWithAuthRequest(authRequest: string = makeAuthRe
     return
   }
 
-  function successCallback() {
-    Logger.info('protocol handler detected')
-    // protocolCheck should open the link for us
+  // Create a unique ID used for this protocol detection attempt.
+  const echoReplyID = Math.random().toString(36).substr(2, 9)
```
jcnelson, line 77: Is this supposed to be cryptographically secure?  What happens if someone (e.g. an attacker) is able to guess this value?  Can they hijack the redirect?

src/auth/authApp.js
```diff
@@ -84,23 +82,123 @@ export function redirectToSignInWithAuthRequest(authRequest: string = makeAuthRe
...
+  // next browser window to check.
+  window.localStorage.setItem(echoReplyKey, 'pending')
+  const cleanUpLocalStorage = () => {
+    window.localStorage.removeItem(echoReplyKey)
   }
 
-  function failCallback() {
-    Logger.warn('protocol handler not detected')
-    window.location = httpsURI
+  const detectionTimeout = 1000
```
jcnelson, line 104: Is this long enough?  What happens if I do this on an underpowered mobile phone requesting the desktop site?

src/auth/authApp.js
zone117x: No, it's supposed to create a unique "auth" session in the situation that the user clicks "sign in with blockstack" on the same web app in several different tabs - or some other situation similar to that. This is the unique ID stored in localStorage for the duration of the protocol detection "ping".

src/auth/authApp.js
zone117x: I've been calling these situations "false-negative" detections. On mobile a false redirect to https auth is largely inconsequential. The native mobile protocol handler app will have already launched.

Due to the nature of this new detection approach, these are unavoidable but we aim to make them unlikely and also not that bad when they do occur. 

See the initial PR description for more info
> False-positive detections are treated as unacceptable. Web app completely broken when user cannot login because we incorrectly assumed a protocol handler.
> False-negatives should be rare, but are acceptable. The web app may be redirect to https auth when the protocol handler worked.

src/auth/authApp.js
```diff
@@ -84,23 +82,123 @@ export function redirectToSignInWithAuthRequest(authRequest: string = makeAuthRe
...
-    // protocolCheck should open the link for us
+  // Create a unique ID used for this protocol detection attempt.
+  const echoReplyID = Math.random().toString(36).substr(2, 9)
+  const echoReplyKey = `echo-reply-${echoReplyID}`
+
+  // Use localStorage as a reliable cross-window communication method.
+  // Create the storage entry to signal a protocol detection attempt for the
+  // next browser window to check.
+  window.localStorage.setItem(echoReplyKey, 'pending')
+  const cleanUpLocalStorage = () => {
```
jcnelson, line 85: Shouldn't this clear out *all* echo reply keys?  For example, what if the user closes their browser (or the browser crashes) while doing an authentication?

src/auth/authApp.js
kantai: I think this period of time is also weird if you don't click "Always open with this app" in your browser-- because it definitely takes my computer longer than 1 second to actually open the native app. Not sure about the trade off there.

src/auth/authApp.js
jcnelson: Thanks!

src/auth/authApp.js
zone117x: @kantai We detect when the "open blockstack app?" browser popup occurs, then pause the redirect timeout. When the popup is closed (user clicks either yes or no/cancel), the timer is restarted. 

If the user clicks "yes open blockstack app", then the page visibility API detects that the protocol handler is *probably* working, and restarts the timer with 3000ms timeout. This is for the purpose of reducing probability of false-negative redirect, while still performing the redirect if something goes wrong with the protocol handler.

src/auth/authApp.js
zone117x: To avoid cluttering the localStorage right? We can, but then run the risk of breaking an in-progress protocol-ping detection, for the same web app in another tab. Unlikely but possible.
Which error condition in either of these is worse?

We can also add a timestamp to the key's value (currently value is just set to "pending"), and clear out all keys older than 1 minute or so?

jcnelson: To avoid re-using echo reply keys by mistake.  Clearing them out if they're e.g. over an hour old should be fine.

zone117x: 👍will get this implemented.

src/auth/authApp.js
zone117x: @kantai For context on this see the comments on these two components:
* Browser popup detection https://github.com/blockstack/blockstack.js/blob/be9f384e31e87aa0af709bd8747e493e6a30d811/src/auth/authApp.js#L133
* Page visibility detection https://github.com/blockstack/blockstack.js/blob/be9f384e31e87aa0af709bd8747e493e6a30d811/src/auth/authApp.js#L163

kantai:
This worked for me pretty seamlessly in Chromium on Linux. I tested it by changing my protocol handler in Linux:

```bash
$ cat ~/.local/share/applications/blockstack-protocol-handler
#!/bin/bash
AUTH=$(echo "$1" | sed s#/##g | sed s/blockstack:// | sed 's/ //g')
xdg-open "https://deploy-preview-1774--reporter-beaver-73821.netlify.com/auth?authRequest=${AUTH}"
```

Which is the netlify preview of the browser built with echo response features. For other Linux testers, this could be the fastest way to test it out.

src/auth/authApp.js
kantai: Cool, that works for me, @zone117x -- and yeah, it correctly detected when that "open blockstack app?" popup occurs -- it might have been an older version I tested with that had that issue.

src/auth/authApp.js
```diff
@@ -84,23 +82,123 @@ export function redirectToSignInWithAuthRequest(authRequest: string = makeAuthRe
...
+  const iframe = document.createElement('iframe')
+  iframe.style.all = 'initial'
+  iframe.style.display = 'none'
+  iframe.src = locationSrc
+  // Flow complains without this check.
+  if (document.body) {
+    document.body.appendChild(iframe)
+  } else {
+    Logger.error('document.body is null when attempting iframe injection for protoocol URI launch')
+  }
```
kantai, line 225: Can this be refactored to look somewhat like `checkedBrowserLaunch(protocolURI, failCallback)` ?

It'd make it a bit more clear later on if we want to do something other than just redirect to browser.blockstack.org.

zone117x: Good idea, I will do that.

src/auth/authApp.js
zone117x: @jcnelson This should be resolved now.

src/auth/authApp.js
zone117x: Refactored the protocol detection code out of `redirectToSignInWithAuthRequest` function and into new `detectProtocolLaunch` function, with callbacks for handlers similar to how it was before.

src/auth/authApp.js
jcnelson: Yup, looks good :+1:

hstove (approved):

src/auth/authApp.js
zone117x: @kantai want to take a look at see if we can resolve this comment?

kantai: Yep, looks good, thanks for the refactor!

kantai:
I just want to make sure I understand the old-version/new-version behavior here:

1. If you're using this version of blockstack.js with an old version of the native browser, the app will (1) open an auth handler in the native browser and (2) also open browser.blockstack.org ?

2. If you're using normal versions of blockstack.js with a new version of the browser, the app will function exactly as it did before this protocol version update.

It seems like case (1) there is new enough behavior that we shouldn't call it backwards-compatible.

zone117x:
@kantai Correct on both those situations.

In situation 1, the native browser will be opened, and then a short time (a few seconds usually) later, the original tab will redirect to the hosted browser. Previous behavior would have left the original tab open on the unauthenticated web app and required being refreshed, or more likely closed by the user. 

I'm calling it backward-compatible because it is not a breaking change, it just changes the UX from one type of confusing behavior to another type. Although, arguably, the new behavior may be more confusing.

yknl (approved):

kantai:
I'm happy to approve this -- we should add this new behavior to the `CHANGELOG.md` file though:

```
If you're using this version of blockstack.js with an old version of the native browser, the app will (1) open an auth handler in the native browser and (2) also open browser.blockstack.org ?
```

So that we're aware whenever we merge this that some behavior will be changing.

kantai (approved):

PR #594, opened 2019-01-07
retry `putFile` with fresh Gaia config if the first write fails

hstove:
This PR adds some retry logic to `putFile`. In the case of Gaia tokens being revoked, app users will have invalid Gaia configuration cached in `localStorage`. All writes will fail until this Gaia config is updated manually.

This PR adds retry logic to `putFile`, so that if the first write fails, it'll make a new Gaia token, and then try writes again.

This should support any and all token revoke events, and make the experience for users much smoother.

TODO:

- [x] create a demo app that uses and demonstrates this. It'll start with bad Gaia config in `localStorage`, and then try an upload. It should work
- [x] add unit tests.

### DIFF
--- a/src/storage/hub.js
+++ b/src/storage/hub.js
@@ -31,7 +31,13 @@ export function uploadToGaiaHub(filename: string, contents: any,
                  },
                  body: contents
                })
-    .then(response => response.text())
+    .then((response) => {
+      if (response.ok) {
+        return response.text()
+      } else {
+        throw new Error('Error when uploading to Gaia hub')
+      }
+    })
     .then(responseText => JSON.parse(responseText))
     .then(responseJSON => responseJSON.publicURL)
 }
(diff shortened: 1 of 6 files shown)


hstove:
Here is a test app that uses this PR: https://gaia-retry-putfile.netlify.com/

To test that retries work:

1. Login with Blockstack
2. Click 'Corrupt local Gaia config'
3. Look in the console to see that your token was corrupted (there is a log message)
4. Click on 'putFile'. This will attempt a basic putFile. You should see one failed store call, then a call to `hub_info` to make new Gaia config, then a successful write after that with a fresh token.
5. Click on 'putFile (signed, unencrypted)'. This calls `putFile` with `{sign: true, encrypt: false}`. This is a branch of our code that makes two storage calls, instead of the usual one. You should see two failed network requests, then two successful ones with a fresh token.

src/storage/hub.js
```diff
@@ -31,7 +31,13 @@ export function uploadToGaiaHub(filename: string, contents: any,
                  },
                  body: contents
                })
-    .then(response => response.text())
```
hstove, line 34: @kantai @yknl what do you think about this error handling? If the response is 4xx, then `fetch` doesn't throw. I had to add this to throw errors for unauthorized writes.

kantai: This is fine with me. Though at some point we may want to make the errors more descriptive.

kantai (approved):

yknl (approved):
Thanks @hstove, this looks good to me 👍

PR #595, opened 2019-01-08
Added TypeScript Definitions

harrysolovay:
These definitions are by no means complete. For full type-safety, we'd need to pass some definitions from the API (profile, for example). Still, this'll do the trick by way of dismissing that pesky `Could not find a declaration file for module 'blockstack'` error, and for providing param/return hints.


zone117x:
This is super useful. Even for those not using typescript, most editors/IDEs will leverage typescript definition files to provide code-completion. 

Was this `types.d.ts` file written manually? If so, I'm not sure how we can currently run automated tests to ensure that the types.d.ts file is correct. This is important because once we have this typedef, many devs will use the editor/IDE provided intellisense as opposed to checking our docs, so an error in the definition file can cause a lot of annoyances for devs.

harrysolovay:
@zone117x yea it was written manually. The checks aren't particularly strict though. The main purpose is just to declare the blockstack module and suppress type errors... what kinds of tests might fail? Please let me know how I can improve this PR.

zone117x:
We would want tests that fail if the typedef (when manually written) does not match the implementation. 

Example scenario: say the argument names inside `getAccountBalance(address: string, tokenType: string)` are erroneously swapped in the typedef - easy to happen when manually maintaining and improving this typedef file. Most editors/IDEs ([stats on js dev environments](https://www.jetbrains.com/research/devecosystem-2018/javascript/)) will show this incorrect typedef info as code-completion / intellisense - even when coding in javascript (without typescript). And so most devs will see and use this function incorrectly until they catch the problem. 


This problem isn't a deal-breaker for this PR, just something to keep in mind. Are there any tools/libs for testing d.ts like this? What about a tool that could generate a d.ts from the jsdoc or Flow, so we don't have to manually maintain this?

harrysolovay:
I think that's very doable. The type signatures in the documentation only need to be subtly transformed. For the most part, it's just changing `String` to `string` and `Promise` to  `Promise<any>`. I could create a simple node tool for generating these... no formal parser even needed. Is there a place where I could find the schema for profile and other structures? If we're auto-generating, might as well make it fully type safe.

kantai:
There are _some_ messages and return types that have documented json schemas -- network functions that return values directly from the core API, e.g., have documented json schemas at https://core.blockstack.org/

Profiles do not have a json schema specified, but they probably should.

harrysolovay:
@kantai awesome –– where is the central repo for that documentation?

kantai:
That documentation is generated from this spec:

https://github.com/blockstack/blockstack-core/blob/master/docs/api-specs.md

However, I'm not sure of the best place for a schema for blockstack.js related things. Probably this repo, itself.

harrysolovay:
And @kantai –– is this document generated, or manually edited?

kantai:
That `md` file is manually edited.

harrysolovay:
@kantai I'm having some trouble assigning names to each interface generated from the given JSON schema. There are no "title" properties... and I don't think I'd write a particularly stable parser for guessing identifiers. Any tips? / would you be willing to add onto each schema with title? If so, I'll be able to spin this up pretty quickly :)

moxiegirl:
@harrysolovay  We are generating the Javascript Reference.  I'll add the instructions for how we are doing that. There are some outstanding issues with our method of generating. I'm looking at replacing or updating the generation mechanism.

harrysolovay:
@moxiegirl sounds good. Please keep me posted :)

harrysolovay:
@moxiegirl any updates?

moxiegirl:
@harrysolovay This is the missing how-to build docs information:

> The documentation is built using documentationjs. Which generates from comments in the `src/` files. Configure the left hand navigation with the `./documentation.yml` file.
> 
> To build the documentation:
> 
>  `npm run build-documentation`
>
> Running this command generates two files `docs.json`and `docs/index.html `. The `index.html` file contains the generated documentation. 
> 
> To learn more about configuration, see https://github.com/documentationjs/documentation/blob/master/docs/CONFIG.md

Regarding the reviews I'll ping the repo own here, @yknl he's currently in Hong Kong so his time zone may not match yours.

zone117x:
@harrysolovay Good news! We're aiming to move all of our js/flow projects over to Typescript. 
Work in in progress nearly complete for this repo: https://github.com/blockstack/blockstack.js/pull/616

In the mean time, here is a full [blockstack.d.ts](https://gist.github.com/zone117x/24397b47cd7fde5644a18a82fc377091) based off blockstack.js v19.

harrysolovay:
@zone117x that's great news!

moxiegirl:
@zone117x thanks for moving this forward.

PR #605, opened 2019-02-05
feature/update_storage_doc

geoffreyp:
By default, putFile and getFile use options for encrypted files. If we don't specify options we create and read encrypted files


yknl:
Thanks for the PR @geoffreyp!

PR #614, opened 2019-03-04
Improve dependency flow related to encryption

hstove:
This PR changes the folder structure related to encryption. I split the elliptic curve functions into one file, and the mnemonic functions into a different one. We still have `/encryption/index.js`, so this is backwards compatible.

The reason I did this is because `bip39` is, by far, our biggest dependency. However, almost all apps don't need our wallet-related code. They just need encryption/decryption functions. This way, an app can use code-splitting tools to only import the `ec` files, and not the `bip39` code. The result is a ~66% reduction in dependency size.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
   build:
     docker:
       # specify the version you desire here
-      - image: circleci/node:8.11.3
+      - image: circleci/node:10.15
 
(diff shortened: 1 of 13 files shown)


zone117x:
Wow nice!
Looks good. Tested against the browser e2e tests.

zone117x (approved):

yknl (approved):
Looks good to me 👍

PR #616, opened 2019-03-07
Typescript port

zone117x:
- [x] `src` compiling as typescript with `--noImplicitAny=true`
- [x] `tests/unitTests` compiling as typescript
- [x] unit tests passing
- [x] port eslint js rules to equivalent ts rules
- [x] documentation generation with `typedoc`
- [x] code coverage & codecov.io reporting
- [x] build `dist/blockstack.js` bundle with ES2015/ES6 support
- [x] build `lib/**/*.js` modules with Node v8.x support (ES2015/ES6 & CommonJS modules)
- [x] `tests/operationsTests` ported and passing
- clean up doc gen (this is being done off of another branch https://github.com/blockstack/blockstack.js/pull/628 - this main PR can be merged and the dependent PRs will be merged after). 

### Dist file output
Browserify was not really working in the new toolchain. It was replaced with Webpack. A side effect is that the `dist/blockstack.js` bundle is now 1.3MB (down from 3MB) due to its ability to perform node_module tree shaking. 

### Lib file output
The `lib` file generation no longer using babel due to Typescript's own ability to transform syntax to any ES lang version, and to output any module type. The tsconfig.json outputs `lib` files that are  es2016/es6 syntax and `CommonJS` modules, giving full support for Node v8.x. As like before these changes, web app bundlers need only the capability to perform module transforms and Node API (like `Buffer`) polyfills. Web apps using blockstack.js via browserify or webpack work the same as before. 

We are now committing `lib/**` files to the repo. This is common practice for Typescript projects. This also allows projects to install the blockstack.js lib from the repo rather than npm. Allowing us and other developers to use specific branches/features without going through the npm release process. For example:
  * `npm install blockstack@blockstack/blockstack.js#feature/beta-mode`
  * or `npm install blockstack@blockstack/blockstack.js#develop`

### Reviewer notes
To reduce the git diff to make reviewing easier, the original doc files have remained in place, and the dist & lib files have not been committed. These can be re-built before and committed after review, but before merging.


src/storage/hub.ts
```diff
@@ -61,8 +62,16 @@ function makeLegacyAuthToken(challengeText: string, signerKeyHex: string): strin
...
                                      + (signerKeyHex.length === 64 ? '01' : ''))
-    const digest = bitcoin.crypto.sha256(challengeText)
-    const signature = signer.sign(digest).toDER().toString('hex')
+    const digest = bitcoin.crypto.sha256(Buffer.from(challengeText))
+
+    const signatureBuffer = signer.sign(digest)
+    const signatureWithHash = bitcoin.script.signature.encode(
+      signatureBuffer, bitcoin.Transaction.SIGHASH_NONE)
+    
+    // We only want the DER encoding so remove the sighash version byte at the end.
```
zone117x, line 71: Note to reviewers: this has been broken in master for quite some time. Since the upgrade from bitcoinjs-lib 3.x to 4.x. 
No unit tests for this and Flow had no understanding of the bitcoinjs-lib typings. Typescript identified this bug and I pulled in the already fixed and tested code from Gaia.

src/operations/skeletons.ts
```diff
@@ -138,7 +138,7 @@ function asAmountV2(amount: AmountType): AmountTypeV2 {
   // convert an AmountType v1 or v2 to an AmountTypeV2.
   // the "units" of a v1 amount type are always 'BTC'
   if (typeof amount === 'number') {
-    return { units: 'BTC', amount: BigInteger.fromByteArrayUnsigned(String(amount)) }
```
zone117x, line 141: Note to reviewers:
`bigi`, a BigInteger lib, is a relatively large, no longer maintained dependency that appears to have been originally used in our code due to it being a child dependency, of a child dependency, of an old version of `bitcoinjs-lib`. 

Since then, all of our child dependencies have settled onto the `bn.js` lib. 

This particular function usage `fromByteArrayUnsigned(someNumber.toString())` is misleading. View the src https://github.com/cryptocoinjs/bigi/blob/cb702633d6587a4326cc8c79b4024444caf9531f/lib/convert.js#L12 to see that this usage always falls through to `fromString()`. 

The refactor to using `bn.js` takes this into consideration.

kantai: Excellent -- yes, that `.fromByteArrayUnsigned` was always an incredibly misleading function.

hstove:
Going through the code with VSCode, `tsc` is showing some flags for using `Promise` with an `es5` target, because `Promise` is not in `es5`. That's a problem, right? Or is a polyfill used? There are similar errors - `Object.assign` and `String.prototype.startsWith`.

zone117x:
> Going through the code with VSCode, `tsc` is showing some flags for using `Promise` with an `es5` target, because `Promise` is not in `es5`. That's a problem, right? Or is a polyfill used? There are similar errors - `Object.assign` and `String.prototype.startsWith`.

Nice catch. I changed this to es2015/es6 (I mixed them up or typo'd after reading through https://node.green). The intention for the `lib` output is to continue support for native Node v8.x out of the box, and to continue support web app packing tools that do not perform non-module ES syntax transforming (like browserify). 

I updated the PR comment with more info as well.

hstove (approved):
Although this is quite a large PR, I am not able to find anything that I've found wrong or needs changing. I certainly could have missed something. I read through the code, diffs, and tests while trying to keep the whole concept together. I've also tested this code, both in Radiks (which I've just converted to TS, and in Banter, while playing around with converting some of it to TS. I did not see anything wrong going on.

It was a great challenge to code review this, because it's such a big change. Luckily, we're not changing any functionality, and it seems like you've caught some bugs, which is epic. It's cool to be able to use these types in other TS projects - that really helps.

Good stuff!

kantai (approved):
This looks great to me @zone117x --

I tested out the changes to the operations files using the integration test (which tests the commands against a regtest spin up of blockstack-core), and they all worked fine, though those integration tests are a little finicky.

zone117x:
Thanks for all the review work @hstove much appreciated. 

Mary has docs at a point where its at least as good as the previous doc gen (with many improvements) in this PR, which is almost ready https://github.com/blockstack/blockstack.js/pull/633

I'm going to go ahead and merge this so unblock the other PRs dependent on this one.

PR #622, opened 2019-03-12
Fixes for makeAuthRequest(...) and redirectToSignIn(...)

zone117x:
Backwards compatibility fixes:
* Fixes https://github.com/blockstack/blockstack.js/issues/618
* Fixes `makeAuthRequest(...)` missing in v19

### DIFF
--- a/src/auth/authApp.js
+++ b/src/auth/authApp.js
@@ -2,7 +2,9 @@
 import queryString from 'query-string'
 import { decodeToken } from 'jsontokens'
 import { verifyAuthResponse } from './index'
-import { BLOCKSTACK_HANDLER, isLaterVersion, hexStringToECPair } from '../utils'
+import { 
+  BLOCKSTACK_HANDLER, isLaterVersion, hexStringToECPair, nextMonth 
+} from '../utils'
 import { getAddressFromDID } from '../index'
 import { InvalidStateError, LoginFailedError } from '../errors'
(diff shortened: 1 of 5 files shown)


yknl (approved):
Thanks for the fix! Looks good to go

src/auth/authApp.js
```diff
@@ -166,12 +134,12 @@ export function signUserOut(redirectURL: ?string = null) { // eslint-disable-lin
  * by special authenticators.
  * @return {String} the authentication request
  */
-export function makeAuthRequest(transitPrivateKey: string,
-                                redirectURI: string,
-                                manifestURI: string,
-                                scopes: Array<string>,
+export function makeAuthRequest(transitPrivateKey: ?string,
+                                redirectURI: string = `${window.location.origin}/`,
```
kantai, line 138: These defaults make this function un-callable from non-browser contexts. It makes sense that something like `redirectToSignIn` wouldn't work outside of a browser context, but `makeAuthRequest` seems like it should.

kantai: Actually, nevermind -- it seems like `makeAuthRequest` is deprecated anyways, and non-browser apps would use the UserSession API anyways.

kantai (approved):
This looks good to me @zone117x

hstove (approved):
Dang.. good catch and thanks for fixing this.

PR #628, opened 2019-03-16
Cleanup `Impl` functions & bug fixes

zone117x:
The v19 `_Impl` methods have caused a lot of bugs, in addition to being nasty code smells. 

I cleaned these up during the process of implementing `deleteFile` and revocation. 

For context..
* For most of our public functions, we had 3 versions: the v19 backwards compat one, the `Impl` suffixed one containing the original logic, and the UserSession method. 
* Most of these had same function signatures and large blocks of copy & pasted docs, with various bugs/mismatches between their behavior, documentation updated in one place but not the others, and **function parameters being ignored** during this bouncing around. 
* Lots of entry points / code paths not covered in tests. 
* Many had confusing call graphs where a given function would be called from *module A*, which would call another definition of the function in *module B*, then *module B* would call back into *module A*.  `redirectToSignIn` for example: 

```ts
// userSessions.ts
redirectToSignIn() {
  return redirectToSignInImpl(this)
}
```
```ts
// authApp.ts
export function redirectToSignIn(redirectURI: string = ..., 
                                 manifestURI: string = ..., 
                                 scopes: string[] = ...) { 
  userSession.redirectToSignIn()
}

export function redirectToSignInImpl(caller: UserSession) {
  const transitKey = caller.generateAndStoreTransitKey()
  const authRequest = caller.makeAuthRequest(transitKey)
  redirectToSignInWithAuthRequestImpl(caller, authRequest)
}
```

Some of the bugs fixed during the refactor:
* Fix `gaiaHubConfig` never getting saved to localStorage after `connectToGaiaHub` call - causing the `/hub_info` fetch and resolving to occur all the time. 
* Added some of the optional parameters from original functions to the equivalent methods on the `UserSession` class so apps updating have less potential changes to make. 
* Perform runtime checks for `window` APIs used in various functions - improves error messaging for environments that invoke these functions incorrectly or without providing the requisite polyfills (partially addresses https://github.com/blockstack/blockstack.js/issues/611). 
* Add explicit default module export to `index` so that it doesn't have to be generated with tooling (like babel) on our end or consumer end. 

### Testing
* Passes blockstack.js unit tests.
* Passes blockstack-browser e2e tests. 
* Works in the hello-blockstack browserify project. 
* Works in a simple [hello-blockstack webpack based project](https://github.com/zone117x/hello-blockstack-webpack), with testing done on just about all the various module importing methods: 
```js
import blockstack from 'blockstack'
import * as blockstack from 'blockstack/dist/blockstack'
import blockstack from 'blockstack/dist/blockstack'
import { isUserSignedIn, isSignInPending, etc... } from 'blockstack'
import { isUserSignedIn, isSignInPending, etc... } from 'blockstack/lib/auth/authApp'
import { isUserSignedIn, isSignInPending, etc... } from 'blockstack/dist/blockstack'
const blockstack = require('blockstack')
```


yknl (approved):
These changes look good to me. Thanks @zone117x

hstove (approved):
A lot of really useful changes in here. Thanks!

PR #630, opened 2019-03-18
Fix `listFiles` not performing gaia reconnection

zone117x:
`listFiles` was not performing the gaia connection retry logic that all of the rest of the storage methods perform. This is the mechanism responsible for a few important things like automatically re-authorizing expired auth tokens, and fetching new gaia hub server if the user changed it and took down their old server, etc.


yknl (approved):
Looks good, verified that it's reconnecting to Gaia hub every time on listFile. (Gaia hub config isn't saved properly in the base branch)

jcnelson (approved):
Looks good to me!

PR #631, opened 2019-03-18
Ensure protocol handler detection happens with modular usage of blockstack.js

zone117x:
Perform the protocol handler echo detection and page redirect when either the main index file is loaded _or_ when `isSignInPending()` is invoked. 

This fixes protocol echo detection for apps that use an auth flow in combination with module packing in ways that prevent the main blockstack.js index file from being executed on their auth handler page. 

Consider an auth flow that is composed of nice, separated endpoints, and lean modular usage of blockstack.js. This should be common among web app devs that try to address the warnings that tools like webpack, Google Insights, etc.. show about pages sizes. 

An example of such a setup might be something like:

`myapp.com/index`:
```js
import { 
  redirectToSignIn
} from 'blockstack/lib/auth/authApp'

signInButton.onclick = () => {
   // Redirect back to a page only designed to handle pending blockstack sign ins. 
   const redirectURI = 'myapp.com/handle_auth'
   blockstack.redirectToSignIn(redirectURI)
}
```

`myapp.com/handle_auth`:
```js
import { 
  isSignInPending,
  handlePendingSignIn 
} from 'blockstack/lib/auth/authApp'

if (isSignInPending()) {
  handlePendingSignIn().then(...)
}
```

The local browser does a protocol detection redirect back to `myapp.com/blockstack_auth?echoReply=...`, and if the web app is bundled/published in a way that does not load the main blockstack.js index file on for page, then this PR ensures that the echo detection code is still ran so that the redirect auth flow works correctly. 

There are a couple minor changes to make sure this doesn't break apps in certain cases. For example, consider an app using following auth flow (extending the above example):

`myapp.com/handle_auth`:
```js
if (isSignInPending()) {
 handlePendingSignIn().then(() => {
    window.location = '/index'
  }).catch((error) => {
    window.alert(`Unexpected error in handlePendingSignIn! ${error}`)
  })
} else {
  // This is page is only meant to handle the auth redirect?! 
  // If `isSignInPending()` is false, the user shouldn't be here.
  // Send them back to the main page where they can click sign in.
  window.location = '/index'
}
```

* This PR makes `isSignInPending()` return true if the protocol echo is detection, in order to prevent something like the `else { window.location = '/index.html' }` condition from the above example. 

* It also makes `handlePendingSignIn()` check for the  protocol echo detection, in order to prevent something like the `window.alert('Unexpected error in handlePendingSignIn!')` condition from the example. If found it returns a Promise that waits a few seconds, during which time the page will have been redirected. If for some reason the redirect in the protocol echo detection code somehow failed, the Promise resolves to an error.


yknl (approved):
Looks good to me 👍

yknl:
new changes look good 👍

src/auth/detectProtocolLaunch.ts
```diff
@@ -0,0 +1,180 @@
...
+ * launched via a custom protocol URI. 
+ * @param {String} authRequest
+ * The encoded authRequest to be used as a query param in the custom URI. 
+ * @param {String} successCallback
+ * The callback that is invoked when the protocol handler was detected. 
+ * @param {String} failCallback
+ * The callback that is invoked when the protocol handler was not detected. 
+ * @return {void}
+ */
+export function detectProtocolLaunch(
```
zone117x: Note to reviewers: this `detectProtocolLaunch` function had no code/logic changes. It was simply moved to its own file. Due to its monolithic nature, and it being all DOM code, which doesn't make sense in context of the other functions in the file it was previously in.

src/auth/protocolEchoDetection.ts
```diff
@@ -0,0 +1,58 @@
...
+ * which indicates that this page was only opened to signal back to the originating 
+ * tab that the protocol handler is installed. 
+ * If found, then localStorage events are used to notify the other tab,
+ * and this page is redirected back to the Blockstack authenticator URL. 
+ * This function caches its result and will not trigger multiple redirects when
+ * invoked multiple times. 
+ * @returns True if detected and the page will be automatically redirected. 
+ * @hidden
+ */
+export function protocolEchoReplyDetection(): boolean {
```
zone117x, line 23: Note to reviewers: this function had no code/logic changes. It was simply moved to its own file. Due to its monolithic nature, and it being all DOM code, which doesn't make sense in context of the other functions in the file it was previously in.

hstove (approved):
This works as expected. I used our react app generator, which uses modular imports out of the box.

PR #633, opened 2019-03-18
New doc gen started from Matt's PR

moxiegirl:
Follows on https://github.com/blockstack/blockstack.js/pull/628 I'll rebase after it is merged.

* `Usersession.getOrSetLocalGaiaHubConnection` not previously exposed should it be?
* Did we intend `to deprecate` the `getFileURl` cin `storage/index.ts` ?
* As far as the include files go, right now I have most of them attached. I'm mulling whether they shouldn't simply be in the using docs and not in here at all.
* `storage/index.ts` has two functions `encryptContent` and `decryptContent` they are not putting out deprecation warnings, should they?
* Existing Javascript is not exposing the `Profile1 object but we have examples using the `toToken` method in there.  I think we want this public


moxiegirl:
@zone117x @kantai Can I get y'all to take a look at this. I think the first pass is done. I'd like to merge this and then do some fine tune work after the fact.

PR #634, opened 2019-03-19
Only require node 8 and above

hstove:
@zone117x is there a reason we are requiring node >=10.13? Otherwise, I think we should keep it, as you can't install `v19.0.1-beta.1` without node >=10.13. If it's for local development, there are other ways we can handle that. If it's required, no problem, we can close this.


zone117x (approved):
Yep this should be okay

yknl (approved):

PR #635, opened 2019-03-20
Fix Linkedin verification proof

otech47:
**Description**

Fixes blockstack/blockstack.js#550

The DOM parsing selectors for the Linkedin
verification have been updated so that the
cheerio library could locate the correct data

The HTML generated from cheerio does not match
the HTML rendered in the browser which makes
it more difficult to parse for the correct
selectors

**Testing**

1. Prepare a local development instance of blockstack-browser
2. Ensure that you are using this [forked branch](https://github.com/otech47/blockstack.js/tree/test-linkedin-verification) (edit package.json to include `"blockstack": "github:otech47/blockstack.js#test-linkedin-verification"`) of blockstack.js before running `npm install`
3. Authenticate with any Blockstack ID and follow the steps to verify your LinkedIn profile
4. Verify the publicly posted link and confirm the blue checkmark appears

I ran this test manually using an account where I had attempted a LinkedIn verification before and was able to get to step 4 successfully. I don't anticipate any issues with new accounts attempting the same since this is simply a matter of changing selectors.

I also revised the unit tests because it seems a [test case was added ](https://github.com/blockstack/blockstack.js/commit/7704d2710ca6b6dedf82694f67568e15fa17578b#diff-5fffa10b590af6297fc38e7d04821de3R71) to handle the Linkedin verification failure. I removed that test and added extra test data to mock the HTML.

This is my 1st PR so I figured I might as well open this now to see if there is any testing guidance or comments from the maintainers or previous contributors @yknl  @shreyasthiagaraj on anything additional needed here.  For example I wasn't sure whether to include the `dist/blockstack.js` file or not.


zone117x:
@otech47 Can you merge this onto develop? I suspect if you remove the `dist/blockstack.js` file diff from the commit then it may not have merge conflicts, because other than that it should just be file renames.

otech47:
@zone117x Just merged the upstream `develop` branch. Any squashing needed here?

yknl:
This doesn't seem to work for me. Is LinkedIn displaying different DOM for different people? Were you able to test this @zone117x ?

otech47:
> This doesn't seem to work for me. Is LinkedIn displaying different DOM for different people? Were you able to test this @zone117x ?

@yknl I do have an active trial of LinkedIn Premium I tested with so I wouldn't rule it out. Will investigate some more

otech47:
Also I realized the testing instructions aren't quite right since this branch doesn't have the `/dist` build anymore so I edited the testing instructions to point to a different branch of my fork

otech47:
@yknl I just tested with a brand new LinkedIn account that doesn't have Premium and I was able to get the blue checkmark so it could be a difference in testing conditions. Which of the 4 testing steps did you get hung up on? Could you post the link to the verification URL you tried?

zone117x:
@otech47 I haven't had a chance to test yet, but here is my verification URL (regular linkedin account) if you'd like to test that as well: https://www.linkedin.com/feed/update/urn:li:activity:6492115381757579264

zone117x:
Just had a chance to test this and it works! Great job, especially with the added tests. Much appreciated.

(Note to @yknl for testing - I pulled and compiled the fork then manually swapped it into the blockstack-browser node_modules, and ran `npm run dev`)

I wouldn't mind merging in current state, but I do have some suggestions on future-proofing this a bit.

The queries are against class names that are almost definitely being generated:

```js
const profileLink = $('.share-update-card__header > a')
```
```js
const postContent = $('.share-update-card__update-text')
```

Looking over their DOM, they implement:
* Standardized Open Graph meta tags (e.g. `<meta property="og:type" content="article">`)
* A mostly-standard element page structure (e.g. `body > main > header`)

Both of these are less likely to change than the generated classes, and luckily for us the two strings we need can be queried using both of these.

The username is inside a URL queryable through a reasonable DOM tree. I'd recommend using:
```js
const profileLink = $('body > main header a')
```

And even better, the post content is in an open graph tag. I'd recommend using:
```js
const postContent = $('head > meta[property="og:title"]')
```

yknl:
Thanks for verifying @zone117x, I was able to get it to work, might have been issue on my end. I agree with using the open graph tag, this is what we do for facebook, twitter and instagram verifications. I'm not seeing the tags in the source of your LinkedIn verification post though.

zone117x:
> I'm not seeing the tags in the source of your LinkedIn verification post though.

You are probably signed in to LinkedIn and viewing the html. They provide a different DOM for logged in users visiting the page directly. Makes sense they wouldn't include the open graph tags in that case, since they are meant for apps/servers like Slack/Twitter/Facebook to fetch and use.

yknl:
@zone117x you're right, it shows a different DOM when signed out. That would work 👍

otech47:
@zone117x @yknl Thanks for the review! I'll go ahead and try those other selectors out today

zone117x (approved):
Works and code looks good. Thank you!

yknl (approved):
Looks great! Thanks for this @otech47

PR #640, opened 2019-03-29
Fix page scrolling from injected elements during protocol detection

zone117x:
Fixes https://github.com/blockstack/blockstack.js/issues/639

**Problem**: Page scrolls and/or a scrollbar would be shown when `redirectToSignIn` was called.

**Cause**: Protocol handler detection injects a hidden `<input>` element and calls `focus()` on it. The element is hidden but was not setting the `position` css style.

**Fix**: Now sets the style to `position: fixed; top: 0; height: 0; width: 0;`, so that the element has zero effect on the viewport. No scrolling occurs at all now. And the superfluous scrollbar no longer appears on pages using something like `<body><div style="height: 100%"> ...`.

### Testing

I just used a hello-blockstack app, set the main div height to 150%, and clicked sign in. Repeated this after scrolling to top of page, bottom of page, and middle of page. Verified that scrollbar no longer appears with main div height set to 100%.

### DIFF
--- a/src/auth/protocolLaunch.ts
+++ b/src/auth/protocolLaunch.ts
@@ -82,15 +82,12 @@ export function launchCustomProtocol(
   
   const inputPromptTracker = window.document.createElement('input')
   inputPromptTracker.type = 'text'
-  const inputStyle: CssStyle = inputPromptTracker.style as any
-  // Prevent this element from inherited any css.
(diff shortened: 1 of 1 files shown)


yknl (approved):
Verified it fixes the scrolling problem 👍

hstove (approved):

PR #644, opened 2019-04-03
API Reference Documentation gen redux

moxiegirl:
Trying this merge again


src/auth/appConfig.ts
```diff
@@ -10,28 +10,34 @@ import { DEFAULT_CORE_NODE, DEFAULT_SCOPE, DEFAULT_BLOCKSTACK_HOST } from './aut
  * `window.location.origin` as the app domain.
  * On non-browser platforms, you need to
  * specify an app domain as the first argument.
+ * 
+ * 
  * @type {AppConfig}
```
zone117x: All of these `@type {...}` are jsdoc/flow leftovers, we can remove them so they don't show up as generic/unknown tags in the typedoc gen

src/auth/appConfig.ts
```diff
@@ -10,28 +10,34 @@ import { DEFAULT_CORE_NODE, DEFAULT_SCOPE, DEFAULT_BLOCKSTACK_HOST } from './aut
...
+ * 
+ *   [[include:auth.md]]
  */
 export class AppConfig {
   /**
    * Blockstack apps are uniquely identified by their app domain.
    * @type {string}
    */
-  appDomain: string
+  public appDomain: string
```
zone117x: I originally suggested we use these `public`/`private` modifiers for a couple reasons:
1) They are a recommended eslint/tslint rule (currently not enabled), which we may want in the future.
2) I originally thought it was one of the only ways to control some parts of the docgen output (before finding out about the `@ignore` feature).

But now, I think we should hold off on introducing them for a couple reasons. Typescript enforces `private` so it controls how other typescript projects can use this lib, and `public` gives devs strong indication of intended usage/purpose. 

I think we'll be in a better position to do something like enable the eslint flag and go through the codebase to make these decisions once we have some of our other typescript repos (and 3rd parties) using this lib.

src/auth/authApp.ts
```diff
@@ -55,26 +57,10 @@ export function isUserSignedIn() {
...
- * @param  {Array} [scopes=DEFAULT_SCOPE] Defaults to requesting write access to
- * this app's data store.
- * An array of strings indicating which permissions this app is requesting.
- * @return {void}
+ * @deprecated v19
+ * 
+ * Use [[UserSession.redirectToSignIn]] instead.
+ * 
  */
 export function redirectToSignIn(redirectURI?: string,
```
zone117x: These `@deprecated` tags and reference links are really handy! I think we should use these, but also keep the desc and param descriptions intact until we have the rest of tutorials/docs, app generators, and e2e tests changed to using the `UserSession` class. 

And maybe keep them until/unless we remove the functions all-together, since the available params and behavior between these functions and the `UserSession` versions may not be the same, or diverge, either intentionally or due to bugs.

src/profiles/profileSchemas/personLegacy.ts
```diff
@@ -1,3 +1,10 @@
+/**
+ * 
+ * @param serviceName
```
zone117x, line 3: There are a few areas with empty `@param`s are these accidental?

zone117x (commented):
Pulled down locally, cleaned and rebuilt, and seeing some items that don't seem like we want to expose? 

image

src/wallet.ts
```diff
@@ -36,49 +43,60 @@ function hashCode(string: string) {
...
- *  paths for a standard blockstack client wallet. This includes paths
- *  for bitcoin payment address, blockstack identity addresses, blockstack
+ *  The `BlockstackWallet` class manages the hierarchical derivation
+ *  paths for a standard Blockstack client wallet. This includes paths
+ *  for Bitcoin payment address, Blockstack identity addresses, Blockstack
  *  application specific addresses.
+ * 
  *  @private
+ * 
+ *  @ignore
```
zone117x: Looks like the whitespacing here is causing the tags to be interpreted as a comments. Should be fixable by just removing the extra column of spaces.

Example: 
image

src/auth/appConfig.ts
moxiegirl: OK. I'll make those changes.

src/auth/authApp.ts
moxiegirl: @zone117x I think that we should just remove them and leave the deprecated. The generated files have links to the code if they are using the existing call and want to maintain it -- a single click.  And the link to the `UserSession` is there if they got here incorrectly or want to go to the latest.

src/profiles/profileSchemas/personLegacy.ts
moxiegirl: These are on ignored calls that aren't in the docs.  So, I didn't spend time on filling them in.

src/auth/userSession.ts
```diff
@@ -31,13 +33,16 @@ import { BLOCKSTACK_DEFAULT_GAIA_HUB_URL, DEFAULT_BLOCKSTACK_HOST, DEFAULT_SCOPE
  *
  * A user can be signed in either directly through the interactive
  * sign in process or by directly providing the app private key.
+ * 
+ * @type {UserSession}
+ * 
  */
 export class UserSession {
```
zone117x, line 39: I noticed that links to type from external files are sometimes not generated. For example `GetFileOptions` here:

----

image

----

I found that using typescript's [`import() types syntax`](https://davidea.st/articles/typescript-2-9-import-types) fixes the issue. Example with `getFile(path: string, options?: import('../storage').GetFileOptions)`:

----

image

src/wallet.ts
moxiegirl: Fixed.

src/auth/userSession.ts
moxiegirl: I'd like to fix this after the merge. I have a number of other things to do. This is just the first pass.

moxiegirl:
Yeah, this wasn't happening before I rebased. Like I said, I need to do another pass for other things I want to do before we go live. I'll address this after the merge.

> 
> image

src/auth/authApp.ts
zone117x: Two main issues I'm concerned about here:

1. `UserSession` versions of these functions are not covered in our integration tests, and could have unknown or undocumented behavioral differences from the original. This has already happened with several of these functions. Leaving docs blank and linking to the other version implies that this is will not be the case. 

2. The `UserSession` versions can (and already have) intentionally diverged in usage & behavior from the original ones. For example, the current docs for `UserSession.redirectToSignIn(...)`, copied over from the original, are not really accurate. It tells devs to use other class methods (like `makeAuthRequest`) if they `require more fine grained control over how the authentication request`. Once updated to explain that class constructor options are the intended way to customize auth flow options, these empty docs with links to UserSession are _less_ than helpful, and will force devs to read through src to figure out what's going on. 

Removing these functions from the docs entirely would cause devs less problems and confusion, but we can't yet do that because all our other material still points devs to these functions. 

Seems reasonable to leave the existing, accurate docs for these intact, then we can eventually start removing things once they are effectively deprecated, rather than just tagged as such.

moxiegirl: 1. Isn't this a problem with how `UserSession` and the tests were implemented? If use of `UserSession` is problematic **we should fix the problems for `UserSessions`** . We deprecated the old functions, we say they are deprecated, we should not be documenting them. But if we haven't safely implemented the new functions why did we deprecate the old?

2. The fact that the UserSessions diverged seem fine to me --- they are new and the old ones are deprecated. Again, the problems with the `UserSession` docs should be *fixed if they are the recommended path forward* not left in place with problems we know exist.

yknl:
To address the questions from the previous PR:

> Usersession.getOrSetLocalGaiaHubConnection not previously exposed should it be?

`getOrSetLocalGaiaHubConnection` should be private, I don't see a reason for developers to need to use this directly.

> Did we intend to deprecate the getFileURl cin storage/index.ts ?

The use of `getFileURl` directly is being deprecated but it should be used via UserSession in v19+

> As far as the include files go, right now I have most of them attached. I'm mulling whether they shouldn't simply be in the using docs and not in here at all.

Are you proposing that we remove the `includes/auth.md` and `includes/storage.md`?

> storage/index.ts has two functions encryptContent and decryptContent they are not putting out deprecation warnings, should they?

They were outputting deprecation warnings in v19. It doesn't look like the way it's exported has changed. So I would say it should be.

> Existing Javascript is not exposing the Profile1 object but we have examples using the toToken` method in there. I think we want this public

Is this referring to file:///Users/Yukan/Desktop/work/blockstack/blockstack.js/docs/classes/profile.html?

yknl:
Do we want to show the Interfaces, Type alias, Variables and Object literals sections? They seem a bit irrelevant for most developers and might be confusing. I would also recommend leaving the social media service classes out of the index.

src/auth/authApp.ts
moxiegirl: OK. I added back in the content for the deprecated methods.

src/auth/userSession.ts
moxiegirl: This worked for the interface `GetFileOptions` but does not work for the type `PutFileOptions`.

moxiegirl:
> getOrSetLocalGaiaHubConnection

Hidden.

@yknl thanks for your replies. Implemented as you suggested.

> Do we want to show the Interfaces, Type alias, Variables and Object literals sections? They seem a bit irrelevant for most developers and might be confusing. I would also recommend leaving the social media service classes out of the index

We are showing these in the other SDKS so we are including them as agreed on in Slack.

src/auth/appConfig.ts
```diff
@@ -64,17 +64,18 @@ export class AppConfig {
...
    * @param {string} manifestPath - path relative to app domain of app's manifest file
    * @param {string} coreNode - override the default or user selected core node
-   * @param {string} authenticatorURL - the web-based fall back authenticator
+   * @param {string} authenticatorURL - the web-based fall back authenticator 
+   * ([[DEFAULT_BLOCKSTACK_HOST]])
    */
   constructor(scopes: Array<string> = DEFAULT_SCOPE.slice(),
               appDomain: string = window.location.origin,
-              redirectPath: string = '',
+              edirectPath: string = '',
```
zone117x: Typo here an a couple lines below

src/auth/userSession.ts
```diff
@@ -1,4 +1,4 @@
-
+import { MemoVoidIterator } from 'lodash'
```
zone117x: Looks like an accidental auto-import typo here

src/auth/userSession.ts
```diff
@@ -100,15 +114,17 @@ export class UserSession {
...
-   * Please note that the protocol handler detection
-   * does not work on all browsers.
-   * @param  {String} authRequest - the authentication request generated by `makeAuthRequest`
-   * @return {void}
+   * Please note that the protocol handler detection does not work on all browsers.
+   * 
+   * @param authRequest A request string built by the [[makeAuthRequest]] function
+   * @param blockstackIDHost The ID of the Blockstack Browser application.
+   * 
+   * @returns {MemoVoidIterator}
```
zone117x: Same here (looks like an accidental auto-import typo)

src/wallet.ts
```diff
@@ -36,49 +43,58 @@ function hashCode(string: string) {
...
+ * The `BlockstackWallet` class manages the hierarchical derivation
+ * paths for a standard Blockstack client wallet. This includes paths
+ * for Bitcoin payment address, Blockstack identity addresses, Blockstack
+ * application specific addresses.
+ * 
+ * @ignore
  */
 export class BlockstackWallet {
-  rootNode: BIP32
+  private rootNode: BIP32
```
zone117x, line 71: Some `private` modifiers here an in a couple other files

package.json
```diff
@@ -142,6 +142,7 @@
     "elliptic": "^6.4.1",
     "form-data": "^2.3.3",
     "jsontokens": "^1.0.0",
+    "lodash": "^4.17.11",
```
zone117x: Looks like an accidental auto-import typo here

src/auth/userSession.ts
moxiegirl: This was a fix recommended by the linter to fix a problem.

src/auth/userSession.ts
moxiegirl: Again...eslint.  Do you want me to turn off the linter?

package.json
moxiegirl: See above comments on the linter.

src/auth/authApp.ts
```diff
@@ -26,23 +26,53 @@ const DEFAULT_PROFILE = {
...
   hubUrl: string;
+  // maybe private: this would be an advanced field for app devs to use. 
   authResponseToken: string;
+  // private: does not get sent to webapp at all.
   coreSessionToken?: string;
+  // private: does not get sent to webapp at all.
   gaiaAssociationToken?: string;
+  // private: does not get sent to webapp at all - also this is a 
+  // dup/bug of the previous field I think.
+  associationToken?: string;
```
zone117x: Hank recently fixed this bug and removed this field.

yknl (approved):

zone117x (approved):

PR #645, opened 2019-04-04
Fixes usage of association token with UserSession

hstove:
When testing [Jude's association token PR](https://github.com/blockstack/blockstack-browser/pull/1833/files), I found a bug when using association tokens with v19. The issue is that, when calling `handlePendingSignIn`, we pass the `associationToken` property of the `authResponse` to the `gaiaAssociationToken` property of `userData`. However, when calling `setLocalGaiaHubConnection`, we were using the `associationToken` property of `userData`. This is always null, so the association token is not used.

I would also suggest that we remove the `associationToken` property from `userData`.

### DIFF
--- a/src/auth/authApp.ts
+++ b/src/auth/authApp.ts
@@ -36,7 +36,6 @@ export interface UserData {
   authResponseToken: string;
   coreSessionToken?: string;
   gaiaAssociationToken?: string;
-  associationToken?: string;
   profile: any;
   gaiaHubConfig?: GaiaHubConfig;
 }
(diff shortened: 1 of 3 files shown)


zone117x:
Nice find @hstove. I've been confused about those two fields. 

> I would also suggest that we remove the associationToken property from userData.

Could you do this in this PR?

zone117x (changes requested):
Nice find @hstove. I've been confused about those two fields. 

> I would also suggest that we remove the associationToken property from userData.

Could you do this in this PR?

yknl (approved):
Thanks for the fix!

jcnelson:
Looks good to me.  Thanks for taking the time @hstove!

jcnelson (approved):

hstove:
@zone117x good call, done :+1:

zone117x (approved):

PR #646, opened 2019-04-08
Prepare script to build lib on install via git

zone117x:
### Goal of this PR
Ability to install blockstack.js dependency from git, without requiring branches/PRs to commit a lot of large dist files. 

`lib` dir is nearly 200 files (js, js.map, d.ts files), and `dist` dir contains a 1.3MB js and a 4MB js.map file. 

When omitted it makes PRs or feature branches needlessly difficult to use/test. 
When committed they clutter the PR and make reviewing diffs needlessly difficult. 

### Fix

See https://docs.npmjs.com/cli/install
> `npm install <git remote url>`:
> If the package being installed contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.

See https://blog.jim-nielsen.com/2018/installing-and-building-an-npm-package-from-github/ for similar example. 

### Implementation

Add a `package.json` `prepare` script so that lib/dist files are built and available when blockstack.js is installed from a git repository. 

Add `lib` and `dist` to directories to `.gitignore` so they are no longer committed. 

Note: the blockstack.js dist bundle is still readily available via [unpkg](https://unpkg.com/): https://unpkg.com/blockstack/dist/blockstack.js

### Testing

Install this branch of blockstack.js into a project via its git source (shorthand):
```
npm i blockstack/blockstack.js#feature/build-as-git-dep
```

Ensure your `package.json` contains
```json
  "dependencies": {
    "blockstack": "github:blockstack/blockstack.js#feature/build-as-git-dep"
  }
```

Ensure `node_modules/blockstack/lib/**/*.js` and `node_modules/blockstack/dist/*.js` files exist in your project dir.


moxiegirl:
@zone117x is this just for PRs? Or, do you anticipate I will need to update all the tutorials or provide docs for a particular use case.

zone117x:
Do we have documentation telling devs to use `https://github.com/blockstack/blockstack.js/blob/master/dist/blockstack.js` ?

If so, those should be updated to point to unpkg:
> Note: the blockstack.js dist bundle is still readily available via unpkg: https://unpkg.com/blockstack/dist/blockstack.js

Otherwise, this doesn't require doc changes for developers consuming the lib. I don't see a need to state that the lib can be installed via git as opposed to regular npm. That is something that some devs know about and expect to just work, and it didn't work prior to this PR. 

There could be some repo contributing guidelines to update, not sure if we need to - @yknl ?

hstove (approved):
This is great - exactly what we wanted to achieve by including `lib` but without the terrible git diffs.

yknl (approved):
Thanks @zone117x this is great!

PR #647, opened 2019-04-12
Minor doc changes

zone117x:
* Hide some misc objects from docs: social profile services, log levels.
* Corrected the type defs on some storage related options. Some boolean values (`encrypt`, `sign`) also support strings for more advanced usage. Updated the types and descriptions to reflect the code. 
* Some param description de-duplication and minor clarifications, using the option interfaces where available. 
* gitignore & remove `docs` directory, they clutter PRs if committed, and are now being generated and hosted on circleCI for PRs.


src/storage/index.ts
```diff
@@ -24,9 +24,23 @@ import { UserSession } from '../auth/userSession'
  * Specify a valid MIME type, encryption, and whether to sign the [[putFile]].
  */
 export interface PutFileOptions {
-  encrypt?: boolean | string,
-  sign?: boolean,
-  contentType?: string
+  /**
+   * Encrypt the data with the app public key. 
+   * If a string is specified, it will use used as the public key.
```
moxiegirl:
```
Encrypt the data with the app public key. 
   * If a string is specified, it is used as the public key. 
   * If the boolean `true` is specified then the current user's app public key is used. 
```

Typo in line 29 and use present tense in tech docs

src/storage/index.ts
```diff
@@ -24,9 +24,23 @@ import { UserSession } from '../auth/userSession'
  * Specify a valid MIME type, encryption, and whether to sign the [[putFile]].
  */
 export interface PutFileOptions {
-  encrypt?: boolean | string,
-  sign?: boolean,
-  contentType?: string
+  /**
+   * Encrypt the data with the app public key. 
+   * If a string is specified, it will use used as the public key. 
+   * If the boolean `true` is specified then the current user's app public key will be used.
```
moxiegirl: will be used > is used

src/storage/index.ts
```diff
@@ -24,9 +24,23 @@ import { UserSession } from '../auth/userSession'
...
+  /**
+   * Encrypt the data with the app public key. 
+   * If a string is specified, it will use used as the public key. 
+   * If the boolean `true` is specified then the current user's app public key will be used. 
+   * @default true
+   */
+  encrypt?: boolean | string;
+  /**
+   * Sign the data using ECDSA on SHA256 hashes with the user's app private key. 
+   * If a string is specified, it will be used as the private key.
```
moxiegirl: will be used > is used

src/storage/index.ts
```diff
@@ -387,46 +386,54 @@ function handleSignedEncryptedContents(caller: UserSession, path: string, stored
...
+   */
+  username?: string;
+  /**
+   * The app to lookup for multi-player storage - defaults to current origin. 
+   * @default `window.location.origin` 
+   * Only if available in the executing environment, otherwise `undefined`.
+   */
+  app?: string;
+  /**
+   * The URL to use for zonefile lookup. If falsey, this will use
```
moxiegirl, line 402:
```  
 The URL to use for zonefile lookup. If falsey, this uses
    the blockstack.js's [[getNameInfo]] function instead. 
```

Check that link to the function ...I didn't see it exposed in your artifact.  The link code is not working

src/storage/index.ts
```diff
@@ -387,46 +386,54 @@ function handleSignedEncryptedContents(caller: UserSession, path: string, stored
...
-  zoneFileLookupURL?: string | null
+export interface GetFileOptions extends GetFileUrlOptions {
+  /**
+   * Try to decrypt the data with the app private key.
+   * @default true
+   */
+  decrypt?: boolean;
+  /**
+   * Whether the content should be verified, only to be used 
+   * when [[putFile]] was set to `sign = true`.
```
moxiegirl: [[UserSession.putFie]] 

The code as written sends the user to the deprecated method

src/storage/index.ts
zone117x: I don't think `getNameInfo` is exposed anywhere. It's a method on the `BlockstackNetwork` class instance. Some methods on that class are wrapped and exposed but not this one. Should we expose the function, or remove the reference to it in the docs?

zone117x: Also the `[[getNameInfo]]` symbol reference was neither added nor broken in this PR. It existed previously (in another location), and has been a broken / useless reference for some time.

yknl (approved):
Looks good to me!

src/storage/index.ts
zone117x: This functions actually seems very useful IMO. I wrapped and exposed it for easy public use, and got the links fixed.

moxiegirl (approved):

PR #650, opened 2019-04-15
Fix several web worker bugs

zone117x:
Several functions using DOM / web-browser / `window` APIs fail if ran inside a web worker. 

This PR switches usages of `window` with a helper function that returns the universal global scope detected at runtime: [`self` or `window`](https://developer.mozilla.org/en-US/docs/Web/API/Window/self) which provides [`WindowOrWorkerGlobalScope`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope) for web-browsers, or `global` for certain types of Node.js polyfills that don't create the `window` object at the global scope. 

It should have the same behavior as before, where either an error or undefined is returned, depending on the execution environment (node.js vs web-browser). 

Came across this while debugging some issues with misthos (note: in their case it was just console errors and was not the actual problem they had). Any apps that offload storage operations (getFile / putFile) in web-workers seem most likely to run into the bug this PR fixes. 

Fixed remaining issues noted in https://github.com/blockstack/blockstack.js/issues/611

TODO: still need to run this through the browser integration tests.


hstove (approved):
I'm taking a look at this for the second time, and it looks good to me. Clever use of generics in the global object helper!

yknl (approved):
Looks good to me, thanks for the clear documentation!

friedger (approved):
Looks good to me. Android tests succeed.

PR #651, opened 2019-04-18
Add types for auth scopes

zone117x:
These are only typing & comment changes, should not result in any logic changes. 
Intention is to help devs see the available/common auth scopes in their editor intellisense / hinting.
It also adds more info to the generated docs aside from just `string` for the scope input.

### DIFF
--- a/src/auth/authConstants.ts
+++ b/src/auth/authConstants.ts
@@ -23,8 +23,8 @@ export const enum AuthScope {
   store_write = 'store_write',
   /**
    * Publish data so that other users of the app can discover and interact with the user.
-   * The user's files stored on Gaia are made visible to others via the apps property in the 
-   * user’s profile.json file. 
+   * The user's files stored on Gaia hub are made visible to others via the `apps` property in the 
+   * user’s `profile.json` file. 
    */
   publish_data = 'publish_data',
   /**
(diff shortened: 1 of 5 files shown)


src/auth/authConstants.ts
```diff
@@ -11,10 +11,34 @@ export const BLOCKSTACK_STORAGE_LABEL = 'blockstack'
...
+ */
+export const enum AuthScope {
+  /**
+   * Read and write data to the user's Gaia hub in an app-specific storage bucket.
+   * This is the default scope.
+   */
+  store_write = 'store_write',
+  /**
+   * Publish data so that other users of the app can discover and interact with the user.
+   * The user's files stored on Gaia are made visible to others via the apps property in the
```
moxiegirl: Never Gaia alone -- Gaia hub

is it `apps` property or apps' property

`profile.json` use the tic marks for filename, api names, properties.

moxiegirl (changes requested):
Couple little nits to fix

yknl (approved):
Looks good to me!

src/auth/authConstants.ts
zone117x: Fixed

moxiegirl (approved):

PR #652, opened 2019-04-19
Update to bitcoinjs-lib v5

zone117x:
Updated to bitcoinjs-lib v5 which has been ported to Typescript.
The bitcoinjs v5 dependency tree now references the native secp256k1 lib as an `optionalDependency`, so node-gyp errors will no longer fail the installs.


hstove (approved):
This all looks good to me, and most of the diff is just small (good) changes in how we're importing things. @zone117x have you seen any recent issues with v5?

zone117x:
@hstove 
> @zone117x have you seen any recent issues with v5?

I haven't seen any, I think we should merge this.

The bitcoinjs lib and dependency now reference the native secp256k1 lib as an optionalDependency, so node-gyp errors no longer fail installs.

yknl (approved):
Looks good to me 👍

PR #656, opened 2019-04-24
Generate dist script file usage with a cdn and the subresource integrity hash

zone117x:
Related to https://github.com/blockstack/blockstack.js/issues/584 and https://github.com/blockstack/blockstack.js/issues/560

Adds example of secure usage of the blockstack.js dist bundle from a cdn using the latest lib version and [Subresource Integrity hash](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity).

Example:
```html
<script src="https://unpkg.com/blockstack@19.1.0/dist/blockstack.js"
 integrity="sha384-iyP2Q91bL61TbHoDcBUQy5tXDFEhbswcT7syEyDLz9rlIa9hBJnd4YCt0vXFCbmK"
 crossorigin="anonymous"></script>
```

This example is generated on dist file build so that we don't have to add an extra manual step to our process (generating hash, setting version number, updating docs). 

@moxiegirl is going to take care of the doc work (something like linking to the example from the readme.md and generated typedoc, etc..)
EDIT: PR for that here: https://github.com/blockstack/blockstack.js/pull/657/files

### DIFF
--- a/mdincludes/script-dist-file.md
+++ b/mdincludes/script-dist-file.md
@@ -1,0 +1,3 @@
+```html
+<script src="https://unpkg.com/blockstack@19.1.0/dist/blockstack.js" integrity="sha384-iyP2Q91bL61TbHoDcBUQy5tXDFEhbswcT7syEyDLz9rlIa9hBJnd4YCt0vXFCbmK" crossorigin="anonymous"></script>
+```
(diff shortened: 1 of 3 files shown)


moxiegirl (approved):

yknl (approved):
Looks good, thanks @zone117x

PR #663, opened 2019-05-03
Prevent 'fetch' from leaking app urls in referrer header

zone117x:
Add the `Referrer-Policy: no-referrer` header to our usages of `fetch()`. 

This prevents potentially sensitive app URLs from being leaked via the http `referer` headers. 
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy

The Gaiahub and Blockstack API endpoints are the notable services that will no longer be leaked this data.

### DIFF
--- a/src/auth/authApp.ts
+++ b/src/auth/authApp.ts
@@ -4,6 +4,7 @@ import queryString from 'query-string'
 import { decodeToken } from 'jsontokens'
 import { verifyAuthResponse } from './authVerification'
 import { isLaterVersion, hexStringToECPair, checkWindowAPI } from '../utils'
+import { fetchPrivate } from '../fetchUtil'
 import { getAddressFromDID } from '../dids'
 import { LoginFailedError } from '../errors'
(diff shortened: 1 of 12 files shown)


yknl (approved):
Good catch, thanks for fixing this!

hstove (approved):
This looks great, but do you think there is a test we could create for `fetchPrivate`? Maybe we could add a global mock for `fetch` to ensure that errors if it's ever invoked without the `no-referrer` flag.

zone117x:
> This looks great, but do you think there is a test we could create for `fetchPrivate`? Maybe we could add a global mock for `fetch` to ensure that errors if it's ever invoked without the `no-referrer` flag.

I'd like to add `js-dom` as a dev-dependency to perform unit tests on this and other heavy DOM API functions. 

For ensuring that future commits are using `fetchPrivate` rather than `fetch`, I think we'd need to perform that using something like a custom eslint rule, or some other analysis at the source code level. Not sure how to enforce this at runtime without polluting the global scope.

PR #668, opened 2019-05-23
Fix user coreNode preference not respected

zone117x:
The user's core node / API endpoint was not being used for anything except during the `handlePendingSignIn()`.

The value was also not being stored in the localStorage userData. 

This PR adds the coreNode to the `UserData` object. The name lookup function now uses (in order of what is available):
1. The user's specified coreNode, if available (note: the authenticator currently always passes this value, even when it's the the default).
2. The app developer specified coreNode in `AppConfig`, if available. 
3. The default hardcoded endpoint. 

@jcnelson @kantai Can you confirm that the above behavior is intended? Prior to this PR, the an app could override the coreNode. And some of the code comments indicated that this was intentional. 

---

Also -- the `SessionOptions` object had several properties that were duplicated from the `UserData` object. Nothing appears to be using this properties so I removed them.

Its my understanding that these `UserSession` features are for app developers to be able to: 
A) Use the `UserSession.AppConfig` property to configure defaults and app-specific values.
B) Use the `UserSession.SessionDataStore` property to controls how `UserData` is stored (localStorage, in-memory, etc.)

And that the user-specified values inside `UserSession.SessionDataStore.UserData` values are used when available. 

Please let me know if this is the intended design.


yknl (approved):
LGTM

yknl:
@timstackblock can you please run a QA review on this PR? Let us know if you need clarifications on the changes.

timstackblock:
Sure no problem, can you please move this issue to the Review/QA column in the Zenhub board I cannot find it under blockstack/blockstack.js for some reason.

timstackblock:
> Its my understanding that these UserSession features are for app developers to be able to:
A) Use the UserSession.AppConfig property to configure defaults and app-specific values.
B) Use the UserSession.SessionDataStore property to controls how UserData is stored (localStorage, in-memory, etc.)

@zone117x   @yknl  can you please elaborate on the real world use cases, examples and acceptance criteria around the two points above. I want to make sure I fully recreate the scope of the change.

zone117x:
@timstackblock some more info one those two points --

> A) Use the UserSession.AppConfig property to configure defaults and app-specific values.

Our docs and example app have an example of instantiating an `AppConfig` instance https://docs.blockstack.org/browser/hello-blockstack.html#understand-the-generated-application-code

The various `AppConfig` properties, with their default values as examples, are documented here: https://blockstack.github.io/blockstack.js/classes/appconfig.html


> B) Use the UserSession.SessionDataStore property to controls how UserData is stored (localStorage, in-memory, etc.)

The `SessionDataStore` is not well documented since only the default implementation (the DOM localStorage API) is generally used, and most app devs don't touch it. However, there is an in-memory implementation -- some devs use this for things like unit testing within the Node.js runtime. Here's an example: https://forum.blockstack.org/t/creating-a-usersession-using-app-private-key/8096/6

---

In blockstack.js the `coreNode` variable is an http endpoint that is used for any APIs that need to communicate with the Blockstack network. For typical apps, it is used for Blockstack ID lookups in the `handlePendingSignIn` function and various storage APIs like `getFile(..., { user: 'some_user.id' })`. 

Prior to this PR, those functions were only using the hardcoded default endpoint in blockstack.js here: https://github.com/blockstack/blockstack.js/blob/0ee8292dfacfa7e28148d5ddbb8d92bc82092d35/src/auth/authConstants.ts#L53


During the authentication flow, the Blockstack Browser passes the user specified `coreNode` endpoint to the app. The user can change the default endpoint using the `Name Lookup URL` field in the Blockstack Browser settings page. An E2E test could specify the endpoint by mocking the auth token payload: https://github.com/blockstack/blockstack-browser/blob/8103c82f5b4ae24c8b1f4e52385326c0a5a60ce8/app/js/auth/index.js#L447-L458

Additionally, an app can also specify a coreNode when configuring the `UserSession`, see https://blockstack.github.io/blockstack.js/classes/appconfig.html#constructor

---

For the minimal acceptance criteria I think we should ensure:

1. That the current usage of blockstack.js in the various downstream projects are not broken (gaia, cli, browser, web app w/ auth & storage). 

2. Test that a non-default user-specified coreNode endpoint is used as expected. The PR implements the isolated unit tests, but there are no integration or e2e tests.

timstackblock:
I used the sample app https://github.com/yknl/publik and was able to successfully used this branch without issue. The information associated with the microblogging application persisted after  authentication. The regression test also passed.

timstackblock (approved):

yknl:
@zone117x is this ready to merge?

zone117x:
@jcnelson @kantai I edited the PR description and tagged you about some behavior that I want to confirm

kantai:
> Can you confirm that the above behavior is intended? Prior to this PR, the an app could override the coreNode. And some of the code comments indicated that this was intentional.

Yep -- there was some tension a long time ago about this issue, whether the app should specify the corenode or the user. I think that the behavior implemented here of the `user -> app -> default` precedence is great and should be what we implement.

kantai (approved):
This looks good to me, and I agree with the proposed order of precedence.

Issue #670, opened 2019-05-29
`GaiaHubConfig` is no longer exported from `index.ts`


MichaelFedora:
Using `blockstack.js` latest, v19.1.0

For some reason the interface `GaiaHubConfig` is no longer available via the normal
```ts
import { GaiaHubConfig } from 'blockstack';
```
but now has to be specified via
```ts
import { GaiaHubConfig } from 'blockstack/lib/storage/hub';
```

Was this on purpose and so the issue can be dismissed, or could it possibly be fixed for ease-of-use for those of us who make third party browsers and the like?

MichaelFedora:
Closed alongside #671 .

PR #671, opened 2019-06-02
Export GaiaHubConfig from storage/index.ts

MichaelFedora:
Solves #670, re:

> Using `blockstack.js` latest, v19.1.0
> 
> For some reason the interface `GaiaHubConfig` is no longer available via the normal
> ```ts
> import { GaiaHubConfig } from 'blockstack';
> ```
> but now has to be specified via
> ```ts
> import { GaiaHubConfig } from 'blockstack/lib/storage/hub';
> ```
> 
> Was this on purpose and so the issue can be dismissed, or could it possibly be fixed for ease-of-use for those of us who make third party browsers and the like?

### DIFF
--- a/src/storage/index.ts
+++ b/src/storage/index.ts
@@ -5,7 +5,6 @@ import {
   connectToGaiaHub, uploadToGaiaHub, getBucketUrl, BLOCKSTACK_GAIA_HUB_LABEL, 
   GaiaHubConfig
 } from './hub'
-// export { type GaiaHubConfig } from './hub'
 
 import {
   encryptECIES, decryptECIES, signECDSA, verifyECDSA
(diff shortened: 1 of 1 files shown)


yknl:
Hey @MichaelFedora, the `GaiaHubConfig` class was probably changed to private in the last major refactor. Is there a use case for this in your app?

MichaelFedora:
Sure:

https://github.com/MichaelFedora/unofficial-blockstack-extension/blob/b36bcd0984eec8299983fb8d8c515959e5cc2958/src/common/vuex/initial-store.ts#L7

Due note it's not an app, but a browser. I suppose if you wanted to keep it "private" that's fine, I can just keep importing it directly from the typings. Feel free to close if that's the case.

yknl:
Thanks @MichaelFedora, it might make more sense to just use the longer import if developers need it.

PR #675, opened 2019-06-23
Fix comment that doesn't match constructor

larrysalibra:
The comment on AppConfig says that on non-browser platform one needs to specify the app domain as the first argument to the constructor when it is actually the second. 

Trying to follow the instructions I wrote last year and finding bugs. :-) 

I'm not sure if you want to merge this hotfix into master or develop. Feel free to change.


yknl (approved):
Thanks for the fix!

PR #676, opened 2019-06-25
Update to typed `jsontokens` lib

zone117x:
Update the to the v2 version of `jsontokens` which is ported to Typescript, and its dependency `key-encoder`:
https://github.com/blockstack/key-encoder-js
https://github.com/blockstack/jsontokens-js

### DIFF
--- a/src/auth/authSession.ts
+++ b/src/auth/authSession.ts
@@ -135,33 +135,34 @@ export function getCoreSession(coreHost: string,
     return Promise.reject('No authRequest provided')
   }
 
-  let payload = null
-  let authRequestObject = null
   try {
-    authRequestObject = decodeToken(authRequest)
+    const authRequestObject = decodeToken(authRequest)
     if (!authRequestObject) {
       return Promise.reject('Invalid authRequest in URL query string')
     }
     if (!authRequestObject.payload) {
       return Promise.reject('Invalid authRequest in URL query string')
     }
-    payload = authRequestObject.payload
+    const payload = authRequestObject.payload
+    if (typeof payload === 'string') {
+      throw new Error('Unexpected token payload type of string')
+    }
+
+    const appDomain = payload.domain_name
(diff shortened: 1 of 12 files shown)


src/auth/authSession.ts
```diff
@@ -135,33 +135,34 @@ export function getCoreSession(coreHost: string,
     return Promise.reject('No authRequest provided')
   }
 
-  let payload = null
-  let authRequestObject = null
   try {
-    authRequestObject = decodeToken(authRequest)
+    const authRequestObject = decodeToken(authRequest)
```
yknl, line 139: Core sessions are no longer used, I think we should remove this code in a separate PR.

yknl (approved):

timstackblock (approved):

timstackblock:
What functionality does this directly impact, or is a regression test of general functionality only needed?

yknl:
This PR doesn't change any functionality. A regression test of the auth flow with this branch of blockstack.js in both the app and browser is what's needed.

timstackblock:
Thanks @yknl starting that now

timstackblock:
When trying to sign in with blockstack browser in the app I run into this issue please take a look at this, I am moving this issue back to in progress.  

Video of the issue
https://screencast-o-matic.com/watch/cqjlI8ORFv

Screen Shot 2019-08-12 at 9 57 28 AM

zone117x:
@timstackblock Server connectivity error looks unrelated. Can you verify the process you are using for testing the Browser with this blockstack.js branch?

timstackblock:
This works the error I was getting was from running in development mode and it happens on this branch and master its unrelated.

zone117x:
Okay great, were you able to verify that the browser e2e tests pass when the browser is using this branch?

timstackblock:
Yes just ran the test locally and we are good to go

PR #677, opened 2019-06-28
Revert "Fix comment that doesn't match constructor"

MHMYunos2020:
Reverts blockstack/blockstack.js#675


yknl:
@MHMYunos2020 can you explain why the change needs to be reverted?

yknl:
Closing this as it does not make sense.

Issue #678, opened 2019-07-02
Use promises for SessionDataStore


icidasset:
👋 

Any thoughts on making `SessionDataStore` asynchronous? I'm trying to make a store that uses indexedDB, but failing because currently it is expected that the data store is synchronous.

_Proposed solution:_  
Have the `get`, `set` and `delete` functions return a `Promise`.

_Context_:  
Trying to use Blockstack in a web worker, so no access to local storage.

_Help_:  
Can make a PR if you me want to.

zone117x:
Seems like a good idea. There are several synchronous code paths dependent on the current synchronous `SessionDataStore ` API. Some of the call graphs have async parent invocations so should be easy to update, but looks like at least one or two with a synchronous entry point which are public API functions. So changing these to return Promises would require a major version bump, and no clear way to remain backwards compatible.

icidasset:
Thanks @zone117x ! Might make a fork and a PR for this soon.

zone117x:
@icidasset That would be much appreciated. We already need some of these APIs to be async for future improvements in encryption methods, but don't currently have that prioritized.

PR #679, opened 2019-07-14
Make the SessionStore asynchronous

icidasset:
**Addresses #678**

_I want to use IndexedDB as a sessionStore,  
and that requires the sessionStore functions to be async._

---

Working build:
https://gist.github.com/icidasset/a888e02d7441aeb2af99263a3add0f73

---

Tests aren't passing yet, because I don't know how to fix them.
I'm getting a whole bunch of weird errors like the following:

```
not ok 247 test exited without ending
  ---
    operator: fail
    at: process.<anonymous> (/Users/icidasset/Other Code/blockstack.js/node_modules/tape/index.js:88:19)
    stack: |-
      Error: test exited without ending
          at Test.assert [as _assert] (/Users/icidasset/Other Code/blockstack.js/node_modules/tape/lib/test.js:225:54)
          at Test.bound [as _assert] (/Users/icidasset/Other Code/blockstack.js/node_modules/tape/lib/test.js:77:32)
          at Test.fail (/Users/icidasset/Other Code/blockstack.js/node_modules/tape/lib/test.js:318:10)
          at Test.bound [as fail] (/Users/icidasset/Other Code/blockstack.js/node_modules/tape/lib/test.js:77:32)
          at Test._exit (/Users/icidasset/Other Code/blockstack.js/node_modules/tape/lib/test.js:188:14)
          at Test.bound (/Users/icidasset/Other Code/blockstack.js/node_modules/tape/lib/test.js:77:32)
          at process.<anonymous> (/Users/icidasset/Other Code/blockstack.js/node_modules/tape/index.js:88:19)
          at process.emit (events.js:189:13)
          at processEmit (/Users/icidasset/Other Code/blockstack.js/node_modules/nyc/node_modules/signal-exit/index.js:149:35)
          at process.emit (/Users/icidasset/Other Code/blockstack.js/node_modules/source-map-support/source-map-support.js:465:21)
  ...
```


moxiegirl:
Thank you for the pull request @icidasset --  @zone117x Can you give @icidasset some guidance for fixing these tests?

src/auth/userSession.ts
```diff
@@ -205,20 +205,20 @@ export class UserSession {
...
 
   /**
    * Check if a user is currently signed in.
-   * 
+   *
    * @returns {Boolean} `true` if the user is signed in, `false` if not.
    */
-  isUserSignedIn() {
-    return !!this.store.getSessionData().userData
+  isUserSignedIn(): Promise<boolean> {
```
zone117x, line 220: This sync -> async API change will probably break every existing app using Blockstack.js. Even with a major semver bump, I anticipate a lot of frustration. 

Should this use the `@deprecated` tag and console warning, and continue to work for synchronous SessionStorage implementations? A new async-only implementation could throw a "not supported" Error. 

Then something like `isUserSignedInAsync`, or function overloading e.g. `isUserSignedIn({async: true}): Promise<bool>`. 

Not sure what to do here..

src/auth/userSession.ts
```diff
@@ -229,19 +229,22 @@ export class UserSession {
...
   /**
    * Retrieves the user data object. The user's profile is stored in the key [[Profile]].
-   * 
-   * @returns {Object} User data object.
+   *
+   * @returns {Promise} User data object.
    */
-  loadUserData() {
-    const userData = this.store.getSessionData().userData
+  async loadUserData(): Promise<UserData> {
```
zone117x: Same concern here as above comment about API breakage

src/auth/userSession.ts
```diff
@@ -372,26 +375,25 @@ export class UserSession {
...
 
     userData.gaiaHubConfig = gaiaConfig
 
-    const sessionData = this.store.getSessionData()
+    const sessionData = await this.store.getSessionData()
     sessionData.userData.gaiaHubConfig = gaiaConfig
-    this.store.setSessionData(sessionData)
 
-    return gaiaConfig
+    return this.store.setSessionData(sessionData).then(_ => gaiaConfig)
```
zone117x, line 397: Nit: new async code should use `await` rather than promise chaining when possible

tests/unitTests/src/unitTestsAuth.ts
```diff
@@ -461,7 +470,9 @@ export function runAuthTests() {
 
     const appConfig = new AppConfig(['store_write'], 'http://localhost:3000')
     const blockstack = new UserSession({ appConfig })
-    blockstack.store.getSessionData().transitKey = transitPrivateKey
+
+    const sessionData = await blockstack.store.getSessionData()
+    sessionData.transitKey = transitPrivateKey
 
     blockstack.handlePendingSignIn(authResponse)
```
zone117x: This may be contributing to the failing tests, RE:
> Error: test exited without ending

Since this function is now marked `async` I think this `blockstack.handlePendingSignIn` needs to be `await`ed. Also, the `t.plan` call above may be confusing the test lib, try removing it.

zone117x: Same with any other tests that are now marked as `async` -- I think the `t.plan(n)` statements may need to be removed.

tests/unitTests/src/unitTestsStorage.ts
```diff
@@ -890,7 +902,7 @@ export function runStorageTests() {
       .then(() => t.ok(true, 'Request should pass'))
   })
 
-  test('getFileUrl', (t) => { 
+  test('getFileUrl', async (t) => {
```
zone117x: Marked `async` but no `await` statements or promises returned, probably causes a breaking test

src/auth/userSession.ts
icidasset: Yeah, there's no way around it, we have to make this async. Because the underlying `SessionDataStore` class is async, this function must be async. There's no way to take a `Promise` and make it synchronous (ie. block the thread and wait for the result).

I don't know how other people structure their apps, but in my case it wasn't that difficult to make this async. So I'm not sure how much frustration there would be. You probably have a better view on this though.

I can just use my own fork for the time being. But I assume you'll want to merge this at some point?

tests/unitTests/src/unitTestsAuth.ts
icidasset: Yeah, this is difficult, this `tape` library is not giving me anything.
I tried removing all the `t.plan` statements, but nothing 🤔 
Will try again later.

src/encryption/ec.ts
```diff
@@ -119,13 +119,13 @@ export function encryptECIES(publicKey: string, content: string | Buffer): Ciphe
   )
 
   const macData = Buffer.concat([initializationVector,
-                                 Buffer.from(ephemeralPK.encodeCompressed()),
+                                 Buffer.from(ephemeralPK.encodeCompressed('array') as any),
```
icidasset, line 122: @zone117x Are you getting type errors here as well? I had to do this to be able to make a build.

icidasset: Type error was that `encodeCompressed` expects one argument and the `as any` is necessary to avoid the `BN can't be cast to a string` type error. Which is weird because I don't think `Buffer.from` expects a string 😕

src/auth/userSession.ts
zone117x: I think the best approach here is to keep all of these functions sync by default, and require a `opts: {async: true}` param to take the async path and return a Promise. Typescript function overloading should be helpful here. 

When a backing SessionStorage implementation does not support sync, it should throw a useful Error stating that the `{async: true}` param should be specified. 

This gives us an easy path in the future to start deprecating the sync version, then eventually remove it once we feel enough devs have updated.

src/auth/userSession.ts
zone117x: Same with this function as the above comment -- default to sync, but take async path when optional param is provided.

tests/unitTests/src/unitTestsAuth.ts
zone117x: Don't worry too much about the tests -- once we get the implementation / API surface down, I can help get the tests working.

src/encryption/ec.ts
zone117x: The typings in the `@types/elliptic` package are incomplete and incorrect in some areas. In this case, I don't think `encodeCompressed` even returns a `BN`. If you merge the develop branch into this PR, it is updated to use a different function with more accurate typings. Something like `Buffer.from(ephemeralPK.encode('array', true) as Buffer)`.

One of my goals is to move us away from the `elliptic` lib all together at some point. For now I recommend going off the actual `elliptic` lib source and forcing the types as necessary.

PR #682, opened 2019-07-23
Throw error on user session overwrite

yknl:
`handlePendingSignIn` now throws an error when a user session already exists in local storage. Addresses https://github.com/blockstack/blockstack.js/issues/680

Testing:
1. Use any Blockstack app and link with this branch
2. Sign in to app 
3. Go back to sign in page on app and click sign in again, choose identity to use
4. App should error out

### DIFF
--- a/tests/unitTests/src/unitTestsAuth.ts
+++ b/tests/unitTests/src/unitTestsAuth.ts
@@ -389,7 +389,6 @@ export function runAuthTests() {
         t.fail('Should not overwrite existing user');
       })
       .catch((err) => {
-        console.log(err.stack)
         t.pass('Should throw error when overwriting existing user session during handle pending sign in')
       })
   })
--- a/tests/unitTests/src/unitTestsAuth.ts
+++ b/tests/unitTests/src/unitTestsAuth.ts
@@ -352,6 +352,47 @@ export function runAuthTests() {
       })
   })
 
+  test('handlePendingSignIn with existing user session', (t) => {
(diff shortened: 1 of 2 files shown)


friedger:
The documentation should suggest how to handle this exception.

tests/unitTests/src/unitTestsAuth.ts
```diff
@@ -352,6 +352,47 @@ export function runAuthTests() {
...
+    const authResponse = makeAuthResponse(privateKey, sampleProfiles.ryan, 'ryan.id',
+                                          metadata, undefined, appPrivateKey, undefined,
+                                          transitPublicKey)
+
+    blockstack.handlePendingSignIn(authResponse)
+      .then(() => {
+        t.fail('Should not overwrite existing user');
+      })
+      .catch((err) => {
+        console.log(err.stack)
```
hstove: Let's remove this, especially since this is the 'expected' code path

hstove (changes requested):
Tiny change requested, otherwise LGTM!

zone117x (approved):
This probably fixes most of these cases and we should merge, but I think there are more situations that would cause similar issues described in https://github.com/blockstack/blockstack.js/issues/680

Several tabs could be open and logged in for an app. User could logout (clear localStorage) on tab 1, then login with different account, and still remained logged in with original account in tab 2, potentially causing issues depending on how the app is designed. This could be addressed in another PR.

PR #684, opened 2019-08-01
[WIP] Collections API

yknl:
Adds collection storage operations:
`getCollectionItem`
`putCollectionItem`
`listCollectionFiles`
`deleteCollectionFile`

Gaia hub config and encryption keys for collection storage operations are fetched from the app storage bucket and cached.


src/auth/userSession.ts
```diff
@@ -398,4 +400,58 @@ export class UserSession {
 
     return gaiaConfig
   }
+
+  /**
+   * Retreive a collection Gaia hub config from local user data
+   * @param collection - the name of the collection to get hub config for
+   */
+  getCollectionGaiaHubConnection(collectionName: string): Promise<GaiaHubConfig> {
```
zone117x: I think there is a bug somewhere around here where the non-collection bucket gets used after a Gaia storage request fails, and the error handler attempts to re-negotiate the gaia hub connection. 

Saw this when testing collection list files while using a faulty gaia hub server.

zone117x (commented):
I've been testing this out using UMD dist bundle ( https://unpkg.com/blockstack@20.0.0-alpha.1/dist/blockstack.js ) with a collection demo app at https://blockstack.github.io/blockstack-collections/page_test/

Seems to be mostly working except for the comment I left about gaia hub re-connections seeming to use the non-collection bucket, and the collection key not working when used across different apps.

src/storage/index.ts
```diff
@@ -648,14 +683,15 @@ export async function putFile(
...
+    wasSigned?: boolean,
+    gaiaHubConfig?: GaiaHubConfig
   },
   caller?: UserSession
 ) {
   if (!caller) {
     caller = new UserSession()
   }
-  const gaiaHubConfig = await caller.getOrSetLocalGaiaHubConnection()
+  const gaiaHubConfig = options.gaiaHubConfig || await caller.getOrSetLocalGaiaHubConnection()
```
zone117x: `options` can be undefined and cause this line to throw. Can be reproduced with `userSession.deleteFile('anything')`

zone117x (changes requested):

PR #687, opened 2019-08-07
Exposing the Blockstack network

moxiegirl:
Exposing the Blockstack network. Adding the caveat to the getName* "T…his is intended for use in third-party wallets or in DApps that register names."  Also, this exposes AuthScope.

Signed-off-by: Mary Anthony <mary@blockstack.com>

### DIFF
--- a/src/network.ts
+++ b/src/network.ts
@@ -8,14 +8,25 @@ import { Logger } from './logger'
 import { config } from './config'
 import { fetchPrivate } from './fetchUtil'
 
-/**
- * @ignore
- */
-export type UTXO = {
-  value?: number,
-  confirmations?: number,
-  tx_hash: string,
-  tx_output_n: number
+export interface UTXO {
+  value?: number;
+  confirmations?: number;
+  tx_hash: string;
+  tx_output_n: number;
+}
+
+export interface PriceInfo {
+  units: string;
+  amount: BN;
+}
+
+export interface AccountTokens {
+  tokens: string[];
+}
+
+export interface TransactionInfo {
+  block_height: number;
+  [key: string]: any;
 }
 
 const SATOSHIS_PER_BTC = 1e8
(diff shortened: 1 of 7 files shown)


moxiegirl:
@kantai you may want to check ou the exposed network methods: https://1667-22058375-gh.circle-artifacts.com/0/docs/classes/blockstacknetwork.html

moxiegirl:
@zone117x I talked to @kantai and I need to tweak the documentation to remove the `layer1` properties. Also need to add some additional docs to the other APIs from our discussion.

shreyasthiagaraj:
This looks good to me!

shreyasthiagaraj (approved):
This looks good to me. The only nit is that `getNetworkedUTXOs` and `getTransactionInfo` don't seem to have any documentation, but adding that may be outside the scope of this PR.

moxiegirl:
@zone117x I did a rebase and force push --- bit different from a straight merge.  If you want me to, I  can go back out to your last commit.

src/network.ts
```diff
@@ -98,19 +113,20 @@ export class BlockstackNetwork {
   }
 
   /**
-  * @ignore
-  */
+   * This is intended for use in third-party wallets or in DApps that register names.
```
yknl, line 116: Should we have to indicate this? Developers should be able to get an idea of what the function does and whether they need to care about it from reading the description right?

moxiegirl: Is there a reason we shouldn't be explicit about what a class is for? I get a lot of questions from newbs that require being explicit; not every "reads" code so easily. Kind of the reason we write code comments.

yknl: I think it's explicit in describing the functionality, but listing use cases seems out of place in reference documentation.

moxiegirl: Use cases and other statements like this appear frequently in reference documentation. 

- https://docs.ubuntu.com/core/en/reference/interfaces/
- https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIAppStartup
- https://help.sap.com/viewer/c5bc243e9e824373a5e8de7c37b1dee1/Cloud/en-US/bdde103263394fdbbf7eee440a9e61ba.html
- https://documenter.getpostman.com/view/220187/super-tech-heroes-api/77cf6KB
- https://i.materialise.com/api-dokuwiki/doku.php?id=delivery-note-retrieval-by-order-id-api
- https://developer.avalara.com/api-reference/avatax/rest/v2/methods/Customers/ListValidCertificatesForCustomer/
- https://docs.pulpproject.org/dev-guide/integration/rest-api/status.html

"Use case" statements save time and reduce support calls because people don't have to go to forums or Slack or into code to discern how to use code.

You may not be aware, StackOverflow has done survey's for years; insufficient API documentation was in the top three frustration for developers for so many years that Stack Overflow created a documentation site just to solve this frustrating problem. (It is a very interesting case study in itself. Happy to give you more details and references, as I have a professional interest in educating myself about what makes good documentation).

yknl (approved):
Looks good to me, will defer to you on the reference documentation.

PR #692, opened 2019-08-12
Fix deprecated crypto warning in Node.js usage

zone117x:
Fixes https://github.com/blockstack/blockstack.js/issues/681

Synthetic default imports were causing the Node `crypto` module properties to be enumerated at runtime, triggering the console warnings:
```
(node:14634) [DEP0091] DeprecationWarning: crypto.DEFAULT_ENCODING is deprecated.
(node:14634) [DEP0010] DeprecationWarning: crypto.createCredentials is deprecated. Use tls.createSecureContext instead.
(node:14634) [DEP0011] DeprecationWarning: crypto.Credentials is deprecated. Use tls.SecureContext instead
```

This PR disables the typescript `esModuleInterop` feature which allowed for default import statements (e.g. `import x from 'x'`) for modules that do not specify a default export. 

---

The JS output used to contain: 
```js
var __importStar = (this && this.__importStar) || function (mod) {
    if (mod && mod.__esModule) return mod;
    var result = {};
    if (mod != null) for (var k in mod) if (Object.hasOwnProperty.call(mod, k)) result[k] = mod[k];
    result["default"] = mod;
    return result;
};
const crypto = __importStar(require("crypto"));
```

This now now looks like:
```js
const crypto = require("crypto");
```

---

Tested with:
Pull cli-blockstack repo, point `package.json` to this branch:
```
"blockstack": "blockstack/blockstack.js#feature/deprecated-crypto-warning",
```
Run
```
npm i
npm link
blockstack-cli address 1Nw25PemCRv24UQAcZdaj4uD11nkTCWRTE
```


timstackblock (approved):

timstackblock:
looks good

yknl (approved):
Thanks @zone117x Looks good to me!

PR #718, opened 2019-09-03
Add a 'Proof' interface for services

expenses:
Add a `Proof` interface for services to formalise the API.

@yknl 

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/profiles/services/proof.ts
+++ b/src/profiles/services/proof.ts
@@ -1,5 +1,5 @@
 export interface Proof {
-  proof_url: string;
+  proofUrl: string;
   identifier: string;
   service: string;
   valid: boolean;
--- a/src/profiles/services/proof.ts
+++ b/src/profiles/services/proof.ts
@@ -1,0 +1,6 @@
+export interface Proof {
+  proof_url: string;
+  identifier: string;
+  service: string;
+  valid: boolean;
+}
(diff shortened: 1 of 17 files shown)


src/profiles/services/proof.ts
```diff
@@ -0,0 +1,6 @@
+export interface Proof {
+  proof_url: string;
```
friedger: Re modelling, this could be "proofUrl".

expenses: True, not sure why the original code was using underscores. Is there a reason `@typescript-eslint/camelcase` is turned off?

expenses:
I've renamed the variable. Do you think https://github.com/blockstack/blockstack.js/blob/master/src/profiles/profileSchemas/personUtils.ts#L114 will need to be changed too?

friedger:
@expenses This ( https://github.com/blockstack/blockstack.js/blob/master/src/profiles/profileSchemas/personUtils.ts#L114) doesn't need to be changed. It is just good for backwards compatibility!

friedger:
@yknl Looks good to me...

yknl (approved):

moxiegirl:
@expenses Can you provide some documentation for how you intend this API to be used?

PR #719, opened 2019-09-03
[WIP] Replace `schema-inspector` with `ajv`

expenses:
Following up on a `TODO` item.

@zone117x 

## Checklist
- [x] `npm run test` passes

### DIFF
--- a/src/profiles/profileSchemas/person.ts
+++ b/src/profiles/profileSchemas/person.ts
@@ -1,5 +1,4 @@
-// @ts-ignore: Could not find a declaration file for module
-import * as inspector from 'schema-inspector'
+import * as Ajv from 'ajv'
 
 import { Profile } from '../profile'
 import { extractProfile } from '../profileTokens'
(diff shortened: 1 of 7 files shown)


src/profiles/profileSchemas/person.ts
```diff
@@ -111,7 +110,7 @@ export class Person extends Profile {
 
   static validateSchema(profile: any, strict = false) {
     schemaDefinition.strict = strict
-    return inspector.validate(schemaDefinition, profile)
+    return new Ajv().validate(schemaDefinition, profile)
```
zone117x, line 113: A few of the schema validation tests are failing -- the tape unit test error messages are not that helpful by default:
```
not ok 120 Person profile should be valid
  ---
    operator: ok
    expected: true
    actual:   undefined

not ok 120 Person profile should be valid
  ---
    operator: ok
    expected: true
    actual:   undefined

not ok 138 Profile should be in a valid format
  ---
    operator: ok
    expected: true
    actual:   false

```

Looks like the `avj` config and return properties are different. For example this `strict` option appears to be specified in a different way in avj.

And the validation return object has different properties. See this code from the unit tests: 

```
    const validationResults = Profile.validateSchema(sampleProfiles.naval)
    t.ok(validationResults.valid, 'Profile should be valid')
```
I don't think `result.valid` is a defined property. And looks like there are some options for reporting better error messages (from avj readme):
```js
var valid = ajv.validate(schema, data);
if (!valid) console.log(ajv.errors);
```

zone117x (changes requested):
Thanks for tackling this @expenses. Will be great to remove this redundant dependency. 

The `avj` validation return object differs from what the unit tests expect -- looks like that should be an easy fix.

Issue #413, opened 2018-04-09
Error Handling: Automate error reporting


jcnelson:
We should consider making a unified `Error()` exception that encodes the following:

* A stack trace to the point in the code where it occurred
* A URL that, when clicked, will submit the stack trace anonymously to a bug reporter

The high-level goal is that if a user's app encounters a bug, and they look in the console and see this error, they'll see a helpful message saying something like "click here to report this bug to Blockstack", followed by a URL that encodes the error.

The URL should resolve to a form with a bug report, pre-filled with the data gathered from the stack trace.  The user would have a chance to consent to submit it.

kantai:
Agreed. This can be really great for getting detailed error reports -- the idea being that we'd just prefill values using URL strings on the form: blockstack.org/bugs

A good place to start would be to look at how other frameworks have implemented this (in particular, I think there are tricks to making the console printout highly visible), and then try to emulate.

hstove:
React does a pretty good job with error messages and warnings. I'm pretty sure if you just include a URL in the message, it gets formatted properly.

Here is an example of them creating [an `Error`](https://github.com/facebook/react/blob/94518b068bf52196abea0c83f4c9926bfe2065c6/packages/shared/invokeGuardedCallback.js#L163), and [a warning](https://github.com/facebook/react/blob/27535e7bfcb63e8a4d65f273311e380b4ca12eff/packages/react-dom/src/server/ReactPartialRenderer.js#L522).

I'm happy to help with this one.

yknl:
Errors have been improved in https://github.com/blockstack/blockstack.js/pull/720
Although there isn't an automatically generated error reporting URL.

PR #720, opened 2019-09-03
Better error responses

### LINKED ISSUE #436: Error handling: improve when decryption fails
We need better error handling when decryption fails (e.g., if the file wasn't encrypted in the first place).

This has been experienced more frequently since the upgrade where we switched to encrypt-by-default (see issue #434).

### LINKED ISSUE #601: getFile not rejecting the Promise
[Blockstack docs (`18.2.0`)](http://blockstack.github.io/blockstack.js/#getfile) states that `getFile` :

> resolves to the raw data in the file or rejects with an error

Nevertheless in practice it does resolve the `Promise` and returns a `null` object instead, thus never getting into the `.catch(` statement.

Even the DEBUG log says it so:
> [DEBUG] getFile my/file.json returned 404, returning null

### LINKED ISSUE #665: putFile(), deleteFile(), and listFiles() need to return the HTTP status on error
Because a POST can fail due to rate-limiting or authentication or other errors, the error paths for these methods should at the very least report the specific HTTP error encountered. Right now, only a generic exception is thrown.

shreyasthiagaraj:
As a Blockstack developer I'd like to get better error codes when operations fail. This PR implements https://github.com/blockstack/blockstack.js/pull/442 in Typescript as a starting point.

- Fixes #436 with a DecryptionFailedError
- Adds a proper ERROR_CODE to all instances of BlockstackException
- #413 by adding a URL to exception messages with a pre-filled URL that points to blockstack.org/bugs with some HTTP parameters that auto-fill the bug form.
- Fixes #665 by returning HTTP response code + error description for storage methods.
- Includes json/string information returned from Gaia in error responses.
- Fixes #601 by returning error instead of null when file is not found.


moxiegirl:
@shreyasthiagaraj I believe @timstackblock  wants the bugs filed in GitHub not the forum --- which is where your PR is going to send them. Tim, I may have misunderstood you.

shreyasthiagaraj:
@moxiegirl I don't understand. My PR will send bugs to the forum?

moxiegirl:
> @moxiegirl I don't understand. My PR will send bugs to the forum?

Yes, @shreyasthiagaraj . if you take the link you refer to in the description (`blockstack.org/bugs`) and in the code, the link redirects to this: https://forum.blockstack.org  which is the forum.

src/storage/hub.ts
```diff
@@ -51,7 +52,8 @@ export async function uploadToGaiaHub(
     }
   )
   if (!response.ok) {
-    throw new Error('Error when uploading to Gaia hub')
+    const responseDescription = await getResponseDescription(response)
```
zone117x: I know this is still on your PR TODO list -- but here is the list of Gaia hub upload error responses: https://github.com/blockstack/gaia/blob/ff1cd245e4df7244e8c6e32f08e95ec2aba9834e/hub/src/server/http.ts#L68-L79

I think we will probably want corresponding error codes and class types that match those.

src/errors.ts
```diff
@@ -32,8 +39,22 @@ export class BlockstackError extends Error {
   parameter?: string
 
   constructor(error: ErrorType) {
-    super(error.message)
-    this.message = error.message
+    let message = error.message
+    let bugMessage = `Error Code: ${error.code}`
+    let stack = null
```
zone117x: `Error.stack` is [weird](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack). Depending on the runtime, I think typically `this.stack` will already be set, and closer to the offending lines.

It might be better for this code to look something like 
```js
let stack = this.stack
if (!stack) { 
  try {
    throw new Error()
  } catch (e) {
     stack = e.stack
  }
}
```

src/encryption/ec.ts
```diff
@@ -3,6 +3,7 @@ import { ec as EllipticCurve } from 'elliptic'
 // @ts-ignore
 import * as BN from 'bn.js'
 import * as crypto from 'crypto'
+import { FailedDecryptionError } from './errors'
```
zone117x: Looks like this is failing tests. I think it needs to be 
```ts
import { FailedDecryptionError } from '../errors'
```

zone117x (changes requested):
Looking good so far! I left some comments for potential changes & fixes

src/errors.ts
shreyasthiagaraj: Unfortunately `this.stack` is inaccessible until I do `super(message)`, where `message` requires the stack :-/.

src/encryption/ec.ts
shreyasthiagaraj: Oops, fixed 👍

src/storage/hub.ts
shreyasthiagaraj: Done 👍

src/errors.ts
zone117x: You should be able to call `super(error.message)` or just `super()` at the beginning of the constructor and the later `this.message` reassignment should work as intended

src/storage/hub.ts
zone117x: Looking good! Couple minor change requests:
* The Gaia has a [`413 PayloadTooLarge`](https://github.com/blockstack/gaia/blob/ef0771833cd56200e08e18631e62467f24e086af/hub/src/server/http.ts#L78) error code _(it was recently introduced, you didn't miss it)_. 
* The [`delete` file endpoint](https://github.com/blockstack/gaia/blob/ef0771833cd56200e08e18631e62467f24e086af/hub/src/server/http.ts#L120-L131) can also return many of these same errors, in addition to a [specific 404 error](https://github.com/blockstack/gaia/blob/ef0771833cd56200e08e18631e62467f24e086af/hub/src/server/http.ts#L127). The blockstack.js function for that is located right underneath this one, see `function deleteFromGaiaHub`.. 
* Same thing for the [`list-files` endpoint](https://github.com/blockstack/gaia/blob/ef0771833cd56200e08e18631e62467f24e086af/hub/src/server/http.ts#L172). See `function listFilesLoop`.. which is in `src/storage/index.ts`.

All these fetch requests to the Gaia server should probably use a common error handling function, i.e. all `fetch` requests involving `GaiaHubConfig.server`.

zone117x (changes requested):

src/storage/hub.ts
```diff
@@ -51,8 +51,20 @@ export async function uploadToGaiaHub(
     }
   )
   if (!response.ok) {
-    throw new Error('Error when uploading to Gaia hub')
-  } 
+    const responseDescription = await getResponseDescription(response)
+    const errorMsg = `Error when uploading to Gaia hub. ${responseDescription}`
```
zone117x: Gaia returns useful JSON error payloads for many error conditions. The function that provides this `responseDescription` value uses `response.text()`, so the `Error` object thrown will have a message string with embedded stringified json. This makes it hard for blockstack.js and apps to perform condition-specific error handling. 

It would be useful if these Gaia response errors contained the hub error response data. For example, the errors could extend something like:
```ts
interface GaiaHubErrorResponse {
  status: number, 
  statusText: string,
  body?: string | any
}

class GaiaHubError extends BlockstackError {
  hubError: {
    message?: string
    error?: string
    statusCode: number
    statusText: string
    [prop: string]: any
  }

  constructor(error: ErrorType, response: GaiaHubErrorResponse) {
    super(error)
    this.hubError = {
      statusCode: response.status,
      statusText: response.statusText
    }
    if (typeof response.body === 'string') {
      this.hubError.message = response.body
    } else if (typeof response.body === 'object') {
      Object.assign(this.hubError, response.body)
    }
  }
}
```

The `getResponseDescription` function could return the JSON if available, something like:
```ts
async function getGaiaErrorResponse(response: Response): Promise<GaiaHubErrorResponse> {
  let responseMsg = ''
  let responseJson: any | undefined
  try {
    responseMsg = await response.text()
    try {
      responseJson = JSON.parse(responseMsg)
    } catch (error) { 
      // ignore... server did not return JSON, use string response
    }
  } catch (error) {
    Logger.debug(`Error getting bad http response text: ${error}`)
  }
  return {
    status: response.status,
    statusText: response.statusText,
    body: responseJson || responseMsg
  }
}
```

src/errors.ts
shreyasthiagaraj: Good idea. Done :).###

src/storage/hub.ts
shreyasthiagaraj: Done! Implemented changes based on your suggestions. Thanks!

shreyasthiagaraj: `getBlockstackErrorFromResponse` in `utils.ts` is the central place to process gaia hub errors.

src/storage/hub.ts
shreyasthiagaraj: Done--`getBlockstackErrorFromResponse` in utils.ts is the central place to process gaia hub errors. Thanks!

shreyasthiagaraj:
Hey!

shreyasthiagaraj:
@zone117x I fixed 1 test that was obviously the result of my changes, but there are 10 more falling that seem totally unrelated. Is this a known issue, or should all tests currently be passing? Thanks!

src/storage/hub.ts
```diff
@@ -72,20 +74,7 @@ export async function deleteFromGaiaHub(
...
-    }
-    const errorMsg = 'Error deleting file from Gaia hub: '
-      + `${response.status} ${response.statusText}: ${responseMsg}`
-    Logger.error(errorMsg)
-    if (response.status === 404) {
-      throw new FileNotFound(errorMsg)
-    } else {
-      throw new Error(errorMsg)
-    }
+    throw getBlockstackErrorFromResponse(response, 'Error deleting file from Gaia hub.')
```
zone117x: I think this needs an `await`

zone117x (changes requested):

src/storage/hub.ts
zone117x: That may explain some of the failing tests, a few `UnhandledPromiseRejectionWarning: Unhandled promise rejection` warnings in the logs.

shreyasthiagaraj: Looking closer at the failures I found that there was 1 failure caused by my change to `getFile` behavior (missing files throw `DoesNotExist` error instead of returning null), and the rest were failing because that one failed. Once I fixed this test the rest all started passing 😀.

shreyasthiagaraj:
Fixed all failing tests, and modified the `getFile throw on 404` test to check some of the values that the newly introduced `GaiaHubError` should contain. Let me know if it looks good, thanks!

src/storage/index.ts
```diff
@@ -230,34 +230,25 @@ export async function getFileUrl(
...
-        return response.text()
-      } else {
-        return response.arrayBuffer()
-      }
-    })
+async function getFileContents(path: string, app: string, username: string | undefined, 
+                               zoneFileLookupURL: string | undefined,
+                               forceText: boolean,
+                               caller?: UserSession): Promise<string | ArrayBuffer | null> {
+  await Promise.resolve()
```
zone117x: This line can be removed, now that the function is marked `async` it is implicitly wrapped in a `Promise.resolve`

zone117x:
This looks great! We should note in the changelog of breaking change: `getFile` throws error on 404 rather than returning null.

Adding @timstackblock as reviewer to perform the regular smoke tests. 

We should also test something like `getFile('bogus123_not_exists')` results in a useful console log message.

zone117x (approved):
Left a last comment where a line of code can be removed. Other than that, code looks good.

timstackblock (approved):

yknl (approved):
This looks good to me 👍

Issue #717, opened 2019-09-03
`npm run test` output is overly verbose


expenses:
Hi, I'm interested in contributing to this repo, but I'm finding the test output way too verbose. Is there a way to output only what tests failed?

expenses:
Perhaps we could use something like https://www.npmjs.com/package/tap-nyc ?

PR #722, opened 2019-09-06
Use tap-min for less verbose test output

expenses:
Hi, I found the default `tape` test output very verbose and inconvenient (#717), so I've added on `tap-min` to display it in a less verbose way (it now just prints lines that errored). It still prints a lot of warnings like `DEPRECATED: TransactionBuilder sign method arguments will change in v6, please use the TxbSignArg interface` though, I'll try and get rid of those at some point.


package.json
```diff
@@ -12,7 +12,7 @@
     "build:node": "shx rm -rf lib && tsc -b tsconfig.build.json",
     "build:node:watch": "shx rm -rf lib && tsc -b tsconfig.build.json --watch",
     "prepare": "npm run build:node && npm run build:webpack",
-    "test": "npm run lint && cross-env NODE_ENV=test nyc node ./tests/unitTests/src/index.ts",
+    "test": "npm run lint && cross-env NODE_ENV=test nyc node ./tests/unitTests/src/index.ts | tap-min; nyc report",
```
zone117x, line 15: I think the pipe operator `|` here is cross-platform (works on Windows cmd.exe and Powershell), but the `;` operator is definitely not. I think the [npm-run-all](https://www.npmjs.com/package/npm-run-all) tool could be used to perform this same thing in a cross platform way.

zone117x: In general, I think less verbose test output can be useful. I have actually ran into the need to increase my terminal line buffer length to see the full output with large tape test suites. 

However, I don't think it should be the default `test` command. The verbose output is useful for inspecting failed circleCI tests, among other things. 

Can this be made into a separate command, something like `test:min`?

zone117x (changes requested):

PR #723, opened 2019-09-16
Enable decryption with custom key

yknl:
## Description

As a developer I would like to be able to decrypt files stored on Gaia using a custom private key instead of the app private key in the user session. In the `getFile` function's options argument, you can now specify a private key in the `decrypt` field. This key will be used to decrypt the file. 

```
let options = { decrypt: 'a5c61c6ca7b3e7e55edee68566aeab22e4da26baa285c7bd10e8d2218aa3b229' }
getFile(path, options).then((decryptedFile) => {
  // decrypted file
})
```

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

Yes, the API reference has been updated.

## Testing information
Things to test for:
1. A file encrypted using `putFile` with a custom public key can be decrypted using `getFile` with the corresponding private key.
2. `getFile` and `putFile` without custom encryption key works properly

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated


zone117x (approved):
LGTM!

zone117x:
Adding @timstackblock as reviewer to perform smoke tests

timstackblock (approved):

PR #724, opened 2019-09-16
Content length check before upload

zone117x:
## Description

As developer, I would like blockstack.js to check the Gaia hub's maximum file upload size when attempting `putFile`, and for it to throw a useful `Error` and/or size info that an app can handle appropriately. 

* Closes https://github.com/blockstack/blockstack.js/issues/636
* Dependent on Gaia features in PR https://github.com/blockstack/gaia/pull/256

* Partially addresses https://github.com/blockstack/blockstack.js/issues/705 `Reduce encrypted file size`

This makes the upload function check the content size against the max upload file size returned from `/hub_info`. 

When encryption is enabled, the expected payload byte size is calculated before having to actually encrypt the data:
 * The AES-CBC ciphertext output size is deterministically calculated using the input byte size. 
 * The JSON envelops for the ciphertext (both signed and unsigned versions) have their stringified length calculated. 

### New encoding option for file size reduction

The changes required for this PR provided a good opportunity to implement an `encoding` option that allows developers to specify `base64`, as opposed to the default of `hex`. 

Hex encoding increases the encrypted payload file size by over 100% compared to the the input data -- every hex-encoded byte takes up 2 bytes in the resulting stringified JSON. Base64, however, is approximately only a 33% size increase. 

Hex encoding is still used by default, so that this PR does not introduce breaking changes. However, we should strongly consider making base64 the default for the next major release. 

#### (minor) putFile error handling for invalid charsets

If `putFile` is called with a `string` value, and with a `contentType` header that specifies a charset other than `charset=utf-8`, an error is thrown immediately. Previously, this would have been ignored, and the `Buffer.from(<string value>)` would default to `utf8`, and possibly cause garbled output later on when decoding. Or, if encryption was disabled, the string value would have been sent directly to the `fetch` method, which also encodes strings as utf8: https://fetch.spec.whatwg.org/
Examples:
* Throws: `putFile(<string value>, { contentType: 'text/plain; charset=x-ebcdic-france-euro' })`
* Okay: `putFile(<string value>, { contentType: 'text/plain' })`
* Okay: `putFile(<string value>, { contentType: 'text/plain; charset=utf-8' })`

#### (minor) putFile retries
The `putFile` upload request previously had a try/catch that would _always_ attempt to re-upload on error. Now that we have proper Gaia error parsing and classes, this behavior has been modified to only attempt re-upload when the response error is of a type that is possible to recover from. 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated


yknl (approved):
Looks good! The file size reduction from changing the encoding is a nice bonus. Good to see that backwards compatibility was also handled.

zone117x:
@timstackblock requested you for QA. Typical smoke tests, probably extra attention to the put/get file operations

reedrosenbluth (approved):

reedrosenbluth:
Was able to verify both that the `PayloadTooLargerError` gets thrown when the payload exceeds the set max size, and that the encoding options settings work (tested with a 1mb file, the POST body contents were ~34% larger than the file rather than ~100%)

PR #731, opened 2019-10-01
Change dist file CDN sample generator to ignore prerelease versions

zone117x:
## Description

Fixes https://github.com/blockstack/blockstack.js/issues/710

CDN usage snippets are no longer written for prerelease versions (e.g. `19.4.0-beta.1`). 

The usage snippets are now also written into a template string in the `README.md` rather than a separate file. 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

This only effects the npm package release process. Needs tested with performing both a release and prerelease. 

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,16 @@ Note: If you're looking for the Blockstack CLI repo it was merged with [Blocksta
 $ npm install blockstack
 ```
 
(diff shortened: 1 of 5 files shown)


yknl (approved):
This works great! Thanks 👍

PR #734, opened 2019-10-03
[DOC] Clarify file count for listFile

friedger:
## Description
This PR extends the result description of `listFiles` for the case the the listing was ended through the callback (returned false)

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

no
## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated

### DIFF
--- a/src/storage/index.ts
+++ b/src/storage/index.ts
@@ -777,7 +777,7 @@ async function listFilesLoop(
  * List the set of files in this application's Gaia storage bucket.
  * @param {function} callback - a callback to invoke on each named file that
  * returns `true` to continue the listing operation or `false` to end it
- * @return {Promise} that resolves to the number of files listed, or rejects with an error.
+ * @return {Promise} that resolves to the number of files listed (excluding the last if ended early), or rejects with an error.
  */
 export function listFiles(
   callback: (name: string) => boolean,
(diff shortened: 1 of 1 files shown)


src/storage/index.ts
```diff
@@ -777,7 +777,7 @@ async function listFilesLoop(
  * List the set of files in this application's Gaia storage bucket.
  * @param {function} callback - a callback to invoke on each named file that
  * returns `true` to continue the listing operation or `false` to end it
- * @return {Promise} that resolves to the number of files listed, or rejects with an error.
+ * @return {Promise} that resolves to the number of files listed (excluding the last if ended early), or rejects with an error.
```
moxiegirl: > that resolves to the number of files listed (excluding the last if ended early), or rejects with an error.

Resolves to the total number of listed files.  If the call is ended early, the last file is excluded or the entire call rejects with an error.

moxiegirl (changes requested):
This read a bit awkwardly. See suggested changs.

yknl:
@friedger can you take a look?

friedger:
@moxiegirl I have adopted your suggestion

friedger:
Is this good to merge?

moxiegirl (approved):

zone117x (approved):

Issue #123, opened 2016-08-18
cli ascii ui design


guylepage3:
Design new cli ascii ui

Items to design:
- blockstack ascii logo
- ascii install loader

muneeb-ali:
Oh that'd be awesome! I've thought about it a couple of times as well.

jcnelson:
@guylepage3 which color do you want?  I can make the ASCII logo show up in any of these colors:

https://en.wikipedia.org/wiki/File:Xterm_256color_chart.svg

guylepage3:
@jcnelson very cool. so the color options we have are all web safe only?

jcnelson:
The colors in the link are available on most UNIX command-line interfaces.  If you can give me the ASCII logo and tell me which color(s) to make it, I can write the code to print it out accordingly.

guylepage3:
Sounds good. When's the next release? I would like to push it out with that.

Also, are we able to execute a cool installation loader animation? Something simple for now. For example how NPM, Grunt or Gulp loaders look?

jcnelson:
Probably not a high priority, since we have Portal now...

PR #736, opened 2019-10-08
Adds expiry time to authToken header sent in connectToGaiaHub call

mascot6699:

## Description

With the current state of code it feels like the token returned by `makeV1GaiaAuthToken` has the possibility of reuse and thus hides the optional functionality of expiry of token which can be set as specified by Gaia.
If the `payload` has the optional `exp` params while creating the token using `makeV1GaiaAuthToken` then gaia's [validate token](https://github.com/blockstack/gaia/blob/73409ca31e0f203d352e9f4851adf3898b0f21da/hub/src/server/authentication.ts#L362) function can correctly identify if token is expired/reuse is possible.

By this change no current applications will be impacted but as a Blockstack developer who wish to create token with higher security bar(by specifying its expiry) can do so, thus preventing any misuse of token via replay attacks. This functionality is particularly useful for creating apps which requires privateKeyHolders approval for each gaia upload call.

For details refer to issue #123

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

n/a

Yes

## Testing information

This can be tested for making a gaia upload call with token which is already expired (i.e call `connectToGaiaHub` with negative `tokenValidityBufferTime`) and it should return 401 with response similar to 
```
{
    "message": "Expired authentication token: expire time of xxxxxxx (secs since epoch)",
    "error": "ValidationError"
}
```
Also testing that `connectToGaiaHub` call without `tokenValidityBufferTime` remains as it is.

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated


friedger:
See also https://github.com/blockstack/docs.blockstack/issues/151 (and maybe https://github.com/blockstack/blockstack.js/issues/607)

zone117x (approved):
Tested this and works as expected, thanks!

timstackblock (approved):

zone117x:
@mascot6699 can you revert the changes made after my PR review? We cannot merge this with the recent changes that are unrelated to the goal of the PR.

agraebe:

PR #740, opened 2019-11-07
feat: Add organisation-wide Prettier config

kyranjamie:
## Description

Per [this issue](https://github.com/blockstack/blockstack.js/issues/739) and discussion on our internal slack, this PR adds our organisation-wide Prettier config.

Note, this PR does not:
- Reformat all the code. For reasons relating to git history, maybe it's more desirable to adhere to the new code style when each file is edited?
- Enforce Prettier conformation with CI tool


reedrosenbluth (approved):
LGTM! I think I agree with the strategy of not reformatting the entire codebase, and letting that happen as a gradual process in order to preserve git history.

yknl:
I don't see any usage instructions. Is that not in scope for this PR?

kyranjamie:
@yknl Hey Ken, I left this PR to quite a bare minimum. I'd say usage is mostly down to a given developers set up/IDE config.

I could add a package json script to prettier everything, but then we should probably format the whole codebase now?

yknl:
Is there a way to add an ignore file for prettier and gradually remove files from the list as we go?

kyranjamie:
Indeed there is a [prettier ignore config we can use](https://prettier.io/docs/en/ignore.html#ignoring-files)

We can add the result of `find src -name "*.ts"` to it – although would need some motivation in place to begin removing the files as we go. If we do this, we can format a handful of files now, and set up CI for it as well?

yknl:
That sounds like a good approach to me.

kyranjamie:
Done, I've had to disable the eslint semi colon rule, which is now enforced by Prettier.

yknl:
Thanks @kyranjamie! I see that `npm run lint:formatting` checks the files for conformance, but doesn't actually tell you what needs to be changed. I tried running the prettier command `node_modules/.bin/prettier src/wallet.ts --write` on a file but it seems to have added semicolons to every line. I'm guessing that it's not picking up the config. What is the correct command? Should it be added to `scripts`?

kyranjamie:
It doesn't say any files because of the `.prettierignore`, remove a few lines and run again and it'll show them.

yknl:
> It doesn't say any files because of the .prettierignore, remove a few lines and run again and it'll show them.

I meant that it doesn't tell you which lines in the files are in violation of what rules. But I suppose that is not a necessity as you can just run the prettifier and look at the diff. 

Also, looks like there isn't a problem with the config, I just thought we were not requiring semi colons. And since we've voted to require semi colons, I think this PR is good to go.

yknl (approved):

PR #743, opened 2019-11-15
Include ETag in Request Header

reedrosenbluth:
## Description

This is the blockstack.js portion of the optimistic concurrency control changes I've been working on for Gaia, referenced in this issue: https://github.com/blockstack/gaia/issues/267

With this change, blockstack.js attempts to parse an `etag` from the response headers of a `getFile` request. The `etag` is a value that represents the version of the file in the datastore. Often it is an md5 hash of the response body. If the `etag` is present, it is stored in memory. When an update to that file is attempted with `putFile`, the saved `etag` value is sent along with the request in the `if-match` header. Gaia will only proceed with the update if the tag matches what is in the data store (meaning, the file hasn't been changed since it was last read by the client requesting the update).

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

Nope. Older versions of Gaia will simply ignore the `if-match` header.

Yes

## Testing information
I've tested this change with a [simple blockstack.js app](http://bin.reedrosenbluth.com/) I built, and a custom deployed Gaia Hub on DigitalOcean.

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


src/storage/index.ts
```diff
@@ -260,6 +267,10 @@ async function getFileContents(path: string, app: string, username: string | und
     throw await getBlockstackErrorFromResponse(response, `getFile ${path} failed.`)
   }
   const contentType = response.headers.get('Content-Type')
+  const etag = JSON.parse(response.headers.get('ETag'))
```
kyranjamie: Could the etag ever be corrupt/unparsable and cause `JSON.parse` to throw an error?

reedrosenbluth: Was using `JSON.parse()` to get rid of redundant double quotes (for some reason the storage drivers return etags with nested quotes). Replacing this with `.replace(/^"|"$/g, '')` as per Matt's suggestion.

src/storage/hub.ts
```diff
@@ -40,16 +41,24 @@ export async function uploadToGaiaHub(
...
+): Promise<PutFileResponse> {
   Logger.debug(`uploadToGaiaHub: uploading ${filename} to ${hubConfig.server}`)
+
+  const headers: { [key: string]: string; } = {
+    'Content-Type': contentType,
+    Authorization: `bearer ${hubConfig.token}`
+  }
+
+  if (etag) {
+    headers['If-Match'] = etag
```
zone117x, line 59: It seems like this is one of the few headers that we will have available in the future for _any_ typeof metadata transmission. 
Should we send structured payload for versioning/future-proofing reasons?
e.g.
```js
headers['If-Match'] = JSON.stringify({
  'v': '1.0',
  'tag': etag,
})
```

zone117x: Identified some potential issues where something like may be needed on initial release https://github.com/blockstack/gaia/pull/276/files#r358970449

src/storage/index.ts
```diff
@@ -27,6 +27,8 @@ import { NAME_LOOKUP_PATH } from '../auth/authConstants'
 import { getGlobalObject, getBlockstackErrorFromResponse, megabytesToBytes } from '../utils'
 import { fetchPrivate } from '../fetchUtil'
 
+const etags: { [key: string]: string; } = {}
```
zone117x, line 30: I think having this as a global variable is okay for now, but we should create an issue for moving this into `UserSession.store` (an instead of the `SessionDataStore` class). 

This would allow the data store to persist this map across sessions. As well fix duplicate etag-file-key bugs with certain applications that may be working with multiple `UserSession` instances. 

It would also prevent memory leaks for apps like long-running Node.js services which may create and dispose of many different UserSession instances over time. 

We will have to complete https://github.com/blockstack/blockstack.js/issues/744 before being able to implement that though.

src/storage/index.ts
```diff
@@ -784,12 +791,15 @@ export async function putFile(
     const signatureContent = JSON.stringify(signatureObject)
 
     uploadFn = async (hubConfig: GaiaHubConfig) => {
-      const fileUrls = await Promise.all([
-        uploadToGaiaHub(path, contentData, hubConfig, contentType),
+      const writeResponse = (await Promise.all([
+        uploadToGaiaHub(path, contentData, hubConfig, contentType, etags[path]),
```
zone117x: The concurrency situation with `signed & unencrypted` uploads here is pretty bad. We should create an issue for attempting to reduce this to a single file that contains both the signature and data (similar to the `signed & encrypted` payload).

tests/unitTests/src/unitTestsStorage.ts
```diff
@@ -737,6 +737,58 @@ export function runStorageTests() {
       })
   })
 
+  test('putFile passes etag to upload function', async (t) => {
```
zone117x, line 740: The `spy` test pattern here is cool. Nice improvement over the way the similar tests work.

I think we need at least one more test, like this one, that triggers the new PreconditionFailed error path.

zone117x (approved):
This looks great! I left a couple comments around where we should open new issues for future improvements in later PRs. 

I think we should also open an issue for something like `stricter etag/concurrency enforcement in putFile`. For example, if the file key is missing from the `etag` map, then the POST request should use the `If-None-Match: *` header. With this current PR, apps can accidentally clobber data if they are built in a way where `putFile` is used without a previous `getFile` within the same session.

src/storage/index.ts
reedrosenbluth: done

yknl (approved):
LGTM

timstackblock:
Looks good in staging, we can move to prod and test after deploy.

PR #767, opened 2020-01-23
Remove deprecated static functions and require UserSession argument in storage methods

### LINKED ISSUE #744: Remove deprecated static functions replaced by UserSession

reedrosenbluth:
Fixes #744

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

Will break apps that are using these static functions instead of UserSession

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated


src/auth/userSession.ts
```diff
@@ -20,6 +19,7 @@ import {
 import { Logger } from '../logger'
 import { GaiaHubConfig, connectToGaiaHub } from '../storage/hub'
 import { BLOCKSTACK_DEFAULT_GAIA_HUB_URL, AuthScope } from './authConstants'
+import { redirectToSignInWithAuthRequest, isSignInPending, handlePendingSignIn, signUserOut, getAuthResponseToken } from './authApp'
```
yknl: Should probably break this into multiple lines

src/storage/index.ts
```diff
@@ -169,10 +164,6 @@ export async function encryptContent(
 }
 
 /**
- * 
- * @deprecated
```
zone117x, line 173: Various storage functions with an optional `caller?: UserSession` param are still being globally exported. 

For example, the `putFile` function here has an optional `caller?: UserSession` param, which was implemented for backwards compatibility when `UserSession` was introduced. 

When the param is not specified, these functions create a temporary instance with code like:
```js
const thing = (caller || new UserSession()).thing;
```

Usage of the functions will cause problems later on, like when we move the file etag map from a global variable to a property accessed from `UserSession`. 

A quick fix could be modifying the `export * from './storage'` statement in `public.ts` so that it only exports the handful of intended items. 


I think an even better approach is refactor to using a storage class w/ a composition pattern. This would also allow us to cleanup some of the hacky, inconvenient unit tests that are forced to `proxyquire` a bunch of these module exports. Example
```ts
export class Storage {
  readonly session: UserSession
  constructor(userSession: UserSession) {
    this.session = userSession
  }
  async putFile(path, content, options?: PutFileOptions) {
    const thing = this.session.getThing();
    ...
  }
}
```

Thoughts?

zone117x (changes requested):

src/storage/index.ts
yknl: Refactoring and removing the optional `UserSession` arguments would be the best approach. This will prevent the use of multiple `UserSessions`, but I don't think that's a real use case right now.

zone117x:
Looks great! Just needs changelog update with the breaking changes listed

zone117x (approved):
LGTM!

yknl (approved):

PR #777, opened 2020-03-25
Update release-checklist.md

agraebe:
## Description

Adding todo to update community through email/discord/slack. This will allow broader community awareness for updates. 

Jenny and Mitchell are aware of and approved this addition. They also suggested to provide more lead time for high impact release and engage with them earlier on to craft a more comprehensive community update.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] Other: Release Checklist

No

No

### DIFF
--- a/release-checklist.md
+++ b/release-checklist.md
@@ -21,6 +21,7 @@
 - [ ] `git branch -D master-test`
 - [ ] `git push origin :master-test`
 - [ ] Announce availability in #engineering and on the forum here: https://forum.blockstack.org/c/releases (Include a link to the changelog)
+- [ ] Announce availability to the community via Jenny Mith <jenny@blockstack.com> and Mitchell Cuevas <mitchell@blockstack.com>. Provide a succinct list of changes, the like to the changelog and latest release as well as a desired date for the community update.
 
 ## Production Release
 - [ ] `git flow release start <VERSION_NUMBER> master` (`<VERSION_NUMBER>` prefixed with `v`)
@@ -47,3 +48,4 @@
 - [ ] `git branch -D master-test`
 - [ ] `git push origin :master-test`
 - [ ] Announce availability in #engineering and on the forum here: https://forum.blockstack.org/c/releases (Include a link to the changelog)
+- [ ] Announce availability to the community via Jenny Mith <jenny@blockstack.com> and Mitchell Cuevas <mitchell@blockstack.com>. Provide a succinct list of changes, the like to the changelog and latest release as well as a desired date for the community update. Consider doing an AMA with the community.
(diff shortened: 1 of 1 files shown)


release-checklist.md
```diff
@@ -21,6 +21,7 @@
 - [ ] `git branch -D master-test`
 - [ ] `git push origin :master-test`
 - [ ] Announce availability in #engineering and on the forum here: https://forum.blockstack.org/c/releases (Include a link to the changelog)
+- [ ] Announce availability to the community via Jenny Mith <jenny@blockstack.com> and Mitchell Cuevas <mitchell@blockstack.com>. Provide a succinct list of changes, the like to the changelog and latest release as well as a desired date for the community update.
```
yknl: Should it be "the link to the changelog" instead?

release-checklist.md
```diff
@@ -47,3 +48,4 @@
 - [ ] `git branch -D master-test`
 - [ ] `git push origin :master-test`
 - [ ] Announce availability in #engineering and on the forum here: https://forum.blockstack.org/c/releases (Include a link to the changelog)
+- [ ] Announce availability to the community via Jenny Mith <jenny@blockstack.com> and Mitchell Cuevas <mitchell@blockstack.com>. Provide a succinct list of changes, the like to the changelog and latest release as well as a desired date for the community update. Consider doing an AMA with the community.
```
yknl: Is the AMA meant to happen for every release?

release-checklist.md
agraebe: I think a link is not sufficient enough. They want two things mainly: 1) a date when this should go out to the community and 2) a message that should go out to the community. Just providing a link to the changelog won't be enough. We should clearly community in brief why and what was changed as well as what the impact is. It's something that will be really hard for Jenny or Mitchell to get from the pure technical changelog is my argument.

release-checklist.md
agraebe: No, optional. I think this would require a judgement call on the releaser, ie "is this important enough to organize an AMA for"? Should I point that out as well?

release-checklist.md
yknl: I was only referring to the spelling error: "like" in place of "link". Otherwise looks good to me.

release-checklist.md
yknl: I think it should be clear when that should happen. i.e. Consider doing an AMA on major/breaking releases.

release-checklist.md
agraebe: Hah, of course, that needs to be fixed :)

release-checklist.md
agraebe: Makes sense, adding that

yknl (approved):
LGTM

Issue #760, opened 2020-01-15
Move etag map to `UserSession.store`


reedrosenbluth:
> I think having this as a global variable is okay for now, but we should create an issue for moving this into `UserSession.store` (an instead of the `SessionDataStore` class). 

> This would allow the data store to persist this map across sessions. As well fix duplicate etag-file-key bugs with certain applications that may be working with multiple `UserSession` instances. 

> It would also prevent memory leaks for apps like long-running Node.js services which may create and dispose of many different UserSession instances over time. 

> We will have to complete https://github.com/blockstack/blockstack.js/issues/744 before being able to implement that though.

_Originally posted by @zone117x in https://github.com/blockstack/blockstack.js/pull/743_

PR #793, opened 2020-06-16
fix: allow force option in putFile

### LINKED ISSUE #788: "Force" option for putFile
There should be a `force` option for `putFile` that bypasses Gaia's etag concurrency control.

aulneau:
## Description

I've been working on a small side project and have been running into a lot of issues with syncing the application data to gaia due to mismatching e-tags A hack I had to use to fix it was doing a `getFile`, then `putFile`. This was not ideal because I didn't want to do two fetches for every state change. This PR implements a simple `force` boolean that ignores the e-tags completely. This will resolve https://github.com/blockstack/blockstack.js/issues/788.

Nope

Yes, to show how to pass the option

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/storage/hub.ts
+++ b/src/storage/hub.ts
@@ -30,21 +30,22 @@ interface UploadResponse {
 }
 
 /**
- * 
- * @param filename 
- * @param contents 
- * @param hubConfig 
- * @param contentType 
- * 
+ *
+ * @param filename
+ * @param contents
+ * @param hubConfig
+ * @param contentType
+ *
  * @ignore
  */
 export async function uploadToGaiaHub(
-  filename: string, 
+  filename: string,
   contents: Blob | Buffer | ArrayBufferView | string,
   hubConfig: GaiaHubConfig,
   contentType: string = 'application/octet-stream',
   newFile: boolean = true,
-  etag?: string
+  etag?: string,
+  force?: boolean
 ): Promise<UploadResponse> {
   Logger.debug(`uploadToGaiaHub: uploading ${filename} to ${hubConfig.server}`)
 
@@ -53,10 +54,12 @@ export async function uploadToGaiaHub(
     Authorization: `bearer ${hubConfig.token}`
   }
 
-  if (newFile) {
-    headers['If-None-Match'] = '*'
-  } else if (etag) {
-    headers['If-Match'] = etag
+  if (!force) {
+    if (newFile) {
+      headers['If-None-Match'] = '*'
+    } else if (etag) {
+      headers['If-Match'] = etag
+    }
   }
 
   const response = await fetchPrivate(
@@ -94,10 +97,10 @@ export async function deleteFromGaiaHub(
 }
 
 /**
- * 
- * @param filename 
- * @param hubConfig 
- * 
+ *
+ * @param filename
+ * @param hubConfig
+ *
  * @ignore
  */
 export function getFullReadUrl(filename: string,
@@ -106,10 +109,10 @@ export function getFullReadUrl(filename: string,
 }
 
 /**
- * 
- * @param challengeText 
- * @param signerKeyHex 
- * 
+ *
+ * @param challengeText
+ * @param signerKeyHex
+ *
  * @ignore
  */
 function makeLegacyAuthToken(challengeText: string, signerKeyHex: string): string {
@@ -129,11 +132,11 @@ function makeLegacyAuthToken(challengeText: string, signerKeyHex: string): strin
     const signatureBuffer = signer.sign(digest)
     const signatureWithHash = script.signature.encode(
       signatureBuffer, Transaction.SIGHASH_NONE)
-    
+
     // We only want the DER encoding so remove the sighash version byte at the end.
     // See: https://github.com/bitcoinjs/bitcoinjs-lib/issues/1241#issuecomment-428062912
     const signature = signatureWithHash.toString('hex').slice(0, -2)
-    
+
     const publickey = getPublicKeyFromPrivate(signerKeyHex)
     const token = Buffer.from(JSON.stringify(
       { publickey, signature }
@@ -145,12 +148,12 @@ function makeLegacyAuthToken(challengeText: string, signerKeyHex: string): strin
 }
 
 /**
- * 
- * @param hubInfo 
- * @param signerKeyHex 
- * @param hubUrl 
- * @param associationToken 
- * 
+ *
+ * @param hubInfo
+ * @param signerKeyHex
+ * @param hubUrl
+ * @param associationToken
+ *
  * @ignore
  */
 function makeV1GaiaAuthToken(hubInfo: any,
@@ -179,7 +182,7 @@ function makeV1GaiaAuthToken(hubInfo: any,
 }
 
 /**
- * 
+ *
  * @ignore
  */
 export async function connectToGaiaHub(
@@ -205,10 +208,10 @@ export async function connectToGaiaHub(
 }
 
 /**
- * 
- * @param gaiaHubUrl 
- * @param appPrivateKey 
- * 
+ *
+ * @param gaiaHubUrl
+ * @param appPrivateKey
+ *
  * @ignore
  */
 export async function getBucketUrl(gaiaHubUrl: string, appPrivateKey: string): Promise<string> {
--- a/src/storage/hub.ts
+++ b/src/storage/hub.ts
@@ -45,7 +45,7 @@ export async function uploadToGaiaHub(
   contentType: string = 'application/octet-stream',
   newFile: boolean = true,
   etag?: string,
-  force?: boolean
+  dangerouslyIgnoreEtag?: boolean
 ): Promise<UploadResponse> {
   Logger.debug(`uploadToGaiaHub: uploading ${filename} to ${hubConfig.server}`)
 
@@ -54,7 +54,7 @@ export async function uploadToGaiaHub(
     Authorization: `bearer ${hubConfig.token}`
   }
 
-  if (!force) {
+  if (!dangerouslyIgnoreEtag) {
     if (newFile) {
       headers['If-None-Match'] = '*'
     } else if (etag) {
--- a/src/storage/index.ts
+++ b/src/storage/index.ts
@@ -784,7 +784,8 @@ export async function putFile(
 
     uploadFn = async (hubConfig: GaiaHubConfig) => {
       const writeResponse = (await Promise.all([
-        uploadToGaiaHub(path, contentData, hubConfig, contentType, newFile, etag, opt.dangerouslyIgnoreEtag),
+        uploadToGaiaHub(path, contentData, hubConfig, contentType, newFile, etag,
+                        opt.dangerouslyIgnoreEtag),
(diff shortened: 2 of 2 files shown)


hstove:
Just as a :+1:, I'm overriding this in the authenticator, too. I have my own mechanisms for avoiding conflicts, because having to `getFile` everytime is just much less efficient.

yknl:
This makes me think we should re-evaluate how the etag system was implemented. But for the time being a force option sounds like a good workaround.

zone117x:
@hstove @aulneau what conditions or usages are causing etag mismatches for you? If you're running into this situation, it _typically_ means your app is prone to race conditions that result in data loss.

hstove:
Well, I don't use `getFile` - I probably could, but since I'm not working with a `UserSession` in the authenticator, I'm doing lower-level fetches and stuff. One example is when I have to update the `profile.json`. I'm still fetching it first to make sure I'm using the latest, and then I use `uploadToGaiaHub`, which has this mandatory etag thing in it. So, the authenticator has it's own little 10-line `uploadToGaiaHub` that doesn't do the etag stuff.

FWIW, we shouldn't worry too much about the authenticator's needs for DX. However, we should consider that folks might be doing stuff with Gaia without using `getFile` - I can think of a few ways I might do that as an app developer. In those instances, I'd still be checking for race conditions, but the etag stuff wouldn't work out of the box.

aulneau:
@zone117x i don't think it's due to race conditions. I've reliably been able to reproduce it on any attempt to put before a get. I think there is a failure in the underlying mechanism to get the latest etag, or something like that.

hstove:
Also, if you do run into this as a developer, it's really really hard to diagnose. And it's not that hard to run into. All you see is a failed fetch, with an exception thrown for that. We should definitely have custom error handling, preferably with a link to some documentation page.

hstove:
>  I've reliably been able to reproduce it on any attempt to put before a get.

That's exactly how it works, on purpose. It only can fetch the existing etag if you run `getFile` first.

aulneau:
> Also, if you do run into this as a developer, it's really really hard to diagnose. And it's not that hard to run into. All you see is a failed fetch, with an exception thrown for that. We should definitely have custom error handling, preferably with a link to some documentation page.

Yeah, the error from the server is only something like "failed"

zone117x:
The etag is also set during a putFile -- successive putFile calls should work, as long as they aren't concurrent to the same file.

When there is a version conflict, Gaia should be returning a [`412 Precondition Failed`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412) response (should be visible in the network inspector), and the putFile call _should_ throw a custom `PreconditionFailedError` instance: https://github.com/blockstack/blockstack.js/blob/52420a4df53b2690cfbc585ee7819af5ded9c2e7/src/utils.ts#L396

(adding a doc link to the error msg is a great idea though)


But it sounds like essentially none of the above holds true right now? If so, there could be a regression at some layer (azure CDN caching, azure storage API / gaia driver, blockstack.js). 

Can this error be reproduced? I think we'd want to make sure any underlying bugs are resolved rather than having developers work around the issue with this flag, which defeats the whole purpose of concurrency control.

aulneau:
@zone117x I have reproduced it here: 
https://putfile-test.aulneau.vercel.app/ 
https://github.com/aulneau/blockstack-putfile-example

log in, hit do putfile, refresh, open console, hit do putfile again and it will error. The first write is fine, but any additional puts to an existing file will result in a 500 from the server, and an error message that doesn't communicate what is happening.

zone117x:
Thanks @aulneau! There is definitely a regression with that 500 response. 

Also, once that regression is fixed, there is another issue with refreshing the page. The etag map is currently only in memory but it should be persisted. We have an issue for this that I've totally forgot about https://github.com/blockstack/blockstack.js/issues/760

hstove:
> The etag map is currently only in memory but it should be persisted.

OK that makes sense. I was a bit confused about this when looking through the code.

reedrosenbluth:
@aulneau I believe the behavior you are describing is working as I'd expect it, minus the `500` error, which I'd expect to be a `412` (https://github.com/blockstack/gaia/blob/master/hub/src/server/http.ts#L94).

I wouldn't expect this second `put` to work since you refreshed the page since the last `get`, which currently wipes the etag cache. Ideally the etag cache should be persisted as Matt mentioned (#760).

aulneau:
@reedrosenbluth So the example in the app I am building is this:

It's a next.js application which is fetching the encrypted data for your state if you're logged in on the server, sends that to the client which then decrypts it (technically right now it's decrypting on the server too, but it will ultimately do that client side for privacy). 

The state is up-to-date at that moment, and with any state change I have a component that puts the new state to the specific file changed. I don't understand the best way to get the etag without doing redundant fetching for all of them. I suppose on the specific page, I can fetch on load, but again it might do that on the sever at which point the client would never have an etag at first.

reedrosenbluth:
So the problem is that you are fetching the data on the server, and then initiating the `put` from the client, which doesn't have access to the servers knowledge of the etag? I hadn't really considered this use case before to be honest... Do you have two separate UserSessions?

aulneau:
> Do you have two separate UserSessions?

I have a wrapper for getting files from any gaia hub outside of a usersession (config is saved in a cookie so the server has access to it), and then the client side user session takes over.

reedrosenbluth:
Ah I see. I think to make this use case work we'd have to either:

1. have a way to manually update the etag map so that the server and client can send etags back and forth
2. implement `force put` as suggested

I agree with @zone117x's assessment above that `force put` more or less defeats the purpose of concurrency control, so I'm more inclined to go with option 1.

How do you guys think we should approach supporting this use case? Are there other options I've missed?

aulneau:
I'd say an additional layer of complexity is for a file that is being edited by more than one person. Let's say you and I are editing the same document, and you added the last change, and as such your instance of the app will be the one to upload the latest version to the shared gaia hub. I then edit it, so my client will push the latest version of that document. The changes are synced via webrtc and updated client side when multiple users are editing them. How would/should e-tags work in that situation?

yknl:
> I'd say an additional layer of complexity is for a file that is being edited by more than one person. Let's say you and I are editing the same document, and you added the last change, and as such your instance of the app will be the one to upload the latest version to the shared gaia hub. I then edit it, so my client will push the latest version of that document. The changes are synced via webrtc and updated client side when multiple users are editing them. How would/should e-tags work in that situation?

I think the etag system is actually great to have in the scenario you described. You wouldn't want your push to overwrite anything that the other person changed. There could easily be times where due to network delays, you would have an outdated version of the file. The etag system will give you an error in that case and prevent you from overwriting the file.

In terms of how etags would work in a system like you described, I would assume whenever the file is synced via webRTC, the locally cached etag could be updated as well. I don't think we expose any APIs to do that right now though.

hstove:
That's the thing - the current solution is good for more novice devs, as we force everyone to getFile before they putFile. But, it doesn't account for the many scenarios where a dev is actually doing conflict detection and handling it properly, but still will run into this scenario. Simply having the etag doesn't mean you've done conflict resolution. In the WebRTC example, you might now have the new etag, but it doesn't mean you've actually handled conflicts from the other end.

reedrosenbluth:
@hstove This is a great point. It's making me think that the etag feature should be turned on by default, but that there should be a way to opt out if you are a more experienced dev who wants more control over how to deal with concurrency-control

zone117x:
> I'd say an additional layer of complexity is for a file that is being edited by more than one person. Let's say you and I are editing the same document, and you added the last change, and as such your instance of the app will be the one to upload the latest version to the shared gaia hub. I then edit it, so my client will push the latest version of that document. The changes are synced via webrtc and updated client side when multiple users are editing them. How would/should e-tags work in that situation?

In this situation, the concurrency control system in place would surface at least a couple problems with the how the above logic _could_ have a buggy implementation.

Multiple sessions (regardless if from the same user or different users) forcefully writing to the same file at the same time is undefined behavior in gaia. This is because cloud storage providers do not have consistent behavior for concurrent competing writes. Consider `user A` initiates a putFile, then `user B` also initiates a putFile _before_ the write from `user A` completes. What is the result when `user B` finishes the write before `user A`, or the inverse? Some cloud storage providers are first-writer-wins while others are last-writer-wins. Either way, the behavior is not intuitive. 

Conditional writes (etags) surface this condition and forces it to be handled. This "lost update problem" is what conditional headers like [`If-Match`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match) are designed to help solve. 

I imagine a system using webRTC to synchronize changes between clients while _also_ correctly synchronizing the changes to Gaia (or any other cloud-storage-like API) would likely want to take advantage of etags and conditional http headers. 

There are certainly use cases that require lossy writes or indiscriminate overwrites to Gaia -- and for those cases it makes sense to provide this `force` flag. But I'm not convinced the webRTC example requires this.

zone117x:
@hstove 
> One example is when I have to update the profile.json. I'm still fetching it first to make sure I'm using the latest, and then I use uploadToGaiaHub, which has this mandatory etag thing in it. So, the authenticator has it's own little 10-line uploadToGaiaHub that doesn't do the etag stuff.

This is close to preventing race conditions / lost updates, but not 100% of the time. For example, a user with (really) slow connection that logs into a couple new multiplayer apps around the same time could potentially result in a missed update to the profile.json `apps` list. That situation is rare enough that I wouldn't worry about it -- but if you wanted to efficiently detect and handle it, you'd definitely want to use conditional writes w/ etags (rather than putfile, followed by getfile to validate that the change wasn't trampled).

hstove:
> This is close to preventing race conditions / lost updates, but not 100% of the time.

Yep, I follow this, and I'd agree that using the `If-Match` would have an added benefit here. Since I'm already re-fetching the file before saving, I already have the ability to get the etag, I just don't actually know how to, and it's not automatic since I'm not using `getFile`.

hstove:
It sounds like we're mostly on the same page that this thing is here for a good reason, and I'd imagine we would benefit from:

- A proper doc for advanced users that outlines what it is and how to use it
- A better error message that links to this doc\
- _Some_ way of overriding. It could be like Reacts `dangerouslySetInnerHTML` - something explicit where the developer waives their rights away and voids any liability :p

reedrosenbluth:
@aulneau I've fixed the 500 error with this PR: https://github.com/blockstack/gaia/pull/305

It will still throw an error, only now it's a 412 PreconditionFailedError with the appropriate message.

yknl:
Sounds like this PR can be accepted if we change `force` to something like `forceDangerously? @zone117x @reedrosenbluth

aulneau:
@yknl I wonder if something that could speak to concurrency would be better, or use words like `unsafe` or `dangerously`. I also do wonder exact _how dangerous_ this is?

reedrosenbluth:
I don't have too strong an opinion about the naming here... @aulneau perhaps we could do  `putFileOverride`

aulneau:
Perhaps something specific: `ignoreEtags`?

reedrosenbluth (approved):

yknl (approved):
LGTM

zone117x (approved):

PR #794, opened 2020-06-17
feat: pass gaia hub to both get and putFile fns

aulneau:
## Description

If someone wants to build an app that allows for collaborative editing of shared files, we need to allow the application to provide any gaia hub config that could contain the scoped auth token for that file. This PR implements the minimum amount of changes to allow for this kind of api:

```tsx
const upload = await userSession.putFile(filename, contents {gaiaHubConfig: gaiaHubConfigWithScopedAuthToken});
const read = await userSession.getFile(filename, {gaiaHubConfig: gaiaHubConfigWithScopedAuthToken});
```

This would also allow the application to generate a new auth tokens that can change the way gaia writes their files (for example: automatic versioning with `putFileArchival`):

```tsx
type ScopeTypes =
  | 'putFile'
  | 'putFilePrefix'
  | 'deleteFile'
  | 'deleteFilePrefix'
  | 'putFileArchival'
  | 'putFileArchivalPrefix';
```
related to https://github.com/blockstack/blockstack.js/pull/545

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

no

Yes, to explain the additional option and probably show them how to generate the tokens

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/storage/hub.ts
+++ b/src/storage/hub.ts
@@ -30,21 +30,22 @@ interface UploadResponse {
 }
 
 /**
- * 
- * @param filename 
- * @param contents 
- * @param hubConfig 
- * @param contentType 
- * 
+ *
+ * @param filename
+ * @param contents
+ * @param hubConfig
+ * @param contentType
+ *
  * @ignore
  */
 export async function uploadToGaiaHub(
-  filename: string, 
+  filename: string,
   contents: Blob | Buffer | ArrayBufferView | string,
   hubConfig: GaiaHubConfig,
   contentType: string = 'application/octet-stream',
   newFile: boolean = true,
-  etag?: string
+  etag?: string,
+  force?: boolean
 ): Promise<UploadResponse> {
   Logger.debug(`uploadToGaiaHub: uploading ${filename} to ${hubConfig.server}`)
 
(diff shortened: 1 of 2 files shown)


hstove:
Dope! Now add the code to add scopes to the auth token :)

aulneau:
@zone117x @reedrosenbluth  @yknl @hstove maybe we can set up some time to talk through this kind of thing / use case in a sync meeting?

yknl:
Let's discuss this, I would prefer an approach that abstracts away the Gaia hub config for developers. For example, when you create a scoped auth token for another user, it could be stored on your hub and encrypted using the public key of the authorized user. When the authorized user is making a putFile request, they would just provide a username and the auth token/hub config lookup would be automatic.

aulneau:
> Let's discuss this, I would prefer an approach that abstracts away the Gaia hub config for developers. For example, when you create a scoped auth token for another user, it could be stored on your hub and encrypted using the public key of the authorized user. When the authorized user is making a putFile request, they would just provide a username and the auth token/hub config lookup would be automatic.

Sure thing! I am just trying to make the thing I am building work, and this seems to be the way that I've found so far. I wonder how what you suggest would work when each file shared has a different auth token? (scoped to that file/path)

yknl:
> I wonder how what you suggest would work when each file shared has a different auth token? (scoped to that file/path)

I think that can be handled by having a list of auth tokens stored on your own Gaia hub, keyed by the hash of the file name. The authorized user already supplies a file name to putFile.

agraebe:

PR #797, opened 2020-06-22
fix: missing optional argument from decryptMnemonic function

kyranjamie:
Adds `?` to decrypt mnemonic function.

See 3aaf773791fc0734d1b49c15b66b08e15e137928 for non-formatting changes

### DIFF
--- a/src/encryption/wallet.ts
+++ b/src/encryption/wallet.ts
@@ -1,10 +1,10 @@
-import { validateMnemonic, mnemonicToEntropy, entropyToMnemonic } from 'bip39'
-import { randomBytes, GetRandomBytes } from './cryptoRandom'
-import { createSha2Hash } from './sha2Hash'
-import { createHmacSha256 } from './hmacSha256'
-import { createCipher } from './aesCipher'
-import { createPbkdf2 } from './pbkdf2'
-import { TriplesecDecryptSignature } from './cryptoUtils'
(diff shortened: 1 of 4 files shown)


zone117x:
I think the legacy "triplesec" handling needs adjusted for this, per unit test failure `Should decrypt legacy encrypted phrase, instead errored: Error: Unable to decrypt. Wrong password.`

src/encryption/wallet.ts
```diff
@@ -149,17 +155,14 @@ function decryptLegacy(dataBuffer: Buffer,
...
   const dataBuffer = Buffer.isBuffer(data) ? data : Buffer.from(data, 'hex')
-  try {
-    return await decryptMnemonicBuffer(dataBuffer, password)
-  } catch (err) {
-    // If it was a password error, don't even bother with legacy
-    if (err instanceof PasswordError) {
-      throw err
-    }
+
+  if (triplesecDecrypt) {
```
zone117x: The `triplesecDecrypt` arg exists so that caller apps can be responsible for including the triplesec lib, and can then pass the triplesec decrypt function to blockstack.js. 

This usage looks like "if triplesecDecrypt exists then do legacy" -- which won't work

kyranjamie: `triplesecDecrypt` is only used for the legacy function though?

zone117x (approved):

PR #801, opened 2020-07-08
feat: persist etags in UserSession

### LINKED ISSUE #760: Move etag map to `UserSession.store`

reedrosenbluth:
## Description

This PR moves the etag map to the UserSession store so the etags persist across sessions.

Issue: #760

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

No

## Testing information

Tested with this sample app: https://blockstack-bin.reedr.vercel.app
Source: https://github.com/reedrosenbluth/bin/tree/persist

Steps:
1. Click `load` to load saved text (if there is any)
2. Modify text in text box
3. Click `save` to overwrite the text and load the new etag
4. Refresh the page and don't `load` the text
5. Enter new text in the text box and click `save`

These steps can be executed without error. This shows that the etag was properly persisted from the previous session (before the refresh).

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Changed
+- Persist etags in UserSession store so they don't always need to be fetched when the page is loaded.
+
 ## [21.1.0] - 2020-06-05
 ### Changed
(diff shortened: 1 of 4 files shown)


zone117x (approved):
LGTM, thanks @reedrosenbluth!

yknl (approved):
LGTM!

PR #824, opened 2020-10-02
build: monorepo fixes

aulneau:
This PR implements TSDX and some tsconfig changes. This change allows for better tree-shaking by producing esmodules in addition to common js and umd packages. It also adjusts things to be more in line with our other mono repos: yarn workspaces and eslint-config.


kyranjamie (approved):
👍 

Would be great to add a prettier --check action

yknl (approved):
Merging this into the main branch. Thanks again for the build improvements!

PR #826, opened 2020-10-08
feat(transactions): export cvToHex, hexToCV, parseReadOnlyResponse

friedger:
## Description

As a blockstack developer, I would like to use the blockstack-api-client library. The method `SmartContractsApi.callReadOnlyFunction` requires to hex encode clarity values, the result of that method needs to be parsed. In  this library (`stacks-transactions-js`) these methods already exists.

These methods, defined in `utils.ts` are now exported.

For details refer to issue https://github.com/blockstack/stacks-transactions-js/issues/126

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

yes, documentation for new methods needs to be published

## Testing information
No code features changed.

## Checklist
- [x] `npm run test` in transactions passes
- [ ] Changelog is updated - where is that?
- [x] Tag @yknl for review

### DIFF
--- a/packages/transactions/src/index.ts
+++ b/packages/transactions/src/index.ts
@@ -29,3 +29,5 @@ export * from './types';
 export * from './constants';
 export * from './contract-abi';
 export * from './signer';
+
+export { cvToHex, parseReadOnlyResponse, hexToCV } from './utils';
--- a/packages/transactions/src/utils.ts
+++ b/packages/transactions/src/utils.ts
@@ -160,12 +160,24 @@ export async function fetchPrivate(input: RequestInfo, init?: RequestInit): Prom
   const fetchResult = await fetch(input, fetchOpts);
   return fetchResult;
 }
-
+/**
+ * Converts a clarity value to a hex encoded string with `0x` prefix
+ * @param {ClarityValue} cv  - the clarity value to convert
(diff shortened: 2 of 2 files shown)


yknl (approved):
Thanks for copying over!

agraebe:
Where will these methods be documented? We currently don't have any documentation for them. In the txjs repo, we would use the README file. @yknl - will these methods be in the new docs for the monorepo?

yknl:
There are typedoc generated reference docs but it's better to have it in the form of a smart contract tutorial.

agraebe:
@friedger could you add some docs updates for this PR?

friedger:
I have provided the documentation in the code. @agraebe isn't it enough to regenerate the docs? I'm using these methods also in the tutorial

agraebe:
I think I wanted to make a broader point for documentation inside our docs as opposed to just inside typedoc. most developers will find it very buried if it is only mentioned in the typedoc. I think it's okay if we tackle the broader issue outside of this PR. 

The only relevant place I can think of (right now) would be https://docs.blockstack.org/stacks-blockchain/integrate-stacking#step-6-display-stacking-status

yknl (approved):

PR #836, opened 2020-10-20
feat: storage calls without user session

yknl:
This PR allows storage calls without having a signed in user session.

https://github.com/blockstack/blockstack.js/issues/802

For details refer to issue #123

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

No

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated

### DIFF
--- a/packages/storage/src/storage.ts
+++ b/packages/storage/src/storage.ts
@@ -827,6 +827,9 @@ function normalizeOptions<T>(
   } & T
 ) {
   const opts = Object.assign({}, options);
+  if (!userSession) {
+    return opts
+  }
   if (opts.username) {
     if (!opts.app) {
       if (!userSession.appConfig) {


markmhendrickson:
This change seems straightforward enough. @yknl do you want me to test this as an integration in my app specifically or are you confident it'll work there?

yknl:
@markmhx  Yes, we can wait until you've had a chance to test it with your app, just want to make sure it resolves the issue for you.

hstove (approved):
Would love to see a test here but :+1:

markmhendrickson:
It appears that the `userSession` object is always available in this function, even if there's no actual session with related `userData` available. As such this conditional doesn't catch the lack of user data before reaching the need to retrieve a `coreNode` value on line 1321:

Screen Shot 2020-11-11 at 21 02 22

markmhendrickson (changes requested):
This conditional doesn't seem to resolve the underlying problem with calling this function when no session user data is available.

markmhendrickson:
This appears to be affecting the todos app as well https://github.com/blockstack/todos/issues/55

agraebe:

PR #839, opened 2020-10-21
Fix cloned broadcast response causing issues in dev tools

### LINKED ISSUE #837: [Transactions] Funky broadcasting logic
_This is an issue with the stacks-transaction library, reporting here due to ongoing mono refactor work_

**Code in question**
https://github.com/blockstack/stacks-transactions-js/blob/ad6ac4fb605c38af6f8d001385e82e0a6c466bfd/src/builders.ts#L188-L201

**Issue**
There's some unusual logic in the broadcast transaction function, added by yours truly, which results in a legitimate network request being made, but **not shown in dev tools** HTTP panel.

This is likely owing to use of the `Request.clone()` method. 

**Resolve**
Avoiding use of `.clone()` with likely result in a more transparent mode of operation

yknl:
This PR fixes the transaction broadcast response cloning issue in https://github.com/blockstack/blockstack.js/issues/837 based on recommendation from @kyranjamie 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

No 

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated

### DIFF
--- a/packages/transactions/src/builders.ts
+++ b/packages/transactions/src/builders.ts
@@ -196,10 +196,11 @@ export async function broadcastRawTransaction(
     }
   }
 
+  const text = await response.text();
   try {
-    return (await response.clone().json()) as TxBroadcastResult;
+    return JSON.parse(text) as TxBroadcastResult;
   } catch (e) {
-    return await response.clone().text();
+    return text;
   }
 }
 


kyranjamie (approved):
Thanks @yknl

PR #841, opened 2020-10-22
feat: builder for unsigned contract call

yknl:
Migrated stacks-transaction-js PR from @hstove: https://github.com/blockstack/stacks-transactions-js/pull/131

## Description

Adds `makeUnsignedContractCall` function. This is built to be almost identical to the `makeUnsignedStacksTokenTransfer` function, which can accept one or more public keys. The actual `makeContractCall`, similarly, uses the unsigned variant to build the transaction, and then it simply signs the TX.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x] Tag 1 of @yknl, @zone117x, @reedrosenbluth for review

### DIFF
--- a/packages/auth/README.md
+++ b/packages/auth/README.md
@@ -1,11 +1,13 @@
-# `auth`
+# @stacks/auth
 
-> TODO: description
+Construct and decode authentication requests for Stacks apps.
+
+## Installation
+
+```
+npm install @stacks/auth
+```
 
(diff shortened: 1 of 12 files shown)


hstove:
Thanks Ken!

PR #842, opened 2020-10-22
feat: export connect from auth package

yknl:
This PR exports the Connect library from the auth package. Developers will no longer have to import Connect separately.

Currently, the build is failing because Connect has `react` as a peer dependency. 

Before:
```typescript
import { useConnect } from '@blockstack/connect';
import { UserSession } from '@stacks/auth';
```
After:
```typescript
import { useConnect, UserSession } from '@stacks/auth';
```

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

Yes 

## Testing information
Apps should continue to function as before after updating imports

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated


agraebe (approved):
lgtm!

PR #845, opened 2020-10-27
Blockstack/pr template

CharlieC3:
## Description

Updating pull request template as requested [here](https://github.com/blockstackpbc/devops/issues/469) to add more info on documentation.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

No

No

## Checklist
- [X] Tag 1 of @yknl, @zone117x, @reedrosenbluth for review


PULL_REQUEST_TEMPLATE.md
```diff
@@ -33,6 +33,18 @@ List the APIs or describe the functionality that this PR breaks.
 Workarounds for or expected timeline for deprecation
 
 ## Are documentation updates required?
+<!--
```
yknl, line 36: Is this supposed to be part of the markdown? It appears as is in the pull request template placeholder.

CharlieC3: @yknl This is supposed to just serve as a helpful comment for the PR creator. When the PR is submitted, it won't appear in the post, it's only viewable when editing the pull request.
[It's not visible when rendered](https://github.com/blockstack/stacks.js/blob/blockstack/pr-template/PULL_REQUEST_TEMPLATE.md#are-documentation-updates-required) but is [visible when editing](https://raw.githubusercontent.com/blockstack/stacks.js/blockstack/pr-template/PULL_REQUEST_TEMPLATE.md).

yknl: Ok thanks for clarifying. Assuming that people will need to replace the text in the template anyway, do you think it might not be necessary to have the comment brackets?

CharlieC3: Personally I feel like it could be helpful, and since it's a comment the PR creator wouldn't necessarily _need_ to delete the comment since it won't appear in the submitted PR anyways. But additionally I'm just forwarding along the change @agraebe proposed in the [reference issue here](https://github.com/blockstackpbc/devops/issues/469#issuecomment-713700569). So if you feel like it needs some changes or customization let me know!

agraebe: I'd just go with a consistent template. the implications here are very low, so I'd say we stick to it :)

PR #846, opened 2020-10-29
fix: yarn has link to dead blockstack/transaction

kyranjamie:
Install failed owing to a reference to this package that doesn't exists on npm.

I think rebuilding yarn wil help this


yknl:
this should be fixed in https://github.com/blockstack/stacks.js/pull/848

PR #683, opened 2019-07-31
Exclude unused bip39 wordlists from dist bundle

zone117x:
Reduce blockstack.js dist bundle size https://github.com/blockstack/blockstack.js/issues/582

Exclude unused bip39 wordlists from dist bundle. Removes around 240KB (uncompressed) of data. 

image

Acceptance criteria:
* Ensure the `dist/blockstack.js` bundle still works normally in web apps -- that is the only file this PR changes.

PR #847, opened 2020-10-29
fix: try to ignore wordlists for bip39

aulneau:
This PR brings over the equivalent of this change https://github.com/blockstack/stacks.js/pull/683 but for rollup. The bundle size of the keychain lib is pretty big, mostly due to these wordlists being included as far as i can tell.


yknl:
@aulneau should we close this since it didn't seem to work?

PR #849, opened 2020-10-30
feat(transactions): more exports

friedger:

## Description

As a developer, I would like 
* to verify signatures 
* to authorize sponsored transactions

This PR exports 
* `deserializeTransaction` from `transactions`
* `BufferReader` from `bufferReader`
* signature methods from `authorization`

These use case for sponsored tx is also mentioned with code in the README

## Type of Change
- [x] New feature
- [x] Bug fix
- [ ] Other

No

Yes

## Testing information
Needs tests with other apps and projects like stacks-blochain-api or testnet explorer

## Checklist
- [ ] `npm run test` passes (???? No such function)
- [ ] Changelog is updated (???? outdated)

### DIFF
--- a/packages/transactions/src/index.ts
+++ b/packages/transactions/src/index.ts
@@ -1,10 +1,16 @@
-export { StacksTransaction } from './transaction';
+export { StacksTransaction, deserializeTransaction } from './transaction';
+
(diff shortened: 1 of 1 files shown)


yknl (approved):
Thanks @friedger looks good to me

Issue #844, opened 2020-10-26
Create a stacking package 


agraebe:
Stacking integration code [like this](https://docs.blockstack.org/stacks-blockchain/integrate-stacking#step-3-display-stacking-info) will likely be duplicated across different projects. We can create a package to cover most important Stacking actions:

- [ ] call the `v2/pox`, `v2/info`, and `/extended/v1/info/network_block_times` endpoints to obtain all info required to display
- [ ] a "isEligible" method, calling the "can-stack-stx" function and checking account balance. returns boolean
- [ ] call the "stack-stx" function
- [ ] call the "get-stacker-info" function
- [ ] add a Clarity value type for BTC addresses
- [ ] call the faucet for Stacking 
- [ ] a "stackingExecution" method, returning a boolean if stacking is happening in the next cycle
- [ ] a "nextCycleStartingAt" method
- [ ] a "unlockingAt" method, requiring the starting cycle and number of cycles to lock up tokens

PR #850, opened 2020-11-02
feat: stacking library

yknl:
This PR adds a package to make it easier to integrate stacking. https://github.com/blockstack/stacks.js/issues/844

Usage documentation: https://github.com/blockstack/stacks.js/blob/feat/stacking/packages/stacking/README.md

Looking for input on naming for package and Stacker class.
`@stacks/stacking` or `@stacks/pox`?
`new Stacker()` or `new Pox()`?

Example:
```typescript
const { StacksTestnet } = require('@stacks/network');
const { Stacker } = require('@stacks/stacking');

const address = 'ST3XKKN4RPV69NN1PHFDNX3TYKXT7XPC4N8KC1ARH';
const stacker = new Stacker(address, network);

// Check stacking eligibility
const stackingEligibility = await stacker.canLockStx({poxAddress, cycles});

// Lock STX
const stackingResults = await stacker.lockStx({ 
  amountMicroStx,
  poxAddress,
  cycles,
  key,
  burnBlockHeight
});
```

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

Yes

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated


hstove:
Hey @yknl , just wanted to flag that there is a test suite for stacking in the wallet, and it would be great to keep at least the unit tests in the library! The integration tests will be harder to setup on CI.

https://github.com/blockstack/stacks-wallet/blob/release/stacking/app/utils/stacking/pox.spec.ts#L78

yknl:
Thanks for flagging, I've added new unit tests with jest to keep it consistent with the rest of the libs.

agraebe:
so great to see this. I'd vote for `@stacks/stacking` - PoX seems a bit too broad. Could you also include a docs update with this? We have some Stacking docs that would need updates

packages/stacking/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+const address = 'ST3XKKN4RPV69NN1PHFDNX3TYKXT7XPC4N8KC1ARH';
+const network = new StacksTestnet();
+const stacker = new Stacker(address, network);
+```
+
+## Check stacking eligibility
+
+```typescript  
+const stackingEligibility = await stacker.canLockStx({poxAddress, cycles});
```
agraebe: Should this be rather `isEligible`? If I recall correctly, the method doesn't check for balances of the STX account. So this could be true but locking wouldn't be enabled

packages/stacking/README.md
```diff
@@ -0,0 +1,94 @@
...
+//   reason: 'ERR_STACKING_INVALID_LOCK_PERIOD',
+// }
+```
+
+## Get PoX info
+```typescript
+const poxInfo = await stacker.getPoxInfo();
+
+// poxInfo:
+// {
```
agraebe, line 79: are we considering new types for the responses?

packages/stacking/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Stack STX
+```typescript
+const poxAddress = '1Xik14zRm29UsyS6DjhYg4iZeZqsDa8D3';
+const amountMicroStx = new BN(100000000000);
+const cycles = 10;
+const key = 'd48f215481c16cbe6426f8e557df9b78895661971d71735126545abddcd5377001';
+const burnBlockHeight = 2000;
+
+const stackingResults = await stacker.lockStx({
```
agraebe: would it make sense to keep the method names similar to the function names in the stacking contract? so, `stackStx`?

otherwise, the different name might reason some folks to assume there might be differences between this JS method and the Clarity method?

packages/stacking/README.md
```diff
@@ -0,0 +1,94 @@
...
+## Installation
+
+```
+npm install @stacks/stacking
+```
+
+## Initialization
+```typescript
+const { StacksTestnet } = require('@stacks/network');
+const { Stacker } = require('@stacks/stacking');
```
agraebe: this is very likely out of scope, but it might be beneficial to think about it before we work on it: how would delegation impact the interface?

packages/stacking/README.md
yknl: The implementation of `canLockStx()` will fetch the account balance and use that as an input to the contract function call, so it does take that into account.

packages/stacking/README.md
yknl: Yes, there's a `PoxInfo` type that the lib exports.

packages/stacking/README.md
yknl: I'm certainly for maintaining consistency with function names. But I feel `lockStx` provides more context as to what actually happens to the tokens. I believe that's how it is presented in the current stacks wallet UI as well.

packages/stacking/README.md
yknl: Yep, I can see it as an additional method in the class for delegation.

packages/stacking/README.md
agraebe: fantastic!

packages/stacking/README.md
agraebe: I guess it is _not_ lock-stx in the Clarity contract because the locking is happening elsewhere? so the abstraction in our library would make sense. I'll reside with your best judgment here

packages/network/src/index.ts
```diff
@@ -39,6 +47,8 @@ export class StacksMainnet implements StacksNetwork {
   accountEndpoint = '/v2/accounts';
   contractAbiEndpoint = '/v2/contracts/interface';
   readOnlyFunctionCallEndpoint = '/v2/contracts/call-read';
+
+  isMainnet= () => this.version === TransactionVersion.Mainnet;
```
kyranjamie: Prettier fail, would be cool to add a ci job to catch this

packages/stacking/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Installation
+
+```
+npm install @stacks/stacking
+```
+
+## Initialization
+```typescript
+const { StacksTestnet } = require('@stacks/network');
```
kyranjamie: es imports?

packages/stacking/src/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+        return AddressHashMode.SerializeP2SH;
+      case 196:
+        return AddressHashMode.SerializeP2SH;
+      default: 
+        throw new Error('Invalid pox address version');
+    }
+  }
+}
+
+export function convertBTCAddress(btcAddress: string) {
```
kyranjamie: `convertBtcAddress` ?

packages/stacking/README.md
```diff
@@ -0,0 +1,94 @@
+# @stacks/stacking [![npm](https://img.shields.io/npm/v/@stacks/stacking?color=red)](https://www.npmjs.com/package/@stacks/stacking)
```
agraebe, line 1: It looks like this readme doesn't cover all available methods:
- getCoreInfo
- getPoxInfo
- getTargetBlockTime
- getAccountBalance
- getCycleDuration
- secondsUntilNextCycle
- stackingEnabledNextCycle
- hasMinimumRequiredStxAmount
- canLockStx
- lockStx
- getStatus
- modifyLockTxFee

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+        return poxInfo.reward_cycle_length * targetBlockTime;
+      })
+  }
+
+  /**
+   * Get number of seconds until next reward cycle
+   * 
+   * @returns {Promise<number>} that resolves to a number if the operation succeeds
+   */
+  async secondsUntilNextCycle(): Promise<number> {
```
agraebe: suggest to rename this to `getSecondsUntilNextCycle` to be consistent

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+        return blocksToNextCycle * cycleDuration;
+      })
+  } 
+
+  /**
+   * Check if stacking is enabled for next reward cycle
+   * 
+   * @returns {Promise<boolean>} that resolves to a bool if the operation succeeds
+   */
+  async stackingEnabledNextCycle(): Promise<boolean> {
```
agraebe: would suggest `isStackingEnabledNextCycle` to be consistent. I'd not use `get`, `is` indicates the response is a boolean

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+
+        return callReadOnlyFunction({
+          network: this.network,
+          contractName,
+          contractAddress,
+          functionName: 'can-stack-stx',
+          senderAddress: this.address,
+          functionArgs: [
+              poxAddressCV,
+              uintCV(balance.toString()),
```
agraebe, line 259: shouldn't this be the balance - tx fees for stacking?

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+  }
+
+  /**
+   * Check if account can lock stx
+   * 
+   * @param {CanLockStxOptions} options - a required lock STX options object 
+   * 
+   * @returns {Promise<StackingEligibility>} that resolves to a StackingEligibility object if the operation succeeds
+   */
+  async canLockStx({
```
agraebe: After thinking about "lock" vs "stack", I think it would be better to use "stack" for the following reasons:
- locking is not the end-goal of this action, it is to get the rewards through stacking
- we use "stacking" in the stacking contract
- our messaging seems to be "start stacking stacks". Sticking to the terminology allows us to remove the mental overhead required to understand that "stacking effectively locks up tokens, therefore locking is the right method to use"

I'd advocate for renaming this method to `canStack`

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+  }
+
+  /**
+   * Generate and broadcast a stacking transaction to lock STX
+   * 
+   * @param {LockStxOptions} options - a required lock STX options object 
+   * 
+   * @returns {Promise<string>} that resolves to a broadcasted txid if the operation succeeds
+   */
+  async lockStx({
```
agraebe: if we'd adopt stacking vs locking, this could be simply `stack()`

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+          firstRewardCycle: firstRewardCycle.value.toNumber(),
+          lockPeriod: lockPeriod.value.toNumber(),
+          poxAddress: {
+            version: version.buffer,
+            hashbytes: hashbytes.buffer
+          },
+          btcAddress: '',
+        }
+      } else if (responseCV.type === ClarityType.OptionalNone) {
+        throw new Error(`No stacking status found for ${this.address}`);
```
agraebe: with this method labeled as `getStatus`, no stacking shouldn't be an error but one of the plausible responses, right?

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+      }
+    });
+  }
+
+  /**
+   * Adjust microstacks amount for locking after taking into account transaction fees
+   * 
+   * @returns {StacksTransaction} that resolves to a transaction object if the operation succeeds
+   */
+  modifyLockTxFee({ tx, amountMicroStx }: { tx: StacksTransaction; amountMicroStx: BN }) {
```
agraebe, line 415: can this be `setStackingTxFee`?

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+  async stackingEnabledNextCycle(): Promise<boolean> {
+    return (await this.getPoxInfo()).rejection_votes_left_required > 0;
+  }
+
+  /**
+   * Check if account has minimum require amount of Stacks for stacking
+   * 
+   * @returns {Promise<boolean>} that resolves to a bool if the operation succeeds
+   */
+  async hasMinimumRequiredStxAmount(): Promise<boolean> {
```
agraebe: could we shorten this to: `hasMinimumStx` ?

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+  cycles: number;
+  poxAddress: string;
+  amountMicroStx: BN;
+  burnBlockHeight: number;
+}
+
+export class Stacker {
+  constructor(
+    public address: string, 
+    public network: StacksNetwork = new StacksMainnet()) {}
```
agraebe: should this default to mainnet or rather testnet? Given that mainnet usage comes with real implications, would it maybe make sense to have the dev specify the mainnet as a type if confirmation?

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+          network: this.network,
+          contractName,
+          contractAddress,
+          functionName: 'can-stack-stx',
+          senderAddress: this.address,
+          functionArgs: [
+              poxAddressCV,
+              uintCV(balance.toString()),
+              uintCV(poxInfo.reward_cycle_id),
+              uintCV(cycles.toString())
```
agraebe, line 261: I believe there are a min and max for cycles - should we verify this before making the contract call?

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+   * @param {LockStxOptions} options - a required lock STX options object 
+   * 
+   * @returns {Promise<string>} that resolves to a broadcasted txid if the operation succeeds
+   */
+  async lockStx({
+    amountMicroStx,
+    poxAddress,
+    cycles,
+    key,
+    burnBlockHeight,
```
agraebe, line 292: do I understand correctly that this will default to the next cycle if not specified?

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+
+  /**
+   * Generate and broadcast a stacking transaction to lock STX
+   * 
+   * @param {LockStxOptions} options - a required lock STX options object 
+   * 
+   * @returns {Promise<string>} that resolves to a broadcasted txid if the operation succeeds
+   */
+  async lockStx({
+    amountMicroStx,
```
agraebe, line 288: will this be the max by default?

packages/stacking/src/constants.ts
```diff
@@ -0,0 +1,21 @@
+export enum StackingErrors {
```
agraebe, line 1: could we have a StackingError class with the error code, type (const), and a human-readable message? With that, developers could do `err.getMessage()` and simply display inside their app. If we just return the string, there's an opportunity for the developer to misinterpret the error string and/or misrepresent what it means to the user

packages/stacking/README.md
```diff
@@ -0,0 +1,94 @@
...
+// 800000000000 
+```
+
+## Stack STX
+```typescript
+const poxAddress = '1Xik14zRm29UsyS6DjhYg4iZeZqsDa8D3';
+const amountMicroStx = new BN(100000000000);
+const cycles = 10;
+const key = 'd48f215481c16cbe6426f8e557df9b78895661971d71735126545abddcd5377001';
+const burnBlockHeight = 2000;
```
agraebe: we should show how to get the latest block if the method doesn't get the height behind the scenes, right?

packages/stacking/src/index.ts
agraebe: we'd likely also consider error checking - past or current height shouldn't be accepted.

On a different note: Because the only options are block numbers in the future, we could also think about this field as a delta to current block height (`startInXBlocks`). For instance, starting in 3 blocks would just require an integer of 3. With that, developers don't have to think about the block height

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+ * Lock stx options
+ *
+ * @param  {String} key - private key to sign transaction
+ * @param  {String} poxAddress - the reward Bitcoin address
+ * @param  {number} cycles - number of cycles to lock
+ * @param  {BigNum} amountMicroStx - number of microstacks to lock
+ * @param  {number} burnBlockHeight -the burnchain block height to begin lock
+ */
+export interface LockStxOptions {
+  key: string;
```
agraebe: could we be more descriptive here? something like `stackerPrivateKey`?

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+              uintCV(balance.toString()),
+              uintCV(poxInfo.reward_cycle_id),
+              uintCV(cycles.toString())
+            ],
+        });
+      })
+      .then((responseCV: ClarityValue) => {
+        if(responseCV.type === ClarityType.ResponseOk) {
+          return {
+            eligible: true
```
agraebe, line 268: This might be a bit confusing because the method name is `canLockStx`, no `isEligible`. Should we rather return something like `canLockStx` instead to be consistent?

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+  rejection_fraction: number;
+  reward_cycle_id: number;
+  reward_cycle_length: number;
+  rejection_votes_left_required: number;
+}
+
+export interface StackerInfo {
+  amountMicroStx: string;
+  firstRewardCycle: number;
+  lockPeriod: number;
```
agraebe: I'm assuming this method will be called to show either # of blocks until unlocked or timestamp of unlock event. Could we calculate these and return them as well for convenience? Something like:

- secondsUntilUnlock
- unlockBlockHeight

packages/stacking/README.md
```diff
@@ -0,0 +1,94 @@
...
+  poxAddress,
+  cycles,
+  key,
+  burnBlockHeight
+});
+
+// stackingResults:
+// {
+//   txid: '0xf6e9dbf6a26c1b73a14738606cb2232375d1b440246e6bbc14a45b3a66618481',
+//   transaction: 'https://testnet-explorer.now.sh/txid/0xf6e9dbf6a26c1b73a14738606cb2232375d1b440246e6bbc14a45b3a66618481'
```
agraebe: the URL needs to be updated, right? we shouldn't link to the `now.sh` url

packages/stacking/README.md
```diff
@@ -0,0 +1,94 @@
...
+## Get account stacking status
+```typescript
+const stackingStatus = await stacker.getStatus();
+
+// stackingStatus:
+// {
+//   amount_microstx: '80000000000000',
+//   first_reward_cycle: 18,
+//   lock_period: 10,
+//   pox_address: {
```
agraebe: the dev will submit a BTC address in a different format. It would be great if we could return the address in the same format as entered, so the dev wouldn't need to understand what we're decoding this to behind the scenes

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+      burnBlockHeight,
+    });
+    const tx = await makeContractCall({
+      ...txOptions,
+      senderKey: key,
+    });
+
+    const res = await broadcastTransaction(tx, txOptions.network as StacksNetwork);
+    if (typeof res === 'string') {
+      return res;
```
agraebe, line 312: shouldn't this also return the explorer URL?

packages/stacking/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+        const someCV = responseCV as SomeCV;
+        const tupleCV: TupleCV = someCV.value as TupleCV;
+        const poxAddress: TupleCV = tupleCV.data["pox-addr"] as TupleCV;
+        const amountMicroStx: UIntCV = tupleCV.data["amount-ustx"] as UIntCV;
+        const firstRewardCycle: UIntCV = tupleCV.data["first-reward-cycle"] as UIntCV;
+        const lockPeriod: UIntCV = tupleCV.data["lock-period"] as UIntCV;
+        const version: BufferCV = poxAddress.data['version'] as BufferCV;
+        const hashbytes: BufferCV = poxAddress.data['hashbytes'] as BufferCV;
+      
+        return {
```
agraebe, line 387: would it be possible to return the txid/explorer URL for the stacking transaction that "reasoned" the status?

packages/stacking/README.md
```diff
@@ -0,0 +1,94 @@
...
+  amountMicroStx,
+  poxAddress,
+  cycles,
+  key,
+  burnBlockHeight
+});
+
+// stackingResults:
+// {
+//   txid: '0xf6e9dbf6a26c1b73a14738606cb2232375d1b440246e6bbc14a45b3a66618481',
```
agraebe: this is outside of the stacking lib, but should one of our packages (maybe `transactions`) expose a helper method to wait for a tx to complete? In the stacking tutorial, we share a `waitForTransactionSuccess` code sample. I'd assume many devs would want to implement that

packages/stacking/src/index.ts
kyranjamie: I think there's a conversation to be had about network defaults.

Personally I'd err on the side of no default at all. Leave it to the user to be explicit every time. (To avoid "Wait, what's the default again?")

packages/stacking/src/utils.ts
zone117x: Had to use this function recently for testing in the sidecar. The current name is meaningless, should be something like `decodeBtcAddress` or `destructureBtcAddress`

yknl: Changed to `decodeBtcAddress`

packages/stacking/src/index.ts
yknl: I think rather than calculating this based on fees which would require another network call, this should be left up to the developer. Additionally this would not take into account sponsored transactions.

packages/stacking/src/index.ts
yknl: Updated to `canStack`

packages/stacking/src/index.ts
yknl: I think it would be easier for consumers to handle an error vs expecting a differently shaped response.

packages/stacking/src/index.ts
yknl: That name seems confusing since it doesn't actually set the fee, but modifies the locked amount.

packages/stacking/src/index.ts
yknl: Removed the default

packages/stacking/src/index.ts
yknl: Since this min/max constant is set in the contract, it's probably a better idea to let the contract check this vs adding a constant in the library which may need to be updated if the contract changes.

packages/stacking/src/index.ts
yknl: It would not default to any value. I think giving developers more control by setting the exact block here is better than making it slightly easier to use but taking away ability to be exact.

packages/stacking/src/index.ts
yknl: It does not have a default.

packages/stacking/src/constants.ts
yknl: Added a helper function to address this.

packages/stacking/README.md
yknl: it shouldn't be showing the explorer URL, fixed the response.

packages/stacking/README.md
yknl: If you look at https://github.com/blockstack/stacks.js/pull/850/files#diff-c7ec1e9b062dbf6b31a30a695d5d88049b306aae4e07be04b6559c9832d0508dR5 I don't think we have enough information to reverse the data to the same BTC address entered.

packages/stacking/src/index.ts
yknl: I think the consumer of the lib should construct the link vs in the lib

packages/stacking/README.md
yknl: A few reasons why I don't think that would be a good idea: 
- The time between broadcast and the transaction being completed might be a very long time on mainnet. Doesn't seem like good engineering practice to have a long running wait like that
- It feels like it should be part of the UI layer.

packages/stacking/src/index.ts
yknl: That would take an extra network call to fetch block info in order to calculate. Probably best to let the developer do that separately if they need it.

kyranjamie:
For the name @yknl, how about `new StackingClient()`?

yknl:
> For the name @yknl, how about `new StackingClient()`?

I like it, what do you think @agraebe ?

packages/stacking/README.md
agraebe: discussed in discord, can't be done unfortunately

packages/stacking/src/index.ts
agraebe: well, I guess the question is how would they use this method. My assumption is it'll be used to show the data. It's probably fair to say the progress or unlocking block/time would be important to show. 

could we allow an optional param to request this data, eg `getStatus(unlockDetails: boolean)`?

packages/stacking/src/index.ts
agraebe: Maybe we could double click on this one. Because the developer will likely assume that a positive response means they're good to go, which wouldn't be the case _always_. 

- Curious about the sponsored txs: are they important for stacking? why would someone pay for these tx fees?
- Could we provide an optional method like `includeTxFees` to account for this not being critical?

packages/stacking/src/index.ts
agraebe: you're right

packages/stacking/src/index.ts
agraebe: I can see the tradeoffs of not being exact. mainly, I'm assuming that devs would want to get the latest block height right before calling this method. That could be avoided with a default and a delta, but I don't feel strongly either way. your call

packages/stacking/src/index.ts
yknl: It would be better as a separate method then

packages/stacking/src/index.ts
agraebe: why? I think we should do it in order to avoid the dev building the URLs - they'd want to use the same URL that we would provide anyways, right? One reason I see for not doing this is if we were to stay open to other explorers and not prioritizing one over the other, but that doesn't seem relevant right now

packages/stacking/src/index.ts
agraebe: what if we were to wrap responses into something like this:

```json
{
   "status": "<active/inactive>",
   "details": {
      // status details
   }
}
```

packages/stacking/src/index.ts
agraebe: right, makes sense

packages/stacking/src/index.ts
agraebe: could be if devs would want to call that outside of the getstatus flow. either, way, I think that info should be easily available to them

packages/stacking/src/index.ts
yknl: We should maintain separation of concern in the code. The URL can be easily constructed by the developer should they need it. And we shouldn't favour any explorer over another, so we shouldn't start doing that here. 

I don't think the goal should be to allow the developer to write as few lines of code as possible. And we should keep in mind good engineering practices and maintainability of code.

packages/stacking/src/index.ts
yknl: It wouldn't be 100% in both cases, since fees can fluctuate and balances, minimums can change between calls. 

> Curious about the sponsored txs: are they important for stacking? why would someone pay for these tx fees?

A Stacking app or custodial service could take care of the fees for the user.

packages/stacking/src/index.ts
aulneau: I agree, I think it's not the responsibility of this package to push people to the explorer, but whatever would be using this package to expose the url

agraebe: as per discussion in the meeting, we can close this out

packages/stacking/README.md
yknl: Yes we're showing that in the tutorial doc

packages/stacking/src/index.ts
yknl: Changed to `privateKey`

packages/stacking/src/index.ts
yknl: Updated to `canStack`

packages/stacking/src/index.ts
yknl: the `getStatus` function is now making an additional call to the account API and returns the unlock block as part of the response

PR #854, opened 2020-11-10
fix: remove class extension of smartbuffer to fix build issues

### LINKED ISSUE #853: deserializeCV does not work due to bad SmartBuffer construction
**Describe the bug**
When calling `deserializeCV` or `callReadOnlyFunction` the following error occurs:
```
 TypeError: Class constructor SmartBuffer cannot be invoked without 'new'
      at new BufferReader (node_modules/@stacks/transactions/src/bufferReader.ts:52:7)
      at deserializeCV (node_modules/@stacks/transactions/src/clarity/deserialize.ts:24:50)
```

@stacks/transactions 1.0.0-beta.9

**Expected behavior**
Just return the correct cv

**Additional context**
https://github.com/friedger/clarity-demo/blob/ecaeb4e61986c8e0386708cf722980d83af93bf9/scripts/deploy.ts#L66

yknl:
This PR changes the way SmartBuffer is used in the BufferReader implementation. It will no longer be extending the SmartBuffer class as it causes build issues.

Fixes: https://github.com/blockstack/stacks.js/issues/853

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/transactions/src/bufferReader.ts
+++ b/packages/transactions/src/bufferReader.ts
@@ -42,28 +42,46 @@ export function isEnum<T extends string, TEnumValue extends number>(
   return isEnum(enumVariable, value);
 }
 
-export class BufferReader extends SmartBuffer {
+export class BufferReader {
+  smartBuffer: SmartBuffer;
+
   static fromBuffer(buffer: Buffer): BufferReader {
     return new BufferReader({ buff: buffer });
   }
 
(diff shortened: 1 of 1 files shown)


friedger (approved):
Looks good! `deserializeCV` and `callReadOnlyFunction` work for me :-)

PR #859, opened 2020-11-17
fix loss of compression byte in private key leading to incorrect stx addresses

psq:
Given the same passphrase, the wallet signer in stacks.js does not generate the same STX addresses (compared to the stacks wallet, blockstack-cli, stacks-gen).

The reason is that 

```
key !== bip32.fromBase58(stxAddressKeychain.toBase58(key))
```

And this is because of the use of 33 bytes private keys, not 32 bytes, the 33rd byte is used to represent whether they key is compressed or not, but is not taken into account by `toBase58`.

Solution is to store the private key as the original buffer, not in encoded base58 format which does not encode the compression.  This also has the benefit that the key does not need to be decoded each time it is used.

Now the STX addresses will match what the other tools will do

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No.  The addresses people are used to when they use connect will change, for the better, but given that testnet gets reset often, this will have no negative effect.

no documentation update needed

## Testing information
manually testing the passphrase now generates the same addresses as the wallet, blockstack-cli, stacks-gen would be nice (I did for the testnet passphrase I use most) and I finally get the expected address 😂 

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/keychain/src/wallet/index.ts
+++ b/packages/keychain/src/wallet/index.ts
@@ -64,7 +64,7 @@ export interface ConstructorOptions {
   encryptedBackupPhrase: string;
   identities: Identity[];
   configPrivateKey: string;
-  stacksPrivateKey: string;
+  stacksPrivateKey: Buffer;
   walletConfig?: WalletConfig;
 }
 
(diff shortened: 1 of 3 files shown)


yknl:
Thanks for the PR @psq! @hstove do you want to verify first before merging?

kyranjamie (approved):
Thanks @psq

PR #862, opened 2020-11-20
More transaction lib exports

zone117x:
Add several exports from the transaction lib that are needed by the sidecar.


yknl (approved):
Thanks, LGTM

PR #869, opened 2020-12-01
fix: remove typing path to non-existant directory

kyranjamie:
Moved more relevant properties to top of file, but most importantly removed:

`"typings": "dist/keychain/src/index.d.ts",`

which pointed the typings to a file that doesn't exist. No types = broken builds

A `types` property, that does the same, remains at the top of the file.

### DIFF
--- a/packages/keychain/package.json
+++ b/packages/keychain/package.json
@@ -7,8 +7,14 @@
     "Blockstack",
     "Keychain"
   ],
+  "sideEffects": false,
+  "main": "dist/index.js",
+  "module": "dist/index.esm.js",
+  "umd:main": "dist/keychain.umd.production.js",
+  "unpkg": "dist/keychain.cjs.production.min.js",
+  "gitHead": "77b4d6d531b74996e4b7a0cbd1cf5b8358a690ce",
   "author": "Hank Stoever",
-  "types": "./dist/index.d.ts",
+  "typings": "./dist/index.d.ts",
   "homepage": "https://blockstack.org",
   "contributors": [
     {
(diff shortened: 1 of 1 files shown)


packages/keychain/package.json
```diff
@@ -81,15 +90,5 @@
...
     "triplesec": "^3.0.27",
     "zone-file": "^1.0.0"
-  },
-  "bugs": {
-    "url": "https://github.com/blockstack/blockstack.js/issues"
-  },
-  "sideEffects": false,
-  "main": "dist/index.js",
-  "module": "dist/index.esm.js",
-  "typings": "dist/keychain/src/index.d.ts",
```
aulneau, line 91: @kyranjamie typings here needs to exist. Wherever the types are built, we need to include that path

aulneau (changes requested):

packages/keychain/package.json
kyranjamie: @aulneau  That path doesn't exist, and a "types" equivalent properties does, that points to correct path

kyranjamie: `"types": "./dist/index.d.ts",`

aulneau: ah my bad!

yknl (commented):
`types` and `typings` should be equivalent in the package.json. Since we're using `typings` in the other libs should we not be consistent?

kyranjamie:
@yknl using typings

kyranjamie:
@yknl using `typings` instead of `types`

kyranjamie:
Would be great to get this merged & published today, it breaks a wallet PR 🙏🏼

yknl (approved):

PR #870, opened 2020-12-02
ci: lint prettier, add ci job

kyranjamie:
Implementing `@stacks/stacking` client, noticed we're not conforming to the org prettier standard. 

This PR fixes those formatting errors and adds a CI job to check


yknl (approved):
Thanks for the fix, looks good to me.

PR #877, opened 2020-12-07
[storage] fix: dangerouslyIgnoreEtag

### LINKED ISSUE #876: [storage] putFile no longer has the dangerouslyIgnoreEtag option
These changes:
https://github.com/blockstack/stacks.js/pull/793/files#diff-189df448d4739484d3f18a1acd3d9f4b947564f111c3fb4ef7c14c284ec4c430R86

are no longer in `master` -> https://github.com/blockstack/stacks.js/blob/master/packages/storage/src/storage.ts#L39

Curious to learn how this regression happened. cc @yknl

aulneau:
There was a regression and this change was no longer included in the latest version of `@stacks/storage`. I also have been doing some testing with ES modules in a side project, and found that the config I had previously added to preserve modules when outputing es modules made little to no real effect, and as such I am removing the custom tsdx config.

fixes https://github.com/blockstack/stacks.js/issues/876


aulneau:
@reedrosenbluth I'm tagging you here too, would you be able to check and see if there is anything else related to etags that might be been lost in this regression?

yknl (approved):
Thanks for fixing this, looks good to me 👍 
Will wait for @reedrosenbluth to confirm before merge

reedrosenbluth (approved):

reedrosenbluth:
LGTM!

PR #881, opened 2020-12-11
fix(keychain): typing reference

kyranjamie:
Updated (again) to correct reference


yknl:
@kyranjamie have you already published beta.21? We're incrementing all of the monorepo package version numbers at the same time using lerna. So it's better to not increment a package separately.

kyranjamie:
@yknl yes, sorry I meant to follow up here as I realised I'd just deployed the single package

Issue #820, opened 2020-09-08
Update needed to CLI output for transactions to fix 2 issues


timstackblock:
**Describe the bug**
Currently we have 2 issues the TXID is missing the first 2 characters 
The link to explorer is not pointing to the current version of the explorer

This is what we get 

{
  txid: 'a5659553e404089f861efba24394268a1bfefbe2d44a1301d45e2505a8ce6c79',
  transaction: 'https://testnet-explorer.now.sh/txid/0xa5659553e404089f861efba24394268a1bfefbe2d44a1301d45e2505a8ce6c79'
}

This is what we should get

{
  txid: '0xa5659553e404089f861efba24394268a1bfefbe2d44a1301d45e2505a8ce6c79',
  transaction: 'https://testnet-explorer.blockstack.org/txid/0xa5659553e404089f861efba24394268a1bfefbe2d44a1301d45e2505a8ce6c79'
}

**To Reproduce**
1. Go to the CLI 
2. Send a a token send transaction
3. look at the output and you will see the 2 issues above

**Expected behavior**
{
  txid: '0xa5659553e404089f861efba24394268a1bfefbe2d44a1301d45e2505a8ce6c79',
  transaction: 'https://testnet-explorer.blockstack.org/txid/0xa5659553e404089f861efba24394268a1bfefbe2d44a1301d45e2505a8ce6c79'
}

**Screenshots**

**Terminal Log**
If applicable, add console log to help explain your problem.
<img width="911" alt="Screen Shot 2020-09-08 at 11 05 47 AM" src="https://user-images.githubusercontent.com/49165468/92494137-524e4780-f1c3-11ea-8c27-350452b9b58d.png">

yknl:
Moving this issue to the stack.js monorepo since the CLI has already been migrated and changes will be made in the new repo.

agraebe:
it looks like what is outstanding here is the URL to the explorer (currently printing: `https://testnet-explorer.now.sh/txid/`). @yknl is that correct? could we pick this issue up and fix in this sprint?

PR #887, opened 2020-12-17
fix: CLI tx response explorer URL

yknl:
This PR changes the explorer URL return by the CLI after submitting a transaction to the correct testnet URL. Reference docs have been updated as well.

https://github.com/blockstack/stacks.js/issues/820

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

References updated

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated

### DIFF
--- a/packages/cli/src/argparse.ts
+++ b/packages/cli/src/argparse.ts
@@ -345,7 +345,10 @@ export const CLI_ARGS = {
         '    $ export PAYMENT="bfeffdf57f29b0cc1fab9ea197bb1413da2561fe4b83e962c7f02fbbe2b1cd5401"\n' +
         '    $ stx call_contract_func SPBMRFRPPGCDE3F384WCJPK8PQJGZ8K9QKK7F59X contract_name' +
         '      contract_function 1 0 "$PAYMENT"\n' +
-        '    a9d387a925fb0ba7a725fb1e11f2c3f1647473699dd5a147c312e6453d233456\n' +
+        '     {\n' +
+        '       txid: \'0x2e33ad647a9cedacb718ce247967dc705bc0c878db899fdba5eae2437c6fa1e1\',' +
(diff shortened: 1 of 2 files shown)


packages/cli/src/argparse.ts
```diff
@@ -345,7 +345,10 @@ export const CLI_ARGS = {
         '    $ export PAYMENT="bfeffdf57f29b0cc1fab9ea197bb1413da2561fe4b83e962c7f02fbbe2b1cd5401"\n' +
         '    $ stx call_contract_func SPBMRFRPPGCDE3F384WCJPK8PQJGZ8K9QKK7F59X contract_name' +
         '      contract_function 1 0 "$PAYMENT"\n' +
-        '    a9d387a925fb0ba7a725fb1e11f2c3f1647473699dd5a147c312e6453d233456\n' +
+        '     {\n' +
+        "       txid: '0x2e33ad647a9cedacb718ce247967dc705bc0c878db899fdba5eae2437c6fa1e1'," +
+        "       transaction: 'https://explorer.blockstack.org/txid/0x2e33ad647a9cedacb718ce247967dc705bc0c878db899fdba5eae2437c6fa1e1'" +
```
aulneau: why is this `explorer.blockstack.org` instead of `testnet-explorer.blockstack.org`?

aulneau:
I wonder if we should have it be conditional in some way? testnet/mainnet vars?

packages/cli/src/argparse.ts
yknl: The example is for a mainnet transaction. The help examples don't currently check for the testnet/mainnet flag.

agraebe:
could we revisit this and:

- [x] update old explorer URLs
- [ ] resolve the discussion above

PR #897, opened 2021-01-11
Feat: `@stacks/wallet-sdk`

hstove:
I didn't like the design of `@stacks/keychain` - I built it when we were just an "authenticator". As we moved towards a full-on wallet, I felt many aspects of Keychain to be lacking.

`@stacks/wallet-sdk` is built 100% intentionally for Stacks Wallets, and it includes a lot of the learnings and things I wish I had.

This library is in use in the Stacks Wallet for Web, in a PR.

I have a WIP branch of documentation updates for this library. Ultimately, I'm very excited about it. Hopefully it can come in handy for @yknl as he builds his wallet.


packages/wallet-sdk/src/derive.ts
```diff
@@ -0,0 +1,84 @@
...
+    throw new TypeError('Unable to derive config key for wallet identities');
+  }
+  return derivedConfigKey;
+};
+
+/**
+ * Generate a salt, which is used for generating an app-specific private key
+ * @param rootNode
+ */
+export const deriveSalt = async (rootNode: BIP32Interface) => {
```
kyranjamie, line 63: Mixing fn expressions and declarations in `/encryption.ts`. For top level fns like this my pref would be function declarations, as should make for a more helpful callstack.

packages/wallet-sdk/src/encryption.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Decrypt an encrypted mnemonic phrase with a password.
+ * Legacy triplesec encrypted payloads are also supported.
+ * @param data - Buffer or hex-encoded string of the encrypted mnemonic
+ * @param password - Password for data
+ * @return the raw mnemonic phrase
+ */
+export async function decrypt(dataBuffer: Buffer | string, password: string): Promise<string> {
+  const result = await decryptMnemonic(dataBuffer, password, triplesecDecrypt);
+  return result;
```
kyranjamie:
```suggestion
  return decryptMnemonic(dataBuffer, password, triplesecDecrypt);
```

packages/wallet-sdk/src/encryption.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Encrypt a raw mnemonic phrase to be password protected
+ * @param phrase - Raw mnemonic phrase
+ * @param password - Password to encrypt mnemonic with
+ * @return The encrypted phrase
+ * */
+export async function encrypt(phrase: string, password: string): Promise<Buffer> {
+  const result = await encryptMnemonic(phrase, password);
+  return result;
```
kyranjamie: Guess this is copied from keychain, but this looks nicer
```suggestion
  return encryptMnemonic(phrase, password);
```

packages/wallet-sdk/src/models/account.ts
```diff
@@ -0,0 +1,145 @@
...
+    {
+      profileUrl,
+    },
+    undefined,
+    appPrivateKey,
+    undefined,
+    transitPublicKey,
+    gaiaHubUrl,
+    undefined,
+    associationToken
```
kyranjamie, line 151: 😭

packages/wallet-sdk/src/derive.ts
hstove: I'm not sure I follow this comment. Sorry, probably just the lack of coffee.

hstove: Oh, you mean `export function x()` vs `export const x = () => {}`? That makes sense, and I like the thought. I tried to be consistent by using destructured options (for more than one arg) and immutability, and this would be another good thing to be consistent about.

hstove:
Remaining items:

- Update functions to use function declarations, as Kyran suggested
- Add a README

agraebe:
@hstove what's the next step here?

hstove:
> @hstove what's the next step here?

@agraebe Basically what is in the comment above yours 😄  

This code is already heavily used and tested in the Web Wallet. I'd just like to add a small README with basic usage, and then I think we can merge it. Since it's a new library, we don't really need to worry about it effecting other things.

hstove:
The remaining blocker on this PR is fixing the Typecheck CI job, and then making sure all the tests pass in CI.

agraebe:
cc @andresgalante

andresgalante:
@kyranjamie @aulneau Do you need context from Hank to pick this one up?

kyranjamie:
Sorry I haven't really given this one much attention. As it's already used in the production extension wallet, regardless of this lib's future, we should merge it in.

Will take a look at fixing the CI

kantai (approved):
This all looks good to me, and way more clear than the old wallet code in encryption.js!

zone117x:
Looks like this need rebased onto master, and likely some of the newly added project files need modified to match the latest mono-repo conventions. @ahsan-javaid is this something you can take on?

ahsan-javaid:
> Looks like this need rebased onto master, and likely some of the newly added project files need modified to match the latest mono-repo conventions. @ahsan-javaid is this something you can take on?

@zone117x Progress update on this:
- Fixed build errors due to missing config in sdk-wallet ts config file
- `npm run build` passes
- `npm run test` passes
- `npm run typecheck` passes
- rebased on master branch
- Updated versions in sdk-wallet package json
Let me know if it needs further improvements.

yarn.lock
```diff
@@ -0,0 +1,10987 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
zone117x: This file should not be committed

packages/wallet-sdk/jest.config.js
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
zone117x: Should this file be updated to uses the conventions from the other packages, for example https://github.com/blockstack/stacks.js/blob/master/packages/transactions/jest.config.js ?
The existing jest.config.js files are known to work correctly with debuggers.

yarn.lock
ahsan-javaid: @zone117x Added this change.

packages/wallet-sdk/jest.config.js
ahsan-javaid: @zone117x Added these changes to follow test conventions from other packages

packages/wallet-sdk/src/derive.ts
```diff
@@ -0,0 +1,102 @@
+import { BIP32Interface } from 'bip32';
+import { Buffer } from 'buffer';
```
zone117x: Usage of the NodeJS `buffer` has been replaced with a runtime agnostic lib, see example usage here:
https://github.com/blockstack/stacks.js/blob/8e99b763fbbd486cce87749e9e1b21015c61e946/packages/transactions/src/clarity/serialize.ts#L1

This allows the libs to work in both Node.js, browsers, and other runtimes without requiring extra bundling/polyfill utils.

Can you update usages to this convention?

packages/wallet-sdk/src/models/account.ts
```diff
@@ -0,0 +1,153 @@
...
+import {
+  DEFAULT_PROFILE,
+  fetchAccountProfileUrl,
+  fetchProfileFromUrl,
+  Profile,
+  signAndUploadProfile,
+} from './profile';
+import { ECPair } from 'bitcoinjs-lib';
+import { connectToGaiaHubWithConfig, getHubInfo, makeGaiaAssociationToken } from '../utils';
+import { Buffer } from 'buffer';
```
zone117x: Same here (see previous comment about buffer usage)

packages/wallet-sdk/src/derive.ts
ahsan-javaid: @zone117x updated these usages to follow the convention.

packages/wallet-sdk/src/models/account.ts
ahsan-javaid: @zone117x updated these usages to follow the convention.

zone117x (approved):
Thanks for fixing this up @ahsan-javaid!

zone117x (approved):
Thanks for fixing this up @ahsan-javaid!

PR #903, opened 2021-01-13
feat: add delegate methods

### LINKED ISSUE #902: Stacking lib: Add methods for delegation
Delegation requires 4 additional contract calls to `pox.clar`:

- delegate-stx
- revoke-stx
- delegate-stack-stx
- stack-aggregation-commit

agraebe:
## Description

Adding delegation methods

For details refer to issue #902

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

Yes, updates needed in:
- README
- Delegation tutorial
- Generated docs

## Testing information
Test behavior for the similar methods is used

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated


agraebe:
this will include a fix for the stacking example in https://github.com/blockstack/stacks.js/issues/913

agraebe:
I couldn't run the `yarn build:docs` command locally. Would appreciate help to fix that

```
> yarn build:docs
yarn run v1.22.4
$ rimraf docs && cross-env typedoc --name "Stacks.js $npm_package_version Library Reference" --readme ./readme.md --tsconfig tsconfig.typedoc.json --out docs --json docs/docs.json packages/**/src

Using TypeScript 4.0.5 from /Users/alexandergraebe/Development/blockstack.js/node_modules/typescript/lib
Error: /Users/alexandergraebe/Development/blockstack.js/packages/auth/src/appConfig.ts(1)
 Cannot find module '@stacks/common' or its corresponding type declarations.
Error: /Users/alexandergraebe/Development/blockstack.js/packages/auth/src/dids.ts(0)
 Cannot find module '@stacks/common' or its corresponding type declarations.
Error: /Users/alexandergraebe/Development/blockstack.js/packages/auth/src/messages.ts(4)
 Cannot find module '@stacks/common' or its corresponding type declarations.
...
```

agraebe:
I'd appreciate a review of this PR 🙏 thanks!

packages/stacking/README.md
```diff
@@ -145,3 +158,99 @@ stackingStatus:
...
+// a BTC address for reward payouts to the delegator
+const deletagorPoxAddress = "msiYwJCvXEzjgq6hDwD9ueBka6MTfN962Z";
+
+const delegatorClient = new StackingClient(delegatorAddress, network);
+
+// STX addresses of all your delegatees / stackers
+const stackers = [address];
+
+
+// note this does not require an await
```
hstove: That can't be right - how is this not async? Is the transaction not broadcasted?

From reading the implementation, this is async

packages/stacking/README.md
```diff
@@ -145,3 +158,99 @@ stackingStatus:
...
+//     txid: '0xf6e9dbf6a26c1b73a14738606cb2232375d1b440246e6bbc14a45b3a66618481',
+//   }
+// ]
+```
+
+#### Commit to stacking
+
+```typescript
+// burn height at which the delegation relationship should be revoked (optional)
+const untilBurnBlockHeight = 5000;
```
hstove: It doesn't look like this variable is used?

hstove (changes requested):
Small changes requested.

packages/stacking/README.md
agraebe: oh, you're right!

packages/stacking/README.md
agraebe: good catch - it seems to be a leftover

packages/stacking/tests/stacking.test.ts
```diff
@@ -68,7 +69,7 @@ test('check stacking eligibility true', async () => {
   const network = new StacksTestnet();
 
   // CV responseOk value = true
-  const functionCallResponse = { type: 7, value: { type: 3 } } 
+  const functionCallResponse = { type: 7, value: { type: 3 } }
```
reedrosenbluth: Can you replace this with:

```typescript
const functionCallResponse = responseOkCV(trueCV());
```

packages/stacking/tests/stacking.test.ts
```diff
@@ -115,7 +116,7 @@ test('check stacking eligibility false bad cycles', async () => {
   const errorType = new BN(2);
   const expectedErrorString = StackingErrors[StackingErrors.ERR_STACKING_INVALID_LOCK_PERIOD];
   // CV responseError value = CV uint 2
-  const functionCallResponse = { type: 8, value: { type: 0, value: errorType } } 
+  const functionCallResponse = { type: 8, value: { type: 0, value: errorType } }
```
reedrosenbluth: Can you replace this with:

```typescript
const functionCallResponse = responseErrorCV(uintCV(2));
```

reedrosenbluth (commented):
Also requested a few small changes

hstove:
> I couldn't run the `yarn build:docs` command locally. Would appreciate help to fix that

I've seen this too. I think you just need to run `yarn build` first.

packages/stacking/tests/stacking.test.ts
agraebe: sure!

packages/stacking/tests/stacking.test.ts
agraebe: Looking at the error types, an `int` is required (https://docs.blockstack.org/references/stacking-contract#err_stacking_invalid_lock_period). This worked:

```
const functionCallResponse = responseErrorCV(intCV(2));
```

agraebe:
> > I couldn't run the `yarn build:docs` command locally. Would appreciate help to fix that
> 
> I've seen this too. I think you just need to run `yarn build` first.

thanks! that did it for me :)

packages/stacking/src/index.ts
```diff
@@ -19,6 +19,8 @@ import {
   ResponseErrorCV,
```
reedrosenbluth, line 18: Can you remove the commented out imports from the top of this file?

packages/stacking/src/index.ts
```diff
@@ -302,6 +347,139 @@ export class StackingClient {
...
+   *
+   * @returns {Promise<string>} that resolves to a broadcasted txid if the operation succeeds
+   */
+  async delegateStx({
+    amountMicroStx,
+    delegateTo,
+    untilBurnBlockHeight,
+    poxAddress,
+    privateKey,
+  }: DelegateStxOptions) {
```
reedrosenbluth: can you add the return type `Promise<TxBroadcastResult>` to this and the functions below?

packages/stacking/src/index.ts
```diff
@@ -302,6 +347,139 @@ export class StackingClient {
...
+    const tx = await makeContractCall({
+      ...txOptions,
+      senderKey: privateKey,
+    });
+
+    const res = await broadcastTransaction(tx, txOptions.network as StacksNetwork);
+    if (typeof res === 'string') {
+      return res;
+    }
+    throw new Error(`${res.error} - ${res.reason}`);
```
reedrosenbluth: Can you replace the above (and similarly below) with:

```typescript
return await broadcastTransaction(tx, txOptions.network as StacksNetwork);
```

packages/stacking/src/index.ts
```diff
@@ -302,6 +347,139 @@ export class StackingClient {
...
```
reedrosenbluth: Same as above

packages/stacking/src/index.ts
```diff
@@ -302,6 +347,139 @@ export class StackingClient {
...
```
reedrosenbluth: same as above

packages/stacking/src/index.ts
```diff
@@ -302,6 +347,139 @@ export class StackingClient {
...
```
reedrosenbluth: same as above

reedrosenbluth (approved):

packages/stacking/src/index.ts
```diff
@@ -341,6 +519,144 @@ export class StackingClient {
...
+      functionName: 'stack-aggregation-commit',
+      functionArgs: [address, uintCV(rewardCycle)],
+      validateWithAbi: true,
+      network,
+    };
+    return txOptions;
+  }
+
+  getRevokeDelegateStxOptions(contract: string) {
+    const [contractAddress, contractName] = contract.split('.');
```
kyranjamie: Maybe we should add some error handling here incase an erroneous string is added?

packages/stacking/src/index.ts
```diff
@@ -341,6 +519,144 @@ export class StackingClient {
...
+  }) {
+    const { hashMode, data } = decodeBtcAddress(poxAddress);
+    const hashModeBuffer = bufferCV(new BN(hashMode, 10).toBuffer());
+    const hashbytes = bufferCV(data);
+    const address = tupleCV({
+      hashbytes,
+      version: hashModeBuffer,
+    });
+
+    const [contractAddress, contractName] = contract.split('.');
```
kyranjamie: Maybe a new fn can be made to abstract out contract name parsing, something like:


```typescript
function parseContrastName (contract: string) {
  if (stringNotValid)  {
    throw new Error('Unable to parse string error');
  }
  return [address, name] = contract.split('.'); 
}
``

File: packages/stacking/src/index.ts (hunk 7 above)
Comment (hstove) [reply packages/stacking/src/index.ts]: No need to `return await ...`, just `return broadcastTransaction(...`. TS will infer it correctly. I think ESLint also doesn't like the `return await` style

File: packages/stacking/src/index.ts (hunk 12 above)
Comment (hstove) [reply packages/stacking/src/index.ts]: You can `import { validateStacksAddress } from '@stacks/transactions'` and use it here. 

We should just have a `validateContractIdentifier` function if we don't already. But don't consider that blocking for this PR.

File: packages/stacking/src/index.ts (hunk 7 above)
Comment (kyranjamie) [reply packages/stacking/src/index.ts]: No need for `await` if returning promise

File: packages/stacking/README.md [hunk 13]
```diff
@@ -145,3 +160,95 @@ stackingStatus:
...
+
+#### Stack delegated STX
+
+```typescript
+// delegators would initiate a different client
+const delegatorAddress = 'ST22X605P0QX2BJC3NXEENXDPFCNJPHE02DTX5V74';
+// delegator private key for transaction signing
+const delegatorPrivateKey = 'd48f215481c16cbe6426f8e557df9b78895661971d71735126545abddcd5377001';
+// the BTC address for reward payouts; either to the delegator or to the BTC address set by the account holder
+const delegatePoxAddress = 'msiYwJCvXEzjgq6hDwD9ueBka6MTfN962Z';
```
hstove: We need to update the comment here that this is only the delegator's BTC address. We should also include that this is optional. I would love to see some link or comment about how this works, since it's been prone to confusion. IMO, this should be omitted by default. If we do keep it, the variable should be renamed to `delegatorPoxAddress`

packages/stacking/README.md
```diff
@@ -145,3 +160,95 @@ stackingStatus:
...
+```
+
+#### Commit to stacking
+
+```typescript
+// reward cycle id to commit to
+const rewardCycle = 12;
+
+const delegetateCommitResponse = await delegatorClient.stackAggregationCommit({
+  poxAddress: delegatePoxAddress,
```
hstove: rename to `delegatorPoxAddress`

packages/stacking/src/index.ts
```diff
@@ -103,6 +105,49 @@ export interface LockStxOptions {
...
+  poxAddress?: string;
+  privateKey: string;
+}
+
+/**
+ * Delegate stack stx options
+ *
+ * @param  {String} stacker - the STX address of the delegator
+ * @param  {BigNum} amountMicroStx - number of microstacks to lock
+ * @param  {String} poxAddress - the reward Bitcoin address
```
hstove: Please update to "the reward Bitcoin address of the delegator"

packages/stacking/src/index.ts
```diff
@@ -103,6 +105,49 @@ export interface LockStxOptions {
...
 }
 
+/**
+ * Delegate stx options
+ *
+ * @param  {BigNum} amountMicroStx - number of microstacks to delegate
+ * @param  {String} delegateTo - the STX address of the delegatee
+ * @param  {number} untilBurnBlockHeight - the burnchain block height after which delegation is revoked
+ * @param  {String} poxAddress - the reward Bitcoin address of the delegator
+ * @param  {String} privateKey - private key to sign transaction
```
hstove: FYI - you don't need to include the types in brackets, because TSDoc will generate them automatically. It leads to errors, too. For example, the type of `amountMicroStx` is `BN`, not `BigNum`. The type of `poxAddress` is `string | undefined`. I recommend just removing the explicit types.

hstove (changes requested):
Mainly requesting changes around comments and naming. I recommend removing all hard-coded types in any TSDoc comments. They will be generated automatically from the Typescript types.

packages/stacking/src/index.ts
agraebe: sure

packages/stacking/src/index.ts
agraebe: 💯

packages/stacking/src/index.ts
agraebe: sure thing, changing!

packages/stacking/src/index.ts
agraebe: done as well

packages/stacking/src/index.ts
agraebe: edit - answering in the comment below!

packages/stacking/src/index.ts
agraebe: using these suggestions for a new method `parseContractId`. please review again 🙏🏼

packages/stacking/README.md
agraebe: I'll remove it for now. and the optional hint will be added too

agraebe: btw, it is mentioned to be optional where the user initiates the delegation relationship

packages/stacking/README.md
agraebe: done

packages/stacking/src/index.ts
agraebe: done as well

hstove (approved):

PR #904, opened 2021-01-13
fix: check okay before parsing result

### LINKED ISSUE #894: TypeError: Cannot read property 'startsWith' of undefined
**Describe the bug**
Following tutorial at https://docs.blockstack.org/understand-stacks/stacking-using-CLI, I tried to experiment with stacking.

Initial request to stacks node api succeeded, I have performed following request
`curl -X POST "https://stacks-node-api.xenon.blockstack.org/extended/v1/faucets/stx?address=xxx&stacking=true"` and got `status: success` and `txId: 0x6de9457b76756975c640ff6bdb3fe4229aece2fe480c60b0fed0600bc3c3fbe1`

However after installing `@stacks/cli@1.0.0-beta.20` and issuing following command `stx can_stack 85255287736277 200 mpiSSte8C8tMvPTjE7bgCwKL6Z27yRJ7io ST1JEADA2C3B9VM005YRV1PTD28EB3506976W0ERN -t` I am getting an error.

`@stacks/cli@1.0.0-beta.20`

Bug happens when I try to check if following address can stack, it happens on Blockstack CLI

**To Reproduce**
1. Install latest `cli`
2. Try invoking following command `stx can_stack 85255287736277 200 mpiSSte8C8tMvPTjE7bgCwKL6Z27yRJ7io ST1JEADA2C3B9VM005YRV1PTD28EB3506976W0ERN -t`

**Expected behavior**
Blockstack CLI should not crash.

**Console log**
```
(torch) ~/D/r/k ❯❯❯ stx can_stack 85255287736277 200 mpiSSte8C8tMvPTjE7bgCwKL6Z27yRJ7io ST1JEADA2C3B9VM005YRV1PTD28EB3506976W0ERN -t
TypeError: Cannot read property 'startsWith' of undefined
    at hexToCV (/Users/pawel/.nvm/versions/node/v14.15.0/lib/node_modules/@stacks/cli/node_modules/@stacks/transactions/src/utils.ts:165:32)
    at parseReadOnlyResponse (/Users/pawel/.nvm/versions/node/v14.15.0/lib/node_modules/@stacks/cli/node_modules/@stacks/transactions/src/utils.ts:186:10)
    at /Users/pawel/.nvm/versions/node/v14.15.0/lib/node_modules/@stacks/cli/node_modules/@stacks/transactions/src/builders.ts:1074:47
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async Promise.all (index 2)
```

 - OS: OSX

friedger:
## Description
if a read only function fails in `callReadOnlyFunction` then the result is a `TypeError`.
 
This PR 
* adds a check in `parseReadOnlyResponse` whether the response is `okay`
* fixes #894 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

## Checklist

### DIFF
--- a/packages/transactions/src/utils.ts
+++ b/packages/transactions/src/utils.ts
@@ -1,5 +1,5 @@
 import { sha256, sha512 } from 'sha.js';
-import { ClarityValue, noneCV, responseErrorCV, serializeCV } from './clarity';
+import { ClarityValue, serializeCV } from './clarity';
 import RIPEMD160 from 'ripemd160-min';
 import randombytes from 'randombytes';
 import { deserializeCV } from './clarity';
--- a/packages/transactions/src/utils.ts
+++ b/packages/transactions/src/utils.ts
@@ -173,20 +173,29 @@ export function hexToCV(hex: string) {
  * @param {string} result - serialized hex clarity value
  */
 
-export interface ReadOnlyFunctionResponse {
-  okay: boolean;
+export interface ReadOnlyFunctionSuccessResponse {
+  okay: true;
   result: string;
 }
+
+export interface ReadOnlyFunctionErrorResponse {
+  okay: false;
+  cause: string;
+}
+
+export type ReadOnlyFunctionResponse =
+  | ReadOnlyFunctionSuccessResponse
+  | ReadOnlyFunctionErrorResponse;
 
 /**
  * Converts the response of a read-only function call into its Clarity Value
  * @param param
  */
-export const parseReadOnlyResponse = ({ okay, result }: ReadOnlyFunctionResponse): ClarityValue => {
-  if (okay) {
-    return hexToCV(result);
+export const parseReadOnlyResponse = (response: ReadOnlyFunctionResponse): ClarityValue => {
+  if (response.okay) {
+    return hexToCV(response.result);
   } else {
-    return responseErrorCV(noneCV());
+    throw new Error(response.cause);
   }
 };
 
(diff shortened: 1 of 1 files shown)


packages/transactions/src/utils.ts
```diff
@@ -182,8 +182,12 @@ export interface ReadOnlyFunctionResponse {
  * Converts the response of a read-only function call into its Clarity Value
  * @param param
  */
-export const parseReadOnlyResponse = ({ result }: ReadOnlyFunctionResponse): ClarityValue => {
-  return hexToCV(result);
+export const parseReadOnlyResponse = ({ okay, result }: ReadOnlyFunctionResponse): ClarityValue => {
+  if (okay) {
+    return hexToCV(result);
+  } else {
+    return responseErrorCV(noneCV());
```
hstove: I don't think this is right - shouldn't we parse the error and return that?

hstove (changes requested):

packages/transactions/src/utils.ts
hstove: Should be:

```ts
export interface ReadOnlyFunctionSuccessResponse {
  okay: true;
  result: string;
}

export interface ReadOnlyFunctionErrorResponse {
  okay: false;
  cause: string;
}

export type ReadOnlyFunctionResponse = ReadOnlyFunctionSuccessResponse | ReadOnlyFunctionErrorResponse;

/**
 * Converts the response of a read-only function call into its Clarity Value
 * @param param
 */
export const parseReadOnlyResponse = (response: ReadOnlyFunctionResponse): ClarityValue => {
  if (response.okay) {
    return hexToCV(response.result);
  } else {
    throw new Error(response.cause);
  }
};
```

friedger: Makes sense!

hstove (approved):

PR #909, opened 2021-01-14
fix: cli - update urls for stacks 2

### LINKED ISSUE #908: CLI: Stacks 1 endpoints used for balance and send_tokens methods
Currently, the old `*core*` URL from Stacks 1 is used. We need to update it to broadcast to `http://stacks-node-api.stacks.co`

agraebe:
should resolve #908


zone117x (approved):

yknl (approved):
LGTM

PR #910, opened 2021-01-15
feat: v2 stacks transactions in blockstack.js

yknl:
This adds bitcoin chain based v2 stacks transactions to the legacy blockstack.js operations.


kantai (approved):
This looks good to me -- the only caveat I would add is that the token transfer has to spend the specific UTXO produced by pre-stx-op. I think this tx building will usually work, but depending on how the address gets funded, it can sometimes fail.

agraebe:
@yknl what is the next step here?

yknl:
Closing as this is already released and should live in it's own branch.

PR #917, opened 2021-01-20
feat(cli): gen keychain with custom path

### LINKED ISSUE #918: Support custom derivation paths with CLI
Hiro DevOps has two addresses which were created on the V1 network before the V2 mainnet launch using the old `blockstack-cli`, and had some BTC and STX sent to these addresses. Because of this, they were created on the old derivation path which is not supported on the Stacks Wallet.

How can someone access the funds on old addresses created using the old derivation path through any means?

kyranjamie:
@CharlieC3 we need something like this to support custom derivation paths

Ideally the cli could use some arg parsing library than arg indexes, I guess this came from the previous version.

I have not tested this code. Wanted to make as a quick placeholder PR for solving issue mentioned here https://github.com/blockstack/stacks-wallet/issues/544

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -339,7 +339,12 @@ async function makeKeychain(network: CLINetworkAdapter, args: string[]): Promise
     );
   }
 
-  const stacksKeyInfo = await getStacksWalletKeyInfo(network, mnemonic);
+  let derivationPath: string | undefined;
+  if (arg[1]) {
+    derivationPath = arg[1]
+  }
+
+  const stacksKeyInfo = await getStacksWalletKeyInfo(network, mnemonic, derivationPath);
   return JSONStringify({
     mnemonic: mnemonic,
     keyInfo: stacksKeyInfo,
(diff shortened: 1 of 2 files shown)


agraebe:
@kyranjamie what would be the next step here?

kyranjamie:
@agraebe To be honest, I made this PR to be as helpful as possible, while falling just short of volunteering to do it myself 😅 . 

It needs to be finished and tested. I can pick it up if needed, or someone else can. Is it an immediate priority?

zone117x:
A unit test for this would be helpful, add to the extensive cli unit tests here 😄 
https://github.com/blockstack/stacks.js/blob/5f4d2517efd080261879fd6df33b2cfb7524d65f/packages/cli/tests/cli.test.ts#L1-L3

agraebe:
@kyranjamie to your point of priority - I believe we settled on this being low priority in the issue discussion.

I simply would like to avoid stale PRs. If we have a low-effort way to completing this, we should. If it is more involved, I'd suggest we close this PR and get to the issue once it becomes more relevant.

kyranjamie:
Alright, let's close it out. 

I'm sure it's doable without huge efforts, but I think the whole development cycle, start through release, is move involved than to just a quick fix now.

PR #927, opened 2021-01-22
fix: name lookup url not working on mainnet, fixes #916

### LINKED ISSUE #916: `verifyAuthRequest` name lookup URL is on mainnet
Although mainnet is live, BNS endpoints are not deployed to mainnet yet. So, if an app uses the latest `@stacks/auth` version, and a user logs in with a username, then `verifyAuthRequest` will fail (this is called by default in all auth flows).

Options to fix temporarily:

- Apps can specify `appConfig.coreNode` to `https://core.blockstack.org` - this is a decent fix for as long as we don't have BNS endpoints or a new subdomain registrar
- Update `@stacks/auth` default URL to use `https://core.blockstack.org` - and we can revert that once BNS is live

We should be careful to test this - the API is not currently 1-for-1, for example, `/v1/names/:name` returns a stacks address (it used to be a BTC address), so there might be slight changes needed to support this change. Part of that might need to be on the authenticator, and some might be in `@stacks/auth` itself.

hstove:
Fixes #916 - name lookups on mainnet are not live yet. This is a temporary fix so that apps using the latest `@stacks/auth` don't break when logging in users with a username.


yknl (approved):
LGTM

PR #933, opened 2021-01-30
Fix: export PostCondition serialization functions

hstove:
What this does:

- export `deserializePostCondition` and `serializePostCondition`
  - needed for transaction signing in Connect. I published an `alpha` from here to resolve a hotfix
- Adds `lodash-es` dep to `@stacks/transactions`. Otherwise I see errors related to it when importing the package, and this has been reported by others too. Not 100% sure what the deal is - something with `tsdx`
- Updates all package versions to be in line with `lerna`

@reedrosenbluth since this is a lerna repo, we should be versioning and packaging with [`yarn lerna publish`](https://github.com/lerna/lerna/tree/main/commands/publish). This accepts different args like `patch`, etc. Otherwise, our package versions get out of sync, which can throw things off.

It's been a TODO of mine to add automatic semantic versioning to this repo, which is easier now that we moved CI to Github Actions.


kyranjamie:
+1 for needing semantic versioning

reedrosenbluth:
@hstove sounds good! I've been using `lerna publish` but was unaware of the args.

reedrosenbluth (approved):

Issue #893, opened 2021-01-06
Support for attachments


lgalabru:
When posting a transaction a blob of data (called attachment) can be attached to the transaction. 
So the post transaction call has been updated and can now **also** receive some `application/json` content.
The following body would be expected:

```rust
#[derive(Serialize, Deserialize)]
pub struct PostTransactionRequestBody {
    pub tx: String,
    pub attachment: Option<String>,
}
```

where tx and attachment are both hex encoded strings.

Relevant Rust code, handling that call:

```rust
    fn parse_posttransaction_octets<R: Read>(
        preamble: &HttpRequestPreamble,
        fd: &mut R,
    ) -> Result<HttpRequestType, net_error> {
        let tx = StacksTransaction::consensus_deserialize(fd).map_err(|e| {
            if let net_error::DeserializeError(msg) = e {
                net_error::ClientError(ClientError::Message(format!(
                    "Failed to deserialize posted transaction: {}",
                    msg
                )))
            } else {
                e
            }
        })?;
        Ok(HttpRequestType::PostTransaction(
            HttpRequestMetadata::from_preamble(preamble),
            tx,
            None,
        ))
    }

    fn parse_posttransaction_json<R: Read>(
        preamble: &HttpRequestPreamble,
        fd: &mut R,
    ) -> Result<HttpRequestType, net_error> {
        let body: PostTransactionRequestBody = serde_json::from_reader(fd)
            .map_err(|_e| net_error::DeserializeError("Failed to parse body".into()))?;

        let tx = {
            let tx_bytes = hex_bytes(&body.tx)
                .map_err(|_e| net_error::DeserializeError("Failed to parse tx".into()))?;
            StacksTransaction::consensus_deserialize(&mut &tx_bytes[..]).map_err(|e| {
                if let net_error::DeserializeError(msg) = e {
                    net_error::ClientError(ClientError::Message(format!(
                        "Failed to deserialize posted transaction: {}",
                        msg
                    )))
                } else {
                    e
                }
            })
        }?;

        let attachment = match body.attachment {
            None => None,
            Some(attachment_content) => {
                let content = hex_bytes(&attachment_content).map_err(|_e| {
                    net_error::DeserializeError("Failed to parse attachment".into())
                })?;
                Some(Attachment::new(content))
            }
        };

        Ok(HttpRequestType::PostTransaction(
            HttpRequestMetadata::from_preamble(preamble),
            tx,
            attachment,
        ))
    }
```

PR #948, opened 2021-02-12
Include attachment with transaction broadcast

reedrosenbluth:
Enable attachments to be included when broadcasting transactions

Issue: #893

### DIFF
--- a/packages/transactions/src/builders.ts
+++ b/packages/transactions/src/builders.ts
@@ -157,12 +157,13 @@ export type TxBroadcastResult = TxBroadcastResultOk | TxBroadcastResultRejected;
  */
 export async function broadcastTransaction(
   transaction: StacksTransaction,
-  network: StacksNetwork
+  network: StacksNetwork,
+  attachment?: Buffer
 ): Promise<TxBroadcastResult> {
   const rawTx = transaction.serialize();
   const url = network.getBroadcastApiUrl();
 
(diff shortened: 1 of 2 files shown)


packages/transactions/src/builders.ts
```diff
@@ -175,18 +178,32 @@ export async function broadcastTransaction(
...
-
-  const options = {
+  let options = {
     method: 'POST',
-    headers: requestHeaders,
-    body: rawTx,
   };
 
+  if (attachment) {
+    options = _.assign(options, {
```
kyranjamie: Could you just use the native `Object.assign`?

reedrosenbluth: Yup! I think lodash uses that under the hood if it’s available. I got into the habit of using lodash for this at a previous gig because IE still doesn’t support `Object.assign`, and we needed to support it. Will make this change as I don’t think we’re too concerned with supporting IE

lgalabru (approved):
Tested end to end, looks good to me.
Thank you @reedrosenbluth!

packages/transactions/src/builders.ts
kyranjamie: I'm not particularly opposed to lodash, but think this imports the entire bundle, versus `import assign from 'lodash-es/assign'`.

Could also use a spread operator.

kyranjamie (approved):
LGTM

PR #949, opened 2021-02-14
fix: don't multiply with cycleDuration (#937)

### LINKED ISSUE #937: StackingClient.getSecondsUntilNextCycle does not output seconds in TestNet
**Describe the bug**
Running `StackingClient.getSecondsUntilNextCycle` returns a value that the developer would expect to be seconds.
An example output would be `391608000`, which appropximates to 4,500 days.

@stacks/network@1.0.0-beta.20
@stacks/stacking@1.0.0-beta.20
@stacks/transactions@1.0.0-beta.20

No

**To Reproduce**
Run the following:

```
import { StackingClient } from '@stacks/stacking';
import {
    StacksTestnet,
    StacksMainnet
} from '@stacks/network';

const stxAddress = "SP....";

const client = new StackingClient(stxAddress, new StacksTestnet());
const secondsUntilNextCycle = await client.getSecondsUntilNextCycle();

console.log("SEC TIL NEXT CYCLE")
console.log(secondsUntilNextCycle)
```
**Console log**
```
SEC TIL NEXT CYCLE
390348000
```

**Expected behavior**
I would expect something on the order of 10-15 days, given that the next cycle should be on the 2/15/21. This issue was created in 2/1/21.

 - OS: MacOS and Ubuntu

friedger:
## Description

This PR
* corrects the calculation of seconds until next cycle
* fixes #937 
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

Yes, the result of StackingClient.getSecondsUntilNextCycle returns different values for the same state of the blockchain.

No

## Testing information
Unit test has been added

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/stacking/src/index.ts
+++ b/packages/stacking/src/index.ts
@@ -244,8 +244,7 @@ export class StackingClient {
           poxInfo.reward_cycle_length -
           ((coreInfo.burn_block_height - poxInfo.first_burnchain_block_height) %
             poxInfo.reward_cycle_length);
-        const cycleDuration = poxInfo.reward_cycle_length * targetBlockTime;
-        return blocksToNextCycle * cycleDuration;
+        return blocksToNextCycle * targetBlockTime;
       }
     );
   }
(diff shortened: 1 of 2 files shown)


kyranjamie (approved):
Thanks @friedger

PR #951, opened 2021-02-16
Refactor: transaction broadcast request options construction

reedrosenbluth:
Small PR to refactor the way that the transaction broadcast request's options are constructed.

Also includes a change to export the `cvToValue` function.


kyranjamie (approved):

yknl:
We should follow the PR templates and start using the changelog since there are many 3rd party integrators of this library now. Any changes breaking or not need to be documented so not to impact downstream apps.

PR #957, opened 2021-02-24
fix: delegateStx invalid argument error

### LINKED ISSUE #955: Delegation: delegate-stx burn block height settings seems wrong
The follow method results in an error:
```
  // how much to stack, in microSTX
  const amountMicroStx = new BN(100000000000);

  // burn height at which the delegation relationship should be revoked (optional)
  const untilBurnBlockHeight = 5000;

  const delegetateResponse = await clientDelegate.delegateStx({
    amountMicroStx,
    delegateTo: <stxAddressDelegate>,
    untilBurnBlockHeight,
    privateKey: <privateKeyDelegate>,
  });
```

The error returned is:
```
node:15924) UnhandledPromiseRejectionWarning: Error: Clarity function `delegate-stx` expects argument 3 to be of type (optional uint), not uint
```

The method `getDelegateOptions` sets the value as such:
```
untilBurnBlockHeight ? uintCV(untilBurnBlockHeight) : noneCV(),
```

I assume the fix would be wrapping the `uint` into `some`:
```
untilBurnBlockHeight ? someCV(uintCV(untilBurnBlockHeight)) : noneCV(),
```

reedrosenbluth:
contract-call for `delegate-stx` should be passed an `(optional uint)` for `until-burn-ht` arg

issue: #955

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - export `cvToValue` function
+- fix `delegateStx` function
 
 ## [21.1.1] - 2020-08-24
 ### Changed
--- a/packages/stacking/src/index.ts
+++ b/packages/stacking/src/index.ts
@@ -378,7 +378,7 @@ export class StackingClient {
       untilBurnBlockHeight,
       poxAddress,
     });
-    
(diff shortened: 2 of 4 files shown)


agraebe (approved):
thanks very much 🙏🏼

reedrosenbluth:
Not sure why lint is failing... Prettier check passes locally for me

PR #958, opened 2021-02-24
Refactor: cleanup prettier dependency

reedrosenbluth:
- `@blockstack/prettier-config` -> `@stacks/prettier-config`
- remove prettier dependency from individual packages in favor of root dependency


kyranjamie (approved):
Nice one @reedrosenbluth

PR #968, opened 2021-03-10
Fix toBuffer() to toArrayLike(Buffer))

FhilF:
### Description
When using the method `.delegateStackStx` and `.getDelegateStackOptions` I get an error `Assertion Failed`

After some debugging, it looks like the error is coming from this call
https://github.com/blockstack/stacks.js/blob/a6fab4e5c3038cf4eaa7bd997b7d4c5104ca6858/packages/stacking/src/index.ts#L530
https://github.com/blockstack/stacks.js/blob/a6fab4e5c3038cf4eaa7bd997b7d4c5104ca6858/packages/stacking/src/index.ts#L576

hashMode returned by const { hashMode, data } = decodeBtcAddress(poxAddress); is 0 and calling bn.js new BN(hashMode, 10).toBuffer() to get the buffer on 0 is throwing an error

**Related issue/pull-request:**
https://github.com/blockstack/stacks.js/pull/925

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] API reference/documentation update
- [ ] Other

## Does this introduce a breaking change?
No

## Are documentation updates required?
No

## Testing information

Provide context on how tests should be performed.

1. Is testing required for this change?
The related pull request manages to fix the error.

2. If it’s a bug fix, list steps to reproduce the bug
```
const network = new StacksTestnet();
const delegateeClient = new StackingClient(delegateeAddress, network);

getNonce(delegateeAddress, network)
      .then((res) => {
        return res.add(new BN(1));
      })
      .then((res) => {
        return delegateeClient.delegateStackStx({
          stacker: delegatorAddress,
          amountMicroStx: amountMicroStx,
          poxAddress: delegateeBtcAddress,
          burnBlockHeight: burnBlockHeight,
          cycles: cycles,
          privateKey: delegateePrivateKey,
          nonce: res, // optional
        });
      })
      .then((res) => {
        console.log(res);
      })
      .catch((err) => {
        console.log(err);
      });
```

and

```
const network = new StacksTestnet();
const delegateeClient = new StackingClient(delegateeAddress, network);

delegateeClient.getDelegateStackOptions({
        contract:"ST000000000000000000002AMW42H.pox",
        stacker: delegatorAddress,
        amountMicroStx: new BN(1000),
        poxAddress: delegateeBtcAddress,
        burnBlockHeight: 100,
        cycles: 1,
      }).then((res) => {
        console.log(res);
      })
      .catch((err) => {
        console.log(err);
      });
```

## Checklist
[...]

PR #970, opened 2021-03-11
Update toBuffer() to toArrayLike(Buffer)

FhilF:
### Description
When using the method `.stackAggregationCommit` I get an error `Assertion Failed`

After some debugging, it looks like the error is coming from this call
https://github.com/blockstack/stacks.js/blob/master/packages/stacking/src/index.ts#L617
https://github.com/blockstack/stacks.js/blob/master/packages/stacking/src/index.ts#L711

hashMode returned by const { hashMode, data } = decodeBtcAddress(poxAddress); is 0 and calling bn.js new BN(hashMode, 10).toBuffer() to get the buffer on 0 is throwing an error

**Related issue/pull-request:**
https://github.com/blockstack/stacks.js/pull/925
https://github.com/blockstack/stacks.js/pull/968

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

No

## Testing information

1. Is testing required for this change?
The related pull request manages to fix the error.

2. If it’s a bug fix, list steps to reproduce the bug

```
const rewardCycle = 12;
const delegateeClient = new StackingClient(delegateeAddress, network);

    delegateeClient
      .stackAggregationCommit({
        poxAddress: delegateeBtcAddress,
        rewardCycle,
        privateKey: delegateePrivateKey,
      })
      .then((res) => {
        console.log(res);
      })
      .catch((err) => {
        console.log(err);
      });
```

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/stacking/src/index.ts
+++ b/packages/stacking/src/index.ts
@@ -614,7 +614,7 @@ export class StackingClient {
     rewardCycle: number;
   }) {
     const { hashMode, data } = decodeBtcAddress(poxAddress);
-    const hashModeBuffer = bufferCV(new BN(hashMode, 10).toBuffer());
+    const hashModeBuffer = bufferCV(new BN(hashMode, 10).toArrayLike(Buffer));
     const hashbytes = bufferCV(data);
     const address = tupleCV({
       hashbytes,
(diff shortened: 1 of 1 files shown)


zone117x (approved):
LGTM!

PR #978, opened 2021-03-19
Feat: BNS package

### LINKED ISSUE #864: Create library for BNS 2.0
Create a library for working with BNS 2.0.

reedrosenbluth:
This PR includes a new `@stacks/bns` package. It is an extension of some work that @yknl started a little while back.

It is a client library for the `bns` contract, with functions corresponding to the methods in the contract.

This PR also includes `typedoc` generated docs for the package. Ideally, we should have docs generated for the entire monorepo, but the root typedoc config is a bit broken right now (#996).


packages/bns/README.md
```diff
@@ -0,0 +1,28 @@
...
+## Installation
+```
+npm install --save @stacks/bns
+```
+
+## Example Usage
+
+```
+import { canRegisterName } = from '@stacks/bns';
+import { StacksTestnet } = from '@stacks/network';
```
kyranjamie:
```suggestion
import { canRegisterName } from '@stacks/bns';
import { StacksTestnet } from '@stacks/network';
```

packages/bns/docs/README.md
```diff
@@ -0,0 +1,47 @@
...
+npm install --save @stacks/bns
+```
+
+## Usage
+
+### `can-register-name`
+
+```
+import { canRegisterName } = from '@stacks/bns';
+import { StacksTestnet } = from '@stacks/network';
```
kyranjamie:
```suggestion
import { canRegisterName } from '@stacks/bns';
import { StacksTestnet } from '@stacks/network';
```

packages/bns/docs/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+const network = new StacksMainnet();
+
+const result = await canRegisterName('name.id', network);
+```
+
+### `get-namespace-price`
+```
+import { get } = from '@stacks/bns';
```
lgalabru:
```suggestion
import { canRegisterName } from '@stacks/bns';
```

packages/bns/docs/README.md
```diff
@@ -0,0 +1,47 @@
...
+import { StacksTestnet } = from '@stacks/network';
+
+const network = new StacksMainnet();
+
+const result = await canRegisterName('name.id', network);
+```
+
+### `get-name-price`
+```
+import { get } = from '@stacks/bns';
```
lgalabru:
```suggestion
import { canRegisterName } from '@stacks/bns';
```

lgalabru (commented):
Thank you for taking on this @reedrosenbluth!
I have some concerns to share on the overall API. I could be missing something bur per my understanding, all the functions here are requiring a private key as an input, which is not a pattern we should encourage IMO. I think that we should be assuming that this APIs will be used by dapps developers, that don't have access to the private and instead interacting with wallet / ledgers. In that sense, functions in that library should be generating unsigned transactions that can be then submitted to wallet / ledgers for signature.
Does that make sense?

reedrosenbluth:
@lgalabru totally makes sense. I was under the impression that this library would be more useful for people writing scripts to interact with BNS, and dapp developers would be using the `@stacks/transactions` library directly.

Happy to remove the private key requirement and the transaction broadcasting functionality though.

yknl:
I think @lgalabru's suggestion is right. This library should be able to generate unsigned transactions that can be passed to either an authenticator or custom signing infrastructure in wallet implementations. There's no reason to limit the library's use to scripts.

From Xverse wallet's perspective, we have our own generic transaction signing code that can take the generated unsigned tx, sign and broadcast it. I'm not sure what kind of input makes most sense for authenticators yet.

reedrosenbluth:
@lgalabru @yknl just made those changes. will push updated unit tests shortly

packages/bns/src/index.ts
```diff
@@ -0,0 +1,764 @@
...
+ * First step in registering a namespace. This transaction does not reveal the namespace that is
+ * about to be registered. And it sets the amount of STX to be burned for the registration.
+ *
+ * Resolves to the generated StacksTransaction
+ *
+ * @param  {PreorderNamespaceOptions} options - an options object for the preorder
+ *
+ * @return {Promise<StacksTransaction>}
+ */
+export async function buildPreorderNamespaceTX({
```
kyranjamie: Why uppercase X, it's not an acronym.

```suggestion
export async function buildPreorderNamespaceTx({
```

packages/bns/src/index.ts
```diff
@@ -0,0 +1,764 @@
...
+ * Generates a namespace reveal transaction.
+ * Second step in registering a namespace.
+ *
+ * Resolves to the generated StacksTransaction
+ *
+ * @param  {RevealNamespaceOptions} options - an options object for the reveal
+ *
+ * @return {Promise<StacksTransaction>}
+ */
+export async function buildRevealNamespaceTX({
```
kyranjamie:
```suggestion
export async function buildRevealNamespaceTx({
```

packages/bns/src/index.ts
```diff
@@ -0,0 +1,764 @@
...
+  namespace,
+  salt,
+  priceFunction,
+  lifetime,
+  namespaceImportAddress,
+  publicKey,
+  network,
+}: RevealNamespaceOptions): Promise<StacksTransaction> {
+  const bnsFunctionName = 'namespace-reveal';
+  const txNetwork = network || new StacksMainnet();
```
kyranjamie: Is setting defaults like this a safe pattern? IMO it'd better to require this to be set explicitly.

packages/bns/src/index.ts
```diff
@@ -0,0 +1,764 @@
...
+ * Generates a namespace name import transaction.
+ * An optional step in namespace registration.
+ *
+ * Resolves to the generated StacksTransaction
+ *
+ * @param  {ImportNameOptions} options - an options object for the name import
+ *
+ * @return {Promise<StacksTransaction>}
+ */
+export async function buildImportNameTX({
```
kyranjamie:
```suggestion
export async function buildImportNameTx({
```

packages/bns/src/index.ts
```diff
@@ -0,0 +1,764 @@
...
+ * Final step in namespace registration. This completes the namespace registration and
+ * makes the namespace available for name registrations.
+ *
+ * Resolves to the generated StacksTransaction
+ *
+ * @param  {ReadyNamespaceOptions} options - an options object for the namespace ready transaction
+ *
+ * @return {Promise<StacksTransaction>}
+ */
+export async function buildReadyNamespaceTX({
```
kyranjamie:
```suggestion
export async function buildReadyNamespaceTx({
```

packages/bns/src/index.ts
```diff
@@ -0,0 +1,764 @@
...
+  b10: BN;
+  b11: BN;
+  b12: BN;
+  b13: BN;
+  b14: BN;
+  b15: BN;
+  b16: BN;
+  nonAlphaDiscount: BN;
+  noVowelDiscount: BN;
+};
```
kyranjamie: Interface?

kyranjamie (commented):
Looks good, most feedback naming related.

Imo BNS and other 3LAs in camelcase should be considered a word `makeBnsContractCall`.

packages/bns/src/index.ts
```diff
@@ -0,0 +1,748 @@
...
+  UnsignedContractCallOptions,
+} from '@stacks/transactions';
+
+import { StacksNetwork } from '@stacks/network';
+
+import { bufferCVFromString, decodeFQN, getZonefileHash, uintCVFromBN } from './utils';
+
+import BN from 'bn.js';
+
+export const BNS_CONTRACT_ADDRESS = 'ST000000000000000000002AMW42H';
```
zone117x: Should this have both testnet and mainnet addresses?
```ts
export const enum BnsContracAddress {
  mainnet = 'SP000000000000000000002Q6VF78',
  testnet = 'ST000000000000000000002AMW42H',
}
```

packages/bns/src/index.ts
```diff
@@ -0,0 +1,748 @@
...
+  functionName: string;
+  functionArgs: ClarityValue[];
+  publicKey: string;
+  network: StacksNetwork;
+  attachment?: Buffer;
+}
+
+async function makeBnsContractCall(options: BnsContractCallOptions): Promise<StacksTransaction> {
+  const txOptions: UnsignedContractCallOptions = {
+    contractAddress: BNS_CONTRACT_ADDRESS,
```
zone117x: Then I suppose this would be something like
```ts
contractAddress: options.network.chainId === ChainID.Mainnet ? BnsContractAddress.mainnet ? BnsContractAddress.testnet
```
Except as a helper function since it's used in a couple places, and maybe check for an unexpected `chainId`, for like, non-TS devs passing it wild numbers

packages/bns/src/index.ts
```diff
@@ -0,0 +1,748 @@
...
+    network,
+  }).then((responseCV: ClarityValue) => {
+    if (responseCV.type === ClarityType.ResponseOk) {
+      if (responseCV.value.type === ClarityType.Int || responseCV.value.type === ClarityType.UInt) {
+        return responseCV.value.value;
+      } else {
+        throw new Error('Response did not contain a number');
+      }
+    } else {
+      const errorResponse = responseCV as ResponseErrorCV;
```
zone117x: Would be a bit safer/friendlier in these Clarity decodings to check for the `ResponseError` type here in an `else if (...)`, then have the `else` throw an error for unexpected Clarity value

packages/bns/src/index.ts
reedrosenbluth: ah yes! good catch

packages/bns/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+const network = new StacksMainnet();
+
+const name = 'name.id';
+const salt =  'example-salt'
+const zonefile =  'example-zonefile'
+const publicKey = 'SPF0324DSC4K505TP6A8C7GAK4R95E38TGNZP7RE'
+
+// construct an unsigned bns register-name transaction
+const unsignedTX = await buildRegisterNameTX(name, salt, zonefile, publicKey, network);
```
zone117x: Should be a single object for the function arg, rather than list of function args, right?

packages/bns/src/index.ts
```diff
@@ -0,0 +1,748 @@
...
+ *
+ * @param {string} fullyQualifiedName - the fully qualified name to check
+ * @param {StacksNetwork} network - the Stacks network to broadcast transaction to
+ *
+ * @returns {Promise} that resolves to true if the operation succeeds
+ */
+export async function canRegisterName(
+  fullyQualifiedName: string,
+  network: StacksNetwork
+): Promise<boolean> {
```
zone117x: What are you thoughts on using an object here for args? Makes it a bit cleaner if the function args need expanded in the future

reedrosenbluth: Not opposed. Do you feel this way about the other similar read only functions too?

zone117x: I think generally functions with two or more args should use an object, since the trend is for the amount of options to increase.

packages/bns/src/index.ts
```diff
@@ -21,9 +22,20 @@ import { bufferCVFromString, decodeFQN, getZonefileHash, uintCVFromBN } from './
...
 export const BNS_CONTRACT_NAME = 'bns';
 
+export const enum BnsContractAddress {
+  mainnet = 'SP000000000000000000002Q6VF78',
+  testnet = 'ST000000000000000000002AMW42H',
+}
+
+function getBnsContractAddress(network: StacksNetwork) {
+  if (network.chainId === ChainID.Mainnet) return BnsContractAddress.mainnet;
+  else return BnsContractAddress.testnet;
```
zone117x:
```suggestion
  else if (network.chainId == ChainID.Testnest) return BnsContractAddress.testnet;
  else throw new Error(`Unexpected ChainID: ${network.chainID}`);
```
?

zone117x (approved):
Looks great to me!

packages/bns/docs/modules.md
```diff
@@ -0,0 +1,385 @@
...
+Defined in: [index.ts:725](https://github.com/blockstack/stacks.js/blob/master/packages/bns/src/index.ts#L725)
+
+___
+
+### transferName
+
+▸ **transferName**(`__namedParameters`: [*TransferNameOptions*](interfaces/transfernameoptions.md)): *Promise*<[*Result*](modules.md#result)\>
+
+Generates and broadcasts a name transfer transaction.
+This changes the owner of the registered name.
```
lgalabru: Should we add a note, stipulating that the underlying NFT will be transferred, meaning a post condition is required?

lgalabru (approved):
LGTM, thank you @reedrosenbluth!

PR #984, opened 2021-03-23
feat: match authResponse username with 2.0 bns lookup

### LINKED ISSUE #945: Integrate Stacks Blockchain API for querying username data
We need to integrate the Stacks Blockchain API for querying username data via Stacks.js.

This is partially in support of https://github.com/blockstack/ux/issues/956.

### LINKED ISSUE #953: Update `@stacks/auth` to work with new BNS APIs
Once BNS APIs are deployed, we need to update `@stacks/auth` to use the new APIs to do username validation within the client application. This will require updating the URL, but also updating some of the logic to validate against an STX address, instead of a BTC address.

hstove:
Fixes #953 

This is the start of the work that will allow apps to handle an `authResponse` properly with 2.0 BNS.

After authentication, `@stacks/auth` calls `verifyAuthResponse`. This function runs a few checks, and also runs a check to ensure that the auth response is signed by the owner of the username provided in the payload.

Right now, you need to use a specific `AppConfig` to have a custom `coreNodeUrl`. When provided, it uses a different BNS lookup URL. For testing, I'm using our testnet registrar.

### DIFF
--- a/packages/auth/src/verification.ts
+++ b/packages/auth/src/verification.ts
@@ -3,6 +3,7 @@ import { getAddressFromDID } from './dids';
 import { publicKeyToAddress } from '@stacks/encryption';
 import { fetchPrivate, isSameOriginAbsoluteUrl } from '@stacks/common';
 import { fetchAppManifest } from './provider';
+import { c32ToB58 } from 'c32check';
 
 /**
  * Checks if the ES256k signature on passed `token` match the claimed public key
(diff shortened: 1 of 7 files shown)


markmhendrickson:
@hstove should someone review these code changes now?

packages/auth/src/verification.ts
```diff
@@ -259,21 +266,31 @@ export async function verifyAuthRequestAndLoadManifest(token: string): Promise<a
...
     doPublicKeysMatchIssuer(token),
-    doPublicKeysMatchUsername(token, nameLookupURL),
   ]);
-  return values.every(val => val);
+  const usernameMatchings = await Promise.all(
+    [nameLookupURL]
+      .concat(fallbackLookupURLs || [])
+      .map(url => doPublicKeysMatchUsername(token, url))
+  );
+  const someUsernameMatches = usernameMatchings.find(doesMatch => doesMatch);
```
reedrosenbluth: nit: use `includes` instead of `find` since it returns a boolean. This way you won't need to "cast" with `!!` on the line below.

```typescript
const someUsernameMatches = usernameMatchings.includes(true);
```

reedrosenbluth (approved):
LGTM! left one small style comment

andresgalante:
I this ready to merge?

hstove:
Yes, ready for merge + release, cc @reedrosenbluth

PR #991, opened 2021-03-30
feat(cli): add WIF format to make_keychain

### LINKED ISSUE #938: Add WIF format to `make_keychain` in the CLI
When you generate a keychain in the CLI, it uses the standard STX derivation path (which starts with `m/5757'/...`). If you try and use the secret key generated from the CLI in a Bitcoin wallet, you won't be able to access the address generated in the CLI. However, one easy fix is for the CLI to output your `WIF Private Key` (Wallet Import Format), which is supported by many BTC wallets.

To do so:
- Get the WIF key in [keys.ts](https://github.com/blockstack/stacks.js/blob/master/packages/cli/src/keys.ts#L139:L139)
  - `const wif = child.toWIF();`
- Add the WIF key to the result of that function
- Update the typescript return type of that function
- Add a test

marcosc90:
## Description

Add WIF format to make_keychain in the CLI

closes #938

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

## Testing information

`stx make_keychain` should output WIF private key

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/cli/jest.config.js
+++ b/packages/cli/jest.config.js
@@ -1,0 +1,16 @@
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+  coverageDirectory: './coverage/',
+  collectCoverage: true,
+  globals: {
+    'ts-jest': {
+      diagnostics: {
+        ignoreCodes: ['TS151001'],
+      },
+    },
+  },
+  moduleFileExtensions: ['js', 'ts', 'd.ts'],
+  setupFilesAfterEnv: ['./tests/setup.js'],
+  testTimeout: 10000,
+};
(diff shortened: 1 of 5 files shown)


marcosc90:
The [error in the pipeline](https://github.com/blockstack/stacks.js/pull/991/checks?check_run_id=2230165151#step:10:191) 

```
    src/cli.ts:1058:13 - error TS2794: Expected 1 arguments, but got 0. Did you forget to include 'void' in your type argument to 'Promise'?

    1058             resolve();
                     ~~~~~~~~~

      ../../node_modules/typescript/lib/lib.es2015.promise.d.ts:33:34
```
Is not related to this PR AFAIK,

It gets solved by changing:

https://github.com/blockstack/stacks.js/blob/5a54a78bf35541760bcb402bc078367a22c87848/packages/cli/src/cli.ts#L1030

to

```typescript
 function downloadFile(hubConfig: GaiaHubConfig, fileName: string): Promise<void> { 
```

Should probably be in another PR.

zone117x (approved):
LGTM, and thanks for getting tests for this package setup!

reedrosenbluth (approved):
LGTM!

PR #998, opened 2021-04-07
Add link to BNS package to readme

reedrosenbluth:


README.md
```diff
@@ -24,7 +25,6 @@ To migrate your app from blockstack.js to stacks.js follow the steps in the [mig
 To setup the development environment for this repository, follow these steps:
 
 1. Clone this package.
-2. Make sure you have both `lerna` and `yarn` installed globally
```
zone117x, line 27: These are both still required right?

reedrosenbluth: For the commands below to work properly, yes.

But we also have it listed as a dev-dependency, so it can be invoked via new package.json scripts or `npx lerna ... `

Will update the README to reflect this

marcosc90 (approved):
LGTM (after addressing @zone117x  comment)

kyranjamie (approved):

PR #1013, opened 2021-04-26
Localnet flag and config file fix

surfdude75:
Fix .blockstack-cli-testnet.conf config file loading.
It was not working on Linux.

Add -l option to use local network testnet.


reedrosenbluth (approved):
This looks mostly good to me! Thanks!

PR #1015, opened 2021-04-29
Fix/monorepo refactor

### LINKED ISSUE #896: Use ES2017 for node.js libs
The `@stacks/transactions` lib appears to be compiling to an unnecessarily old version of js. Pre-ES2017 without native async-await support. 

I'm running into a bug where a readonly contract call function is throwing `TypeError: Cannot read property 'startsWith' of undefined`, and debugging the issue is difficult because the stack trace is scrambled with regenerator-runtime spaghetti (async-await polyfill code). 

image

This is extra painful since source maps do not appear to be working correctly. For example, the error is thrown inside `parseReadOnlyResponse`, so stepping through to look at the calling function results in unreadable polyfill code:

image

ES2017 has been supported since around Nodejs v7. We should at least update the commonjs (node) libs. Even global browser share is at 95% support for ES2017+.

### LINKED ISSUE #975: [DX] No not require global lerna installation or yarn
The current development environment setup requires unnecessary global installations:
> Make sure you have both `lerna` and `yarn` installed globally

Lerna should be a devDependency of the root package.json. 

Required usage of yarn should be reconsidered. Npm is the canonical node package manager that everyone has installed. Yarn has subtle differences in behavior that can be frustrating to debug. Yarn can also be installed in several ways that do not play nicely with nvm, causing issues with nodejs version isolation, global packages, etc.

Some yarn related issue failing here: https://github.com/blockstack/stacks.js/pull/974/checks?check_run_id=2137992586
```
Run yarn --frozen-lockfile --ignore-scripts
  yarn --frozen-lockfile --ignore-scripts
  shell: /usr/bin/bash -e {0}
  env:
    NODE_OPTIONS: --max-old-space-size=4096
yarn install v1.22.5
[1/4] Resolving packages...
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Error: Process completed with exit code 1.
```

As a nodejs dev who doesn't use yarn, I need to go poke around to figure out why my PR just broke this even though the dependencies and lockfiles were not touched.

zone117x:
* Configure jest tsconfig path mapping. This fixes a development issue where changes to a package's sibling dependent wouldn't show up without a full mono repo rebuild. Also fixes the debugging issue where it wasn't possible to step into a package's dependency source.
* Setup debugger config for all the packages.
* Remove lots of unused dependencies.
* Use typescript project references for faster cjs and esm build times during development.
* Use the `buffer` npm lib rather than the NodeJS global `Buffer`. This allows the tsc output to directly work in NodeJS environments and modern browser environments, without needing to run the source through another preprocessing tool.
* Fix issue with missing source map files in published packages.
* Use lerna hoisting for faster bootstrap times.


zone117x:
@reedrosenbluth this is ready for review and beta package publishing.

reedrosenbluth (approved):
LGTM! Awesome refactor!

PR #1022, opened 2021-05-17
feat: remove default anchormode from tx builders

### LINKED ISSUE #990: Transactions: Remove default anchor mode
Generated transactions by default set the anchor mode to `any`. Our microblock discussions revealed that it would be better to not set any default. Instead, we should educate the developer (https://github.com/blockstack/docs/issues/1023) and make sure they get familiar with the trade-offs and implications.

Developers should be required to set the anchor mode in the transaction options.

**Additional context**
Because this is a breaking change with downstream dependencies for apps, we need to think about a rollout plan (deprecation notice, upgrade docs, etc)

reedrosenbluth:
This PR changes the behavior of the tx builder functions. They now require an anchormode to be provided in the `options` arg, and no longer fall back to a default anchormode of `any`.

This is a **breaking** change.

issue: #990


kyranjamie (approved):

zone117x (changes requested):
These changes look good. I think the jsdoc for the anchor mode enum and options should be expanded a bit to explain what they do, and a `@see` links to the SIP explaining in more detail. 
https://github.com/blockstack/stacks.js/blob/66741ec46e1f37fb2ade7d2228f21a29635bdd04/packages/transactions/src/constants.ts#L41-L51

PR #1025, opened 2021-06-02
stacks/transactions: verifySingleSig() and verifyMultiSig() work like the rust code

fariedt:
This PR adds the code to match `stacks-blockchain`'s verification of signing conditions in transactions.  In particular, this adds support for verifying multisig signatures.

This code is needed to complete the STX integration for [BitGo](https://github.com/BitGo/BitGoJS).

## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Other

Yes -- see the failing test: *STX token transfer transaction multi-sig uncompressed keys serialization and deserialization*

The code in `stacks-blockchain` appears to reject such transactions (line 296 or so in `src/chainstate/stacks/auth.rs`):
```rust
if have_uncompressed && self.hash_mode == MultisigHashMode::P2WSH {
            return Err(net_error::VerifyingError(
                "Uncompressed keys are not allowed in this hash mode".to_string(),
            ));
        }
```

All changes are internal.

## Testing information

Needs more tests.

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


fariedt:
Adapting the tests from `stacks-blockchain` to work with this code is difficult -- some of the constants used by `authorization.ts` are not exported, like `AuthType` and `AddressHashMode`.

What should the next steps be, @reedrosenbluth?

reedrosenbluth:
@fariedt those constants appear to both be exported and usable from tests

fariedt:
I got that wrong -- what isn't exported is `AuthFieldType`.  It's not necessary, though.

I've added three more tests from `stacks-blockchain` for multi-sig support.  I've also added support for P2WSH and P2WPKH address types (needed for some of the authorization tests).

reedrosenbluth:
@fariedt is this still WIP?

fariedt:
@reedrosenbluth I've rebased this against master, and fixed a couple of lint + type errors, and removed the WIP label.

There is an older test case that fails: #bcf66b75b --  STX token transfer transaction multi-sig uncompressed keys serialization and deserialization.  It doesn't seem like `stacks-blockchain` allows the use of uncompressed keys here, and my code reflects that.

kantai (changes requested):
This implementation looks good to me, but I'd like to see more tests of the verification function that use something like transaction vectors (both valid and invalid) to test that the verification routine is operating correctly.

ahsan-javaid:
> This implementation looks good to me, but I'd like to see more tests of the verification function that use something like transaction vectors (both valid and invalid) to test that the verification routine is operating correctly.


Update:
- Added the aforementioned tests (valid/invalid) using transaction vectors
- Added missing checks in deserialize spending conditions
- Fixed and updated breaking tests
Ready for rereview or if it's needs any further actions.

kantai (approved):
LGTM! Thanks @fariedt

packages/transactions/src/authorization.ts
```diff
@@ -488,11 +517,77 @@ function verifySingleSig(
...
 
+function verifyMultiSig(
+  condition: MultiSigSpendingConditionOpts,
+  initialSigHash: string,
+  authType: AuthType
+): string {
+  const publicKeys: StacksPublicKey[] = [];
+  let curSigHash = initialSigHash;
+  let haveUncompressed = false;
+  const numSigs = new Uint16Array(1);
```
reedrosenbluth: why is `numSigs` a `Uint16Array`? It appears that only the first element is accessed.

reedrosenbluth (approved):
LGTM! Left a minor comment.

packages/transactions/src/authorization.ts
asimm241: @reedrosenbluth  I think Faried used it to save space, number takes 2^53, so he used `Uint16Array`, because I don't think there is a built-in `Uint16` type supported in the typescript. 

I can change it to the `number` if you want?

reedrosenbluth: I don't think this is necessary. @fariedt can you explain why you did this?

fariedt: To be honest, I don't remember why I went with this approach because I haven't seen it used elsewhere in this repository, or used it myself somewhere.  Maybe I was trying to make it explicit that it was a 16-bit value (the Rust code uses a `u16`).

I don't think I was trying to save space by using a single-element `Uint16Array`; it likely takes up more space than a normal variable.

reedrosenbluth: @asimm241 thanks for refactoring this! This paradigm is used [elsewhere in the file](https://github.com/blockstack/stacks.js/pull/1025/files#diff-0bc9083c9c7534a77c7291df62234e5e3b66abd23dfadd8320bb17e44a72854fR329). can you also make the change there and wherever else this is done? 🙏

Issue #1016, opened 2021-05-01
lookupProfile throwing error with new BNS system


pradel:
**Describe the bug**

Calling `lookupProfile` is throwing the following error with a newly created account after BNS was restored on mainnet.

```
[ERROR] resolveZoneFileToProfile: error fetching token file https://gaia.blockstack.org/hub/1AZoc8UktzfawbanMDuPWGp8m3Vz1tp2nK/profile.json: Error: Token issuer public key does not match the verifying value
Error: Token issuer public key does not match the verifying value
    at verifyProfileToken (/Users/leo/dev/pradel/sigle/node_modules/@stacks/auth/node_modules/@stacks/profile/dist/profileTokens.js:77:15)
    at Object.extractProfile (/Users/leo/dev/pradel/sigle/node_modules/@stacks/auth/node_modules/@stacks/profile/dist/profileTokens.js:93:24)
    at /Users/leo/dev/pradel/sigle/node_modules/@stacks/auth/node_modules/@stacks/profile/dist/profile.js:300:49
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

**What version of Blockstack.js are you using?**

1.5.0-alpha.0

**Is the bug present in the Blockstack CLI, Gaia hub, Connect or Blockstack Browser?**

`@stacks/auth`

**To Reproduce**

To reproduce run the following snippet:

```
import { lookupProfile } from '@stacks/auth';

await lookupProfile({ username: 'testusenameregistration.id.stx' });
```

**Expected behavior**

The profile should be returned.

**Additional context**

@hstove I tried to update the SDK after your comment https://github.com/blockstack/stacks-wallet-web/issues/946#issuecomment-828628579

markmhendrickson:
@agraebe Is this something the splat team can prioritize soon to unblock @pradel with the restoration of usernames per https://github.com/blockstack/stacks-wallet-web/issues/946#issuecomment-830640990? He's been waiting patiently for BNS support this year so it'd be great to resolve his related issues promptly here.

agraebe:
I'll check but probably cannot get to look at it before next week. cc @zone117x @reedrosenbluth

pradel:
Any update? 🙏

reedrosenbluth:
@pradel looking into this!

pradel:
@reedrosenbluth thanks 😊

pradel:
I took some time and found the issue.
[...]

PR #1028, opened 2021-06-04
fix: make `lookupProfile` work with newest stacks API

pradel:
## Description

In the latest version of the SDK, `lookupProfile` is throwing an error when you call it to query old or new usernames.

I faced this issue when migrating Sigle to the newest SDK version.

- the old SDK version is calling https://core.blockstack.org/v1/names/sigleapp.id.blockstack that returns the `address` field as a BTC address
- the new SDK version is calling https://stacks-node-api.mainnet.stacks.co/v1/names/sigleapp.id.blockstack returns the `address` field as a STX address

When you call `lookupProfile`, internally the `verifyProfileToken` function is called.
If you take a look here https://github.com/blockstack/stacks.js/blob/626c5fccb4dff213f000923ab69c7ed4f5eb48a0/packages/profile/src/profileTokens.ts#L115-L123
`publicKeyOrAddress` is the STX address returned by the new API and we compare it with the BTC address that we got from the public key so the check is failing.

For details refer to issue https://github.com/blockstack/stacks.js/issues/1016

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

There is a breaking change in the `verifyProfileToken` function only if a user uses it directly. By looking at the code it looks like `lookupProfile` is the only part where it is called.

## Testing information

I tested this patch directly into the Sigle codebase and it solved the issue that I am facing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
 	"requires": true,
 	"packages": {
 		"": {
+			"name": "stacks.js",
 			"version": "1.4.1",
 			"workspaces": [
 				"packages/**"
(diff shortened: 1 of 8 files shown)


reedrosenbluth (approved):
LGTM. Thanks @pradel!

zone117x (approved):
LGTM, thanks for the fix!!

PR #1029, opened 2021-06-10
chore: fix #983 and bugs with Clarity int/uint value encoding

### LINKED ISSUE #983: cvToValue throws exception for uint with more than 53 bits
**Describe the bug**
`cvToJSON` throws expection for `uint` or `int` with more than 53 bits. `cvToJSON` uses `cvToValue` 

transactions 1.3.3

no

**To Reproduce**
```
  it("should convert uintCV to hex", async () => {
    const value = cvToJSON(uintCV("1000000000000000"));
    expect(JSON.stringify(value)).equal('{"type":"uint","value":10000000000000000}');
  });
``` 

**Expected behavior**
Not clear how to handle this. For small numbers it makes sense to just return the number, for bigger numbers we need BN values.

**Console log**

```
 should convert uintCV to hex:
     Error: Number can only safely store up to 53 bits
      at assert (node_modules/bn.js/lib/bn.js:6:21)
      at BN.toNumber (node_modules/bn.js/lib/bn.js:519:7)
      at cvToValue (node_modules/@stacks/transactions/src/clarity/clarityValue.ts:108:24)
      at Object.cvToJSON (node_modules/@stacks/transactions/src/clarity/clarityValue.ts:144:51)
      at Context.<anonymous> (test/unit/pool-tool-utils.spec.ts:60:19)
      at processImmediate (internal/timers.js:461:21)
```

zone117x:
This PR fixes #983 in a backwards compatible way, as well as fixing a handful of other issues related to how Clarity Ints and UInts were being constructed and serialized.

The fix for #983 is a bit of a kludge in order to prevent breaking changes. The `cvToValue` function checks if the BigNum is larger than 53 bits (Number.MAX_SAFE_INTEGER), if it is then a string quoted integer is returned, otherwise it preserves the original behavior of returning a js `number`. For example, the Clarity value `123` is returned as a js number `123`, while the Clarity value `9007199254740992` is returned as a js string `"9007199254740992"`. 

I left a TODO comment around this code for fixing this in the future to always return the same type (either a string or a bigint) because returning different types dependent on the value is surprising behavior. @friedger @reedrosenbluth let me know if we just want to go ahead and make the breaking change now to avoid this strange behavior. Easy change.

In addition to the above change, several fixes were made around serializing and deserializing Clarity integer types. Prior to this PR, both signed-ints and unsigned-ints were being encoded with twos-complement. This caused unsigned integers to incorrectly accept negative input values. It also caused negative signed-integers to only work when passing through both the serialize and deserialize function. For example, the following code would fail:
```js
const intNegativeTen = intCV(-10);
expect(intNegativeTen.value.toString()).toEqual('-10');
// Error:
//    Expected: "-10"
//    Received: "340282366920938463463374607431768211446"
```

The `intCV(...)` and `uintCV(...)` functions were also expanded to accept `bigint` values and hex strings. Max/min bit size and base-10 number range checks were also added to prevent out-of-bounds integers from being passed to these functions. `TypeError`s and `RangeError`s are now thrown in relevant conditions.


packages/transactions/src/clarity/types/intCV.ts
```diff
@@ -1,41 +1,103 @@
...
+    if (signed) {
+      if (value.byteLength < CLARITY_INT_BYTE_SIZE) {
+        const paddedBytes = new Uint8Array(CLARITY_INT_BYTE_SIZE);
+        // Extend sign-bit while padding to full byte length.
+        if (value[0] >>> 7) {
+          paddedBytes.fill(0xff);
+        }
+        paddedBytes.set(value, CLARITY_INT_BYTE_SIZE - value.byteLength);
+        bn = new BigNum(paddedBytes, 'be').fromTwos(CLARITY_INT_SIZE);
+      } else {
```
zone117x: @kantai tagged you as a reviewer just to take a look at this sign extension code. 

Edit: code has now been simplified to use `.fromTwos(value.byteLength * 8)` since we operate on 8-bit boundaries (buffers / uint8arrays), and the 128-bit padding is performed later during serialization.

reedrosenbluth:
Sweet! I think we should just go ahead and make the breaking change since we will be releasing `v2.0.0` soon anyway.

reedrosenbluth (approved):

kantai (approved):
This LGTM -- is there an issue tracking the odd return type of integers (i.e., the switch from native number to string at large values)? If not, it probably makes sense to create one.

zone117x:
> This LGTM -- is there an issue tracking the odd return type of integers (i.e., the switch from native number to string at large values)? If not, it probably makes sense to create one.

@kantai there's not an issue, but there's follow up PR that fixes it https://github.com/blockstack/stacks.js/pull/1030

PR #1030, opened 2021-06-11
Feat/native bigint

zone117x:
Creating a separate PR from https://github.com/blockstack/stacks.js/pull/1029 because this one introduces breaking changes, and adds a bunch more unit test update clutter.

Int and UInt Clarity values now store a native `bigint`, and the `cvToValue(...)` function returns a `bigint` rather than `number` or `string`. The `cvToJSON(...)` function preserves valid JSON output, and returns integers as string quoted. These are both breaking changes. 

Generally, most of the function input types were made permissive so that numeric arguments like `fee`, `nonce`, `amount`, etc accept the type `type IntegerType = number | string | bigint | Uint8Array | BN`. This is _not_ a breaking change and should make these functions a bit friendlier/easy to use. E.g. constructing a `BN.js` instance is no longer required. 

Generally, most of the function output types that were previously a `BN.js` instance are now a `bigint` instance. This one is a breaking change.


packages/transactions/src/clarity/serialize.ts
```diff
@@ -43,12 +44,14 @@ function serializeBufferCV(cv: BufferCV): Buffer {
 }
 
 function serializeIntCV(cv: IntCV): Buffer {
-  const buffer = cv.value.toTwos(CLARITY_INT_SIZE).toArrayLike(Buffer, 'be', CLARITY_INT_BYTE_SIZE);
+  const buffer = new BN(cv.value.toString())
+    .toTwos(CLARITY_INT_SIZE)
+    .toArrayLike(Buffer, 'be', CLARITY_INT_BYTE_SIZE);
```
reedrosenbluth, line 49: Is bn.js still the preferred way to do this conversion?

reedrosenbluth (approved):
Really appreciate how thorough and well thought out this PR is! Awesome work @zone117x!

packages/transactions/src/clarity/serialize.ts
zone117x: I have some code to do it natively with `bigint` from a while back, but I'm not 100% confident in it. Needs a bunch of tests:
```ts
export function toTwos(value: bigint) {
  // make sure its in range given the number of bits
  if (value < -(1n << (128n - 1n)) || value > (1n << (128n - 1n)) - 1n)
    throw 'Integer out of range given 128 bits to represent.';

  // if positive, return the positive value
  if (value >= 0n) return value;

  // if negative, convert to twos complement representation
  return ~((-value - 1n) | ~((1n << 128n) - 1n));
}

export function fromTwos(value: bigint) {
  if ((value & (1n << (128n - 1n))) > 0n) {
    value = value - (1n << 128n);
  }
  return value;
}
```
Maybe a good task for a separate issue?

zone117x:
@kyranjamie could use a review and approval when you get a chance

PR #1031, opened 2021-06-14
fix: fix #1026 [BNS] Incorrect optional argument encoding

### LINKED ISSUE #1026: [BNS] Incorrect optional argument encoding.
**Describe the bug**
The functions `buildTransferNameTx` and `buildRenewNameTx` incorrectly encode some of the contract call arguments. Optional arguments are either included or omitted (as can be seen [here](https://github.com/blockstack/stacks.js/blob/master/packages/bns/src/index.ts#L683)), which results in the created transactions being rejected (either not enough arguments, or a `TypeError` is thrown because the optional arguments are of the wrong type).

2.0.0-beta.0

**To Reproduce**
Build a transaction by calling one of the aforementioned `buildTransferNameTx` or `buildRenewNameTx` functions (with or without passing the optional arguments), and broadcast the resulting transaction. This will result in an error.

**Additional context**
While debugging the issue, I've updated the relevant functions to encode the arguments correctly. The required changes can be seen in [this commit](https://github.com/Exulansis/stacks.js/commit/8a70ea8ac4c7b294a2f4220af12502743c1d3853). I have tested the fixes locally against a mocknet instance, and they seem to address the problem. I have not updated the tests.

ahsan-javaid:
## Description
This PR fixes the issue of optional argument encoding in BNS package in `buildTransferNameTx` and `buildRenewNameTx` functions. The onwerAddress and zonefile parameters are incorrectly encoded which results in the created transactions being rejected (either not enough arguments, or a TypeError is thrown because the optional arguments are of the wrong type)
Below is the script to reproduce the issue
```
void (async () => {
  const fullyQualifiedName = 'test.id';
  const newOwnerAddress = '<Owner address>';
  const pk = '<private key>';
  const zonefile = 'zonefile';
  const publicKey = publicKeyToString(pubKeyfromPrivKey(pk));
  const unsignedTransaction = await buildTransferNameTx({
    fullyQualifiedName,
    newOwnerAddress,
    publicKey,
    zonefile,
    network
  });
  const signer = new TransactionSigner(unsignedTransaction);
  signer.signOrigin(createStacksPrivateKey(pk));
      signer.transaction,
      network
  );
  console.log(reply);
})();
// This will result in TypeError and transaction will be rejected.
```

For details refer to issue #1026 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No breaking changes. 

No

## Testing information
The existing test cases related to `buildTransferNameTx` and  `buildRenewNameTx` are updated to include the optional types. And two new test cases are added to test the optional type parameters particularly. This change can be seen in BNS test cases in this PR. 

Steps to test
1. Run `npm run test` inside BNS package to test changes
2. Other way of testing this pr is making the contract calls to `buildTransferNameTx` and `buildRenewNameTx` using the script mentioned in the aforementioned script in the description section 

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/bns/src/index.ts
+++ b/packages/bns/src/index.ts
@@ -15,6 +15,8 @@ import {
   ResponseErrorCV,
   StacksTransaction,
   standardPrincipalCV,
+  someCV,
+  noneCV,
   UnsignedContractCallOptions,
 } from '@stacks/transactions';
 
@@ -677,12 +679,9 @@ export async function buildTransferNameTx({
   const functionArgs = [
     bufferCVFromString(namespace),
     bufferCVFromString(name),
-    bufferCVFromString(newOwnerAddress),
+    standardPrincipalCV(newOwnerAddress),
+    zonefile ? someCV(bufferCV(getZonefileHash(zonefile))): noneCV()
   ];
-
-  if (zonefile) {
-    functionArgs.push(bufferCV(getZonefileHash(zonefile)));
-  }
 
   return makeBnsContractCall({
     functionName: bnsFunctionName,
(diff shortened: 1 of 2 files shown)


ahsan-javaid:
cc: @asimm241

reedrosenbluth (approved):
Thanks!

PR #1032, opened 2021-06-17
Add missing stx burn and NFT post conditions

### LINKED ISSUE #1027: [BNS] Missing STX burn / NFT spend post conditions.
**Describe the bug**
When a user preorders a new BNS `namespace` or `name`, they must burn a specific amount of STX (depending on the `name` / `namespace` price). In order for the funds to be correctly spent, and the transaction to succeed, an [appropriate post-condition](https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md#transaction-post-conditions) needs to be included in the transaction.

The post-condition needs to be created as part of the appropriate function call (i.e. [`buildNamespacePreorderTx`](https://github.com/blockstack/stacks.js/blob/master/packages/bns/src/index.ts#L288), [`buildNamePreorderTx`](https://github.com/blockstack/stacks.js/blob/7b6ceb33c835804822e19ad124e162d1b3cd3879/packages/bns/src/index.ts#L504)) and passed further to the [`makeBnsContractCall`](https://github.com/blockstack/stacks.js/blob/master/packages/bns/src/index.ts#L299), and eventually the [`makeUnsignedContractCall`](https://github.com/blockstack/stacks.js/blob/7b6ceb33c835804822e19ad124e162d1b3cd3879/packages/bns/src/index.ts#L85) functions.

 2.0.0-beta.0

**To Reproduce**
Build a transaction by calling one of the aforementioned functions, and broadcast the resulting transaction. This will result in the transaction being aborted and reverted because of the (lack of) post-conditions.

**Additional context**

While debugging the issue, I've updated the relevant functions to encode the arguments correctly. The required changes can be seen [in this commit](https://github.com/Exulansis/stacks.js/commit/c7a2996c874d80a75de4bb0594d9edbc14914576). I have tested the fixes locally against a mocknet instance, and they seem to address the problem. I have not updated the tests.

It is important to point out that a **post-condition is required for transferring the name as well**. I have made an attempt to tackle this as well (as can be seen [here](https://github.com/Exulansis/stacks.js/commit/c7a2996c874d80a75de4bb0594d9edbc14914576#diff-be52d9579cd369a5e9dec98f1827bc27ac5899883cb293587e32dc0a4e5fa1c0R756)), **but the resulting transaction still fails, most likely because the post-conditions are incorrectly created (or not sufficient)**. (p.s. I have tried different configurations, with one, or two post-conditions, but with no luck)

I will continue digging on this topic a bit further, but in case anyone more familiar with post-conditions for non-fungible assets can point out my mistake, I'd be thankful!

ahsan-javaid:
## Description
This PR fixes the issue of missing post conditions in BNS package in following functions:
* `buildPreorderNamespaceTx`
* `buildPreorderNameTx`
* `buildRenewNameTx`
* `buildTransferNameTx`
These functions are missing the stx burn and NFT post conditions which results in transactions being rolled back or aborted during the execution. The stacks explorer displays the error message like
`This transaction would have succeeded, but was rolled back by a supplied post-condition.`
Below is the script to reproduce the issue
```
import {
  TxBroadcastResult,
  broadcastTransaction,
  publicKeyToString,
  TransactionSigner,
  createStacksPrivateKey,
  pubKeyfromPrivKey,
} from './stacks.js/packages/transactions';

import {
  StacksTestnet
} from './stacks.js/packages/network';

import {
   buildPreorderNameTx
} from './stacks.js/packages/bns/src';
import BN from "bn.js";
void (async () => {
  const fullyQualifiedName = 'test.id';
  const salt = 'salt';
  const stxToBurn = new BN(200);
  const pk = '<public key here>';
  const publicKey = publicKeyToString(pubKeyfromPrivKey(pk));
  const unsignedTransaction = await buildPreorderNameTx({
    fullyQualifiedName,
    salt,
    stxToBurn,
    publicKey,
    network
  });
  const signer = new TransactionSigner(unsignedTransaction);
  signer.signOrigin(createStacksPrivateKey(pk));

      signer.transaction,
      network
  );
  console.log('Tx response', reply);
  console.log('Verify at: ', `https://explorer.stacks.co/txid/${reply}?chain=testnet`);
})();
// This will result in Notice: This transaction would have succeeded, but was rolled back by a supplied post-condition.
Example: https://explorer.stacks.co/txid/becf70878b539e4fcebbb06e376aa061465a1166cb75cfa700310aea9f819a3a?chain=testnet
```

This pull request closes the issue #1027

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No breaking changes. 

No

## Testing information
The existing test cases related to aforementioned functions are updated to include the postconditions. This change can be seen in BNS test cases in this PR.

Steps to test
1. Run `npm run test` inside BNS package to test changes
2. Other way of testing this pull request is by making the contract calls to aforementioned functions using the script mentioned in the description section. The script will print out the broadcasted transaction id and its url on testnet where the result of the transaction can be verified.
`Example: https://explorer.stacks.co/txid/605ff2961d82201bb1c4c2ba3c4b3984b3b80de95a7948a7a37d213f4aa53fc6?chain=testnet`

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/bns/src/index.ts
+++ b/packages/bns/src/index.ts
@@ -18,6 +18,16 @@ import {
   someCV,
   noneCV,
   UnsignedContractCallOptions,
+  PostCondition,
+  createSTXPostCondition,
+  createStacksPublicKey,
+  publicKeyToAddress,
+  FungibleConditionCode,
+  AddressVersion,
+  createNonFungiblePostCondition,
+  NonFungibleConditionCode,
+  parseAssetInfoString,
+  tupleCV,
 } from '@stacks/transactions';
 
 import { StacksNetwork } from '@stacks/network';
@@ -39,6 +49,12 @@ function getBnsContractAddress(network: StacksNetwork) {
   if (network.chainId === ChainID.Mainnet) return BnsContractAddress.mainnet;
   else if (network.chainId == ChainID.Testnet) return BnsContractAddress.testnet;
   else throw new Error(`Unexpected ChainID: ${network.chainId}`);
+}
+
+function getAddressVersion(network: StacksNetwork) {
+  return network.chainId === ChainID.Mainnet
+    ? AddressVersion.MainnetSingleSig
+    : AddressVersion.TestnetSingleSig;
 }
 
 export interface PriceFunction {
@@ -70,6 +86,7 @@ export interface BnsContractCallOptions {
   publicKey: string;
   network: StacksNetwork;
   attachment?: Buffer;
+  postConditions?: PostCondition[];
 }
 
 async function makeBnsContractCall(options: BnsContractCallOptions): Promise<StacksTransaction> {
@@ -82,6 +99,7 @@ async function makeBnsContractCall(options: BnsContractCallOptions): Promise<Sta
     validateWithAbi: false,
     network: options.network,
     anchorMode: AnchorMode.Any,
+    postConditions: options.postConditions,
   };
 
   return makeUnsignedContractCall(txOptions);
@@ -298,11 +316,18 @@ export async function buildPreorderNamespaceTx({
   const saltedNamespaceBuffer = Buffer.from(`${namespace}${salt}`);
   const hashedSaltedNamespace = hash160(saltedNamespaceBuffer);
 
+  const burnSTXPostCondition = createSTXPostCondition(
+    publicKeyToAddress(getAddressVersion(network), createStacksPublicKey(publicKey)),
+    FungibleConditionCode.Equal,
+    stxToBurn
+  );
+
   return makeBnsContractCall({
     functionName: bnsFunctionName,
     functionArgs: [bufferCV(hashedSaltedNamespace), uintCVFromBN(stxToBurn)],
     publicKey,
     network,
+    postConditions: [burnSTXPostCondition],
   });
 }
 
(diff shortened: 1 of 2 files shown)


ahsan-javaid:
cc: @asimm241

ahsan-javaid:
@reedrosenbluth Need your review on this pull request.

ahsan-javaid:
@agraebe Need to assign @reedrosenbluth as reviewer on this pull request.

packages/bns/src/index.ts
```diff
@@ -682,13 +714,32 @@ export async function buildTransferNameTx({
     standardPrincipalCV(newOwnerAddress),
     zonefile ? someCV(bufferCV(getZonefileHash(zonefile))) : noneCV(),
   ];
+  const nameTransferPostConditionOne = createNonFungiblePostCondition(
```
reedrosenbluth: nit: could you give these variables more descriptive names?

perhaps something like: `postConditionSender`, `postConditionReceiver`

reedrosenbluth (approved):
Left a small comment, but otherwise this LGTM! I think we should wait for another review before merging.

packages/bns/src/index.ts
ahsan-javaid: @reedrosenbluth Variable names changed. Give it a second look.

kyranjamie (approved):
My review is just of the code—haven't pulled the branch—but LGTM

PR #1040, opened 2021-06-23
ci: improve ci jobs

kyranjamie:
These changes break the 1 "code checks" into more specific CI tasks.

While there's some duplication of the job scaffolding/caching etc, this is much more developer-friendly. 

- It's faster as they run in parallel
- You can see if multiple checks fail (rather than it exiting immediately when the first fails)
- You get easier insights into what breaks "linting" rather than generic "code checks"

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,12 +1,12 @@
-name: stacks.js
+name: Lint
 
 on:
   push:
-  pull_request:
+    branches:
+      - '**'
 
 jobs:
-  code_checks:
-    name: Code Checks
+  lint:
     runs-on: ubuntu-latest
     env:
       NODE_OPTIONS: --max-old-space-size=4096
@@ -23,7 +23,7 @@ jobs:
         with:
           path: |
             node_modules
-            */*/node_modules
+            packages/*/node_modules
           key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
 
       - name: Install dependencies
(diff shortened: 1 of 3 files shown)


CharlieC3 (approved):
LGTM :shipit: 

Also once I get to work on [this issue](https://github.com/hirosystems/devops/issues/726), I plan on making some changes to CI as well which is going to modify these new workflows, _but_ it it will not change or eliminate any of the 3 benefits you listed in the description. The biggest reason is because if linting or typechecks fail, CI should not attempt a release. And if the steps are split into separate workflows, we can't wire that logic up in a way that makes sense.

kyranjamie:
All good, though have you seen https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run I think it's possible now @CharlieC3

CharlieC3:
> All good, though have you seen https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run I think it's possible now @CharlieC3

Thanks for looking into it -- I'm aware of that feature, but it makes certain things like figuring out if the workflow was successful or not, and passing arguments between workflows still difficult. If they're in the same workflow, you can have all that plus the parallelization and the other benefits you mentioned above. It just requires a different setup than what the original workflow was doing.

Plus linting and code checks are very much a CI-type of activity. So it makes sense to have all of these CI activities under the same umbrella. It's a simplified approach that also helps reduce the burden on anyone managing the CI workflows. But the changes this PR introduces are a good compromise to address the issues you mentioned above until I can spend some time on the CI.

kyranjamie:
@CharlieC3 any ideas why the codecov is failing? It's saying this CI-only change introduces a -7% code coverage 🤷🏼

CharlieC3:
@kyranjamie I'm getting an HTTP 400 error when trying to view the codecov report. I suspect the answer lies in there. Are you able to access it?

agraebe:
@kyranjamie do you want to merge this in and @CharlieC3 will tackle the rest in a new PR?

PR #1042, opened 2021-06-23
Add consistent broadcastTransaction return type

### LINKED ISSUE #830: Transactions: broadcastTransaction return type is inconsistent
The `broadcastTransaction` method either returns a string (txid) if successful or a JSON object with error details:

```js
export declare type TxBroadcastResultOk = string;
export declare type TxBroadcastResultRejected = {
    error: string;
    reason: TxRejectedReason;
    reason_data: any;
    txid: string;
};
export declare type TxBroadcastResult = TxBroadcastResultOk | TxBroadcastResultRejected;
```

We should make the response type consistent and always return an object. Developers would use the response and simply check ```if(resp.error)``` for error handling. 

Question: Could we also return null for the txid if a transaction is rejected? From a developer standpoint, what is the value of a rejected txid?

### LINKED ISSUE #1054: `broadcastRawTransaction` returns string error message on certain errors
I was trying to deploy a contract with an emoji in it, and saw that our extension was failing silently. We're using `broadcastRawTransaction` to send out our transactions, and in this case the response was a string value of `Failed to deserialize posted transaction: Invalid Stacks string: non-printable or non-ASCII string.`. Our code expected it to throw rather than return a string. For now I have https://github.com/blockstack/stacks-wallet-web/pull/1434 to address the issue, but the function should only return a string if it's a successful call.

ahsan-javaid:
## Description
This PR fixes the issue of inconsistent broadcastTransaction return type. The broadcastTransaction returns txid as string which is inconsistent w.r.t transaction rejection response which resolves to object as per below types.
```
type TxBroadcastResultOk = string;
type TxBroadcastResultRejected = {
  error: string;
  reason: TxRejectedReason;
  reason_data: any;
  txid: string;
}
type TxBroadcastResult = TxBroadcastResultOk | TxBroadcastResultRejected;
```
Above two data types are inconsistent as one is string and other is object.  To resolve the issue i have updated the `TxBroadcastResultOk` as follows to make both datatypes consistent.
```
export type TxBroadcastResultOk = {
  txid: string;
};
export type TxBroadcastResultRejected = {
  error: string;
  reason: TxRejectedReason;
  reason_data: any;
  txid: string;
};
export type TxBroadcastResult = TxBroadcastResultOk | TxBroadcastResultRejected;

```
This pull request closes the issue #830

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

Users of stacks.js after broadcasting transaction should expect the transaction return type as aforementioned types.

No changes in readMe file as the broadcast transaction response is not documented in readMe of this package.

## Testing information
Steps to test
1. Run `npm run test` inside transaction package to test the changes
2. Other way of testing this pull request is by broadcasting a test transaction which will resolve to below sample response
```
// Send a sample transaction using 
      transaction,
      network
  );
  console.log('Tx response', reply);
Tx response {
  txid: '2b2badc7d9168f92f87971fcf0837a98c961ab2232d430c8552ce38f8367f438'
}

```

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/transactions/src/builders.ts
+++ b/packages/transactions/src/builders.ts
@@ -52,7 +52,7 @@ import {
 
 import { AssetInfo, createLPList, createStandardPrincipal, createContractPrincipal } from './types';
 
-import { cvToHex, parseReadOnlyResponse, omit } from './utils';
+import { cvToHex, parseReadOnlyResponse, omit, validateTxId } from './utils';
 
 import { fetchPrivate } from '@stacks/common';
 
@@ -137,13 +137,184 @@ export async function estimateTransfer(
   return feeRate * txBytes;
 }
 
-export type TxBroadcastResultOk = string;
-export type TxBroadcastResultRejected = {
+export type SerializationRejection = {
   error: string;
-  reason: TxRejectedReason;
-  reason_data: any;
+  reason: TxRejectedReason.Serialization;
+  reason_data: {
+    message: string;
+  };
   txid: string;
 };
+
+export type DeserializationRejection = {
+  error: string;
+  reason: TxRejectedReason.Deserialization;
+  reason_data: {
+    message: string;
+  };
+  txid: string;
+};
+
+export type SignatureValidationRejection = {
+  error: string;
+  reason: TxRejectedReason.SignatureValidation;
+  reason_data: {
+    message: string;
+  };
+  txid: string;
+};
+
+export type BadNonceRejection = {
+  error: string;
+  reason: TxRejectedReason.BadNonce;
+  reason_data: {
+    expected: number;
+    actual: number;
+    is_origin: boolean;
+    principal: boolean;
+  };
+  txid: string;
+};
+
+export type FeeTooLowRejection = {
+  error: string;
+  reason: TxRejectedReason.FeeTooLow;
+  reason_data: {
+    expected: number;
+    actual: number;
+  };
+  txid: string;
+};
+
+export type NotEnoughFundsRejection = {
+  error: string;
+  reason: TxRejectedReason.NotEnoughFunds;
+  reason_data: {
+    expected: string;
+    actual: string;
+  };
+  txid: string;
+};
+
+export type NoSuchContractRejection = {
+  error: string;
+  reason: TxRejectedReason.NoSuchContract;
+  reason_data?: undefined;
+  txid: string;
+};
+
+export type NoSuchPublicFunctionRejection = {
+  error: string;
+  reason: TxRejectedReason.NoSuchPublicFunction;
+  reason_data?: undefined;
+  txid: string;
+};
+
+export type BadFunctionArgumentRejection = {
+  error: string;
+  reason: TxRejectedReason.BadFunctionArgument;
+  reason_data: {
+    message: string;
+  };
+  txid: string;
+};
+
+export type ContractAlreadyExistsRejection = {
+  error: string;
+  reason: TxRejectedReason.ContractAlreadyExists;
+  reason_data: {
+    contract_identifier: string;
+  };
+  txid: string;
+};
+
+export type PoisonMicroblocksDoNotConflictRejection = {
+  error: string;
+  reason: TxRejectedReason.PoisonMicroblocksDoNotConflict;
+  reason_data?: undefined;
+  txid: string;
+};
+
+export type PoisonMicroblockHasUnknownPubKeyHashRejection = {
+  error: string;
+  reason: TxRejectedReason.PoisonMicroblockHasUnknownPubKeyHash;
+  reason_data?: undefined;
+  txid: string;
+};
+
+export type PoisonMicroblockIsInvalidRejection = {
+  error: string;
+  reason: TxRejectedReason.PoisonMicroblockIsInvalid;
+  reason_data?: undefined;
+  txid: string;
+};
+
+export type BadAddressVersionByteRejection = {
+  error: string;
+  reason: TxRejectedReason.BadAddressVersionByte;
+  reason_data?: undefined;
+  txid: string;
+};
+
+export type NoCoinbaseViaMempoolRejection = {
+  error: string;
+  reason: TxRejectedReason.NoCoinbaseViaMempool;
+  reason_data?: undefined;
+  txid: string;
+};
+
+export type ServerFailureNoSuchChainTipRejection = {
+  error: string;
+  reason: TxRejectedReason.ServerFailureNoSuchChainTip;
+  reason_data?: undefined;
+  txid: string;
+};
+
+export type ServerFailureDatabaseRejection = {
+  error: string;
+  reason: TxRejectedReason.ServerFailureDatabase;
+  reason_data: {
+    message: string;
+  };
+  txid: string;
+};
+
+export type ServerFailureOtherRejection = {
+  error: string;
+  reason: TxRejectedReason.ServerFailureOther;
+  reason_data: {
+    message: string;
+  };
+  txid: string;
+};
+
+export type TxBroadcastResultOk = {
+  txid: string;
+  error?: undefined;
+  reason?: undefined;
+  reason_data?: undefined;
+};
+
+export type TxBroadcastResultRejected =
+  | SerializationRejection
+  | DeserializationRejection
+  | SignatureValidationRejection
+  | BadNonceRejection
+  | FeeTooLowRejection
+  | NotEnoughFundsRejection
+  | NoSuchContractRejection
+  | NoSuchPublicFunctionRejection
+  | BadFunctionArgumentRejection
+  | ContractAlreadyExistsRejection
+  | PoisonMicroblocksDoNotConflictRejection
+  | PoisonMicroblockHasUnknownPubKeyHashRejection
+  | PoisonMicroblockIsInvalidRejection
+  | BadAddressVersionByteRejection
+  | NoCoinbaseViaMempoolRejection
+  | ServerFailureNoSuchChainTipRejection
+  | ServerFailureDatabaseRejection
+  | ServerFailureOtherRejection;
+
 export type TxBroadcastResult = TxBroadcastResultOk | TxBroadcastResultRejected;
 
 /**
@@ -199,10 +370,15 @@ export async function broadcastRawTransaction(
   }
 
   const text = await response.text();
-  try {
-    return JSON.parse(text) as TxBroadcastResult;
-  } catch (e) {
-    return text;
+  // Replace extra quotes around txid string
+  const txid = text.replace(/["]+/g, '');
+  const isValidTxId = validateTxId(txid);
+  if (isValidTxId) {
+    return {
+      txid: txid,
+    } as TxBroadcastResult;
+  } else {
+    throw new Error(text);
   }
 }
 
--- a/packages/transactions/src/utils.ts
+++ b/packages/transactions/src/utils.ts
@@ -7,6 +7,7 @@ import { deserializeCV } from './clarity';
 import fetch from 'cross-fetch';
 import { c32addressDecode } from 'c32check';
 import lodashCloneDeep from 'lodash/cloneDeep';
+import { with0x } from '@stacks/common';
 
 export { randombytes as randomBytes };
 
@@ -205,4 +206,11 @@ export const validateStacksAddress = (stacksAddress: string): boolean => {
   } catch (e) {
     return false;
   }
+};
+
+export const validateTxId = (txid: string): boolean => {
+  if (txid === 'success') return true; // Bypass fetchMock tests
+  const value = with0x(txid).toLowerCase();
+  if (value.length !== 66) return false;
+  return with0x(BigInt(value).toString(16).padStart(64, '0')) === value;
 };
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -1648,8 +1648,8 @@ async function stack(network: CLINetworkAdapter, args: string[]): Promise<string
         return response;
       }
       return {
-        txid: `0x${response}`,
-        transaction: generateExplorerTxPageUrl(response as string, txNetwork),
+        txid: `0x${response.txid}`,
+        transaction: generateExplorerTxPageUrl(response.txid, txNetwork),
       };
     })
     .catch(error => {
--- a/packages/common/src/utils.ts
+++ b/packages/common/src/utils.ts
@@ -389,3 +389,7 @@ export function intToBigInt(value: IntegerType, signed: boolean): bigint {
     `Invalid value type. Must be a number, bigint, integer-string, hex-string, BN.js instance, or Buffer.`
   );
 }
+
+export function with0x(value: string): string {
+  return !value.startsWith('0x') ? `0x${value}` : value;
+}
--- a/packages/transactions/tests/builder.test.ts
+++ b/packages/transactions/tests/builder.test.ts
@@ -955,7 +955,7 @@ test('Transaction broadcast success', async () => {
   expect(fetchMock.mock.calls.length).toEqual(1);
   expect(fetchMock.mock.calls[0][0]).toEqual(network.getBroadcastApiUrl());
   expect(fetchMock.mock.calls[0][1]?.body).toEqual(transaction.serialize());
-  expect(response as TxBroadcastResultOk).toEqual('success');
+  expect(response as TxBroadcastResultOk).toEqual({ txid: 'success'});
 });
 
 test('Transaction broadcast with attachment', async () => {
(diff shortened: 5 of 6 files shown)


ahsan-javaid:
cc: @asimm241, @agraebe , @kyranjamie

kyranjamie:
I'm onboard with this change, but it is breaking. Can you add a `BREAKING CHANGE` message to the commit, per https://semantic-release.gitbook.io/semantic-release/#commit-message-format

aulneau (commented):
Thanks this is a long time coming change! I was always confused about the inconsistent returns.

I wonder if it's possible or worth it to type reason_data?

reedrosenbluth (approved):

ahsan-javaid:
> Thanks this is a long time coming change! I was always confused about the inconsistent returns.
> 
> I wonder if it's possible or worth it to type reason_data?

@aulneau reason_data is already a part of `TxBroadcastResultRejected ` type. Here are the types.
```
export type TxBroadcastResultOk = {
  txid: string;
};
export type TxBroadcastResultRejected = {
  error: string;
  reason: TxRejectedReason;
  reason_data: any;
  txid: string;
};
export type TxBroadcastResult = TxBroadcastResultOk | TxBroadcastResultRejected;
```
So now for example if broadcast fails then in the output you will be able to get reason_data. Here is the example output in case of failure:
Output in case of failure:
```
 {
  error: 'transaction rejected',
  reason: 'FeeTooLow',
  reason_data: { actual: 0, expected: 191 },
  txid: '52ced9a7073ad8e19ed25c741c7c0c02a39801b4c565faca4896579f066990df'
}
```

@aulneau @reedrosenbluth Let me know if it makes sense or in case i am misunderstanding anything, do let me know.

ahsan-javaid:
> > Thanks this is a long time coming change! I was always confused about the inconsistent returns.
> > I wonder if it's possible or worth it to type reason_data?
> 
> @aulneau reason_data is already a part of `TxBroadcastResultRejected ` type. Here are the types.
> 
> ```
> export type TxBroadcastResultOk = {
>   txid: string;
> };
> export type TxBroadcastResultRejected = {
>   error: string;
>   reason: TxRejectedReason;
>   reason_data: any;
>   txid: string;
> };
> export type TxBroadcastResult = TxBroadcastResultOk | TxBroadcastResultRejected;
> ```
> 
> So now for example if broadcast fails then in the output you will be able to get reason_data. Here is the example output in case of failure:
> Output in case of failure:
> 
> ```
>  {
>   error: 'transaction rejected',
>   reason: 'FeeTooLow',
>   reason_data: { actual: 0, expected: 191 },
>   txid: '52ced9a7073ad8e19ed25c741c7c0c02a39801b4c565faca4896579f066990df'
> }
> ```
> 
> @aulneau @reedrosenbluth Let me know if it makes sense or in case i am misunderstanding anything, do let me know.

@aulneau Any thoughts on this if it needs further modification?

aulneau:
> > > Thanks this is a long time coming change! I was always confused about the inconsistent returns.
> > > I wonder if it's possible or worth it to type reason_data?
> > 
> > 
> > @aulneau reason_data is already a part of `TxBroadcastResultRejected ` type. Here are the types.
> > ```
> > export type TxBroadcastResultOk = {
> >   txid: string;
> > };
> > export type TxBroadcastResultRejected = {
> >   error: string;
> >   reason: TxRejectedReason;
> >   reason_data: any;
> >   txid: string;
> > };
> > export type TxBroadcastResult = TxBroadcastResultOk | TxBroadcastResultRejected;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So now for example if broadcast fails then in the output you will be able to get reason_data. Here is the example output in case of failure:
> > Output in case of failure:
> > ```
> >  {
> >   error: 'transaction rejected',
> >   reason: 'FeeTooLow',
> >   reason_data: { actual: 0, expected: 191 },
> >   txid: '52ced9a7073ad8e19ed25c741c7c0c02a39801b4c565faca4896579f066990df'
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @aulneau @reedrosenbluth Let me know if it makes sense or in case i am misunderstanding anything, do let me know.
> 
> @aulneau Any thoughts on this if it needs further modification?

Right -- but currently `reason_data` is `any`, I was suggesting we type the possible responses.

zone117x:
I agree with @aulneau, the error data appears to be well defined https://github.com/blockstack/stacks-blockchain/blob/bf4a577113bde7b98c9bffc5e8aa09d19d8946d4/docs/rpc-endpoints.md#post-v2transactions, would be great to have these typed. 

Could be in a separate PR as well?

packages/transactions/src/builders.ts
```diff
@@ -202,9 +204,9 @@ export async function broadcastRawTransaction(
 
   const text = await response.text();
   try {
-    return JSON.parse(text) as TxBroadcastResult;
+    return { txid: JSON.parse(text) } as TxBroadcastResult;
   } catch (e) {
-    return text;
+    return { txid: text } as TxBroadcastResult;
```
zone117x: This is an interesting error condition. Looks like it would only happen on failure to parse JSON, yet this code is only reached on a successful (2xx) http response? If so, it seems like that would indicate a totally incorrect http endpoint is being specified (like something that isn't a stacks-node or API http server), and this should throw similar to the line above, like 
``throw new Error(`Failed to parse broadcasted transaction response ${e}`);``

But if there's an expected reason for this catch condition to be reached that I'm not thinking of, then the error `e` should be included in the return object.

ahsan-javaid: @zone117x This scenario is already handled if the response is not ok at line 204 right after `fetchPrivate` call. The purpose of extra try catch and JSON.parse was to remove the extra double quotes that comes in api response around the txid.  Sample response is here:
```
'"9d82896f72b3af3c0ed9364b22a524597cc3ea662eaf681e4a8f55f2be2a53f3"'
```
You can see the extra double quotes inside the string. So previously JSON.parse was used. I tried to make it more cleaner using replace function to remove the extra quotes. Let me know if it makes sense.

ahsan-javaid:
> I agree with @aulneau, the error data appears to be well defined https://github.com/blockstack/stacks-blockchain/blob/bf4a577113bde7b98c9bffc5e8aa09d19d8946d4/docs/rpc-endpoints.md#post-v2transactions, would be great to have these typed.
> 
> Could be in a separate PR as well?

Done in this PR @zone117x

ahsan-javaid:
> > > > Thanks this is a long time coming change! I was always confused about the inconsistent returns.
> > > > I wonder if it's possible or worth it to type reason_data?
> > > 
> > > 
> > > @aulneau reason_data is already a part of `TxBroadcastResultRejected ` type. Here are the types.
> > > ```
> > > export type TxBroadcastResultOk = {
> > >   txid: string;
> > > };
> > > export type TxBroadcastResultRejected = {
> > >   error: string;
> > >   reason: TxRejectedReason;
> > >   reason_data: any;
> > >   txid: string;
> > > };
> > > export type TxBroadcastResult = TxBroadcastResultOk | TxBroadcastResultRejected;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > So now for example if broadcast fails then in the output you will be able to get reason_data. Here is the example output in case of failure:
> > > Output in case of failure:
> > > ```
> > >  {
> > >   error: 'transaction rejected',
> > >   reason: 'FeeTooLow',
> > >   reason_data: { actual: 0, expected: 191 },
> > >   txid: '52ced9a7073ad8e19ed25c741c7c0c02a39801b4c565faca4896579f066990df'
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > @aulneau @reedrosenbluth Let me know if it makes sense or in case i am misunderstanding anything, do let me know.
> > 
> > 
> > @aulneau Any thoughts on this if it needs further modification?
> 
> Right -- but currently `reason_data` is `any`, I was suggesting we type the possible responses.

@aulneau Thanks for this great suggestion. I have added these changes.

packages/transactions/src/builders.ts
zone117x: Ah I see, looks good to me 👍

packages/transactions/src/builders.ts
```diff
@@ -139,11 +139,20 @@ export async function estimateTransfer(
...
   reason: TxRejectedReason;
-  reason_data: any;
+  reason_data: {
+    message?: string;
+    expected?: string | number;
+    actual?: string | number;
+    is_origin?: boolean;
+    principal?: string;
+    contract_identifier?: string;
+  };
```
zone117x: I think fully typed errors would be a type union, something like:
```ts
interface SerializationRejection {
  reason: 'Serialization';
  reason_data: {
    message: string;
  }
}

interface BadNonceRejection {
  reason: 'BadNonce';
  reason_data: {
    expected: number;
    actual: number;
    is_origin: boolean;
    principal: boolean;
  }
}

type TxBroadcastResultRejected = SerializationRejection | BadNonceRejection | (all the other ones...);
```

ahsan-javaid: @zone117x Added these type changes. Give it a look when you get a chance.

zone117x: Wow, now that's typed! One lasts question around how to easily check if a `TxBroadcastResult` result is success vs an error. For example, looking at the final type union, it doesn't look like it's possible to do simple check without typecasting, something like this doesn't seem like it would work:
```ts
const result: TxBroadcastResult;
if (result.error) {
  throw new Error(...);
}
```

ahsan-javaid: @zone117x Exactly to check the error the result needs to be type casted to TxBroadcastResultRejected
Example: 
```
      transaction, network
  );
  if ((reply as TxBroadcastResultRejected).error) {
   // throw new Error(...);
    console.log('Reason: ', (reply as TxBroadcastResultRejected).reason);
  } else {
    console.log('Txid: ', reply.txid);
  }

```
I spent some time find out some easy way to achieve it in a way as you mentioned above. But eventually we have to typecast it to exact type to check the error or reason property. Other workaround would be be generalize the TxBroadcastResult keeping all the properties of TxBroadcastResultRejected but that might not be a good way. In case you have any suggestion to improve this solution, let me know so that i can incorporate it to make it better.

zone117x: In the past, I've used a `error: null` or `error?: undefined` property on the success object (assuming all the failure types have a `error: <truthy type>` property. That should at least allow easily checking for success. Getting specific error properties will require branching on that `reason` string, no matter what.

ahsan-javaid: @zone117x Let's talk about what's currently have in master branch. Here is the snapshot of types:
```
type TxBroadcastResultOk = string;
type TxBroadcastResultRejected = {
  error: string;
  reason: TxRejectedReason;
  reason_data: any;
  txid: string;
}
type TxBroadcastResult = TxBroadcastResultOk | TxBroadcastResultRejected;
```
You can see here that TxBroadcastResult will be string in case of success. So i am wondering now how you used `error: null` on string type in past.

zone117x: Not in this codebase -- I've used the practice of typing `result` objects with a `interface success { error?: undefined }` (or something to this effect) in the past so that the type union can easily be used.

ahsan-javaid: @zone117x Not sure about the other codebase that you are talking about. If you can mention link to your other codebase or example types to achieve that scenario that would be great. 

@kyranjamie Adding you in this discussion. Let me know if you have any ideas how we can achieve as @zone117x mentioned above.

packages/transactions/src/builders.ts
```diff
@@ -139,13 +139,172 @@ export async function estimateTransfer(
...
+    message: string;
+  }
+  txid: string;
+};
+
+export type TxBroadcastResultOk = {
+  txid: string;
+};
+
+export type TxBroadcastResultRejected =   SerializationRejection |
```
zone117x: @kyranjamie any idea why this is passing lint checks? Locally, vscode inline shows it as a prettier error.

packages/transactions/src/builders.ts
zone117x: Sure, here's an example I just tested and seems to work as expected:
```ts
export type TxBroadcastResultOk = {
  txid: string;
  error?: undefined;
};

let example: TxBroadcastResult;

if (example.error) {
  switch (example.reason) {
    // Example of special handling for a specific error type
    case TxRejectedReason.BadNonce:
      alert(
        `Nonce expected: ${example.reason_data.expected}, actual: ${example.reason_data.actual}`
      );
      throw new Error(example.error);
    default:
      throw new Error(example.error);
  }
} else {
  console.log('success: ' + example.txid);
}
```

packages/transactions/src/builders.ts
kyranjamie: It's an error. It's because it's running from a fork, and the jobs are disabled there, it seems. Or just aren't running at all.

zone117x: @ahsan-javaid can you make sure the lint/formatting is fixed? In the future, you should open PRs from a branch off this repo (you should have maintainer permission for that), and github actions will catch these errors.

ahsan-javaid: @zone117x 
- Alright will make PR's from branch within the repo in future pull requests
- Thanks for the above example. It helped a lot to understand the scenario. I have added the changes to incorporate the error scenario. Now the error can be checked easily with if condition. Great idea :)
```
// This works now.
      transaction, network
  );
  if (reply.error) {
    console.log(reply.reason);
  } else {
    console.log(reply.txid);
  }
``` 
- lint errors are also removed and `npm run lint` passes successfully

zone117x (approved):
LGTM, thanks @ahsan-javaid!

ahsan-javaid:
This is ready for merge. cc: @agraebe @kyranjamie @zone117x

packages/transactions/src/builders.ts
```diff
@@ -201,11 +372,10 @@ export async function broadcastRawTransaction(
...
 
   const text = await response.text();
-  try {
-    return JSON.parse(text) as TxBroadcastResult;
-  } catch (e) {
-    return text;
-  }
+  return {
+    // Replace extra quotes around txid string
+    txid: text.replace(/["]+/g, ''),
```
aulneau: I think there needs to be some validation here. There are times when the response is ok, but the text is not a txid. see https://github.com/blockstack/stacks.js/issues/1054

ahsan-javaid: @aulneau Added the validation on txid. Give it final look for approval or let me know if it needs further changes. cc: @zone117x

packages/transactions/src/utils.ts
```diff
@@ -206,3 +206,11 @@ export const validateStacksAddress = (stacksAddress: string): boolean => {
     return false;
   }
 };
+
+export const validateTxId = (tx_id: string): boolean => {
+  if (tx_id === 'success') return true; // Bypass fetchMock tests
+  const value = !tx_id.startsWith('0x') ? `0x${tx_id}` : tx_id;
+  const regex = /0x[A-Fa-f0-9]{64}/;
+  const matches = regex.exec(value);
+  return matches?.[0] === value;
```
zone117x: Looks like the regex allows capital hexadecimal letters but the final condition here requires the original input casing.

ahsan-javaid: @zone117x Removed the capital hexadecimal letters from the regex. It passes the test cases to detect if there is a error message in txid string or if its a valid txid string. cc: @aulneau

ahsan-javaid:
@zone117x Let me know if this PR needs any improvement or ready for merge.

packages/transactions/src/utils.ts
```diff
@@ -206,3 +206,11 @@ export const validateStacksAddress = (stacksAddress: string): boolean => {
     return false;
   }
 };
+
+export const validateTxId = (tx_id: string): boolean => {
```
aulneau: This function has been updated here https://github.com/blockstack/stacks-wallet-web/blob/a6ac94ab12169d55136349991d135f5ea25f25dc/src/common/validation/validate-tx-id.ts for a different/cleaner implementation. Just a suggestion, both work

ahsan-javaid: Alright i will update this as well.

ahsan-javaid: @aulneau Added these changes.

ahsan-javaid:
@zone117x @aulneau Give a look to this PR if it can be merged or need further changes.

PR #1044, opened 2021-06-24
[Network] Implement StacksNetwork API

### LINKED ISSUE #895: [Network] Propose StacksNetwork API
**Proposal for a different API for networks** (post mainnet)

Currently, if you want to initialise a `StacksNetwork` class with a custom URL, one must first create it, then imperatively mutate the URL. This happens a lot as this interface is a required argument of many functions exposed.

```typescript
const network = new StacksMainnet();
network.coreApiUrl = 'https://...";
```

Mutating properties like this can be error prone, as it can happen at any point synchronously, making the behaviour of async calls unpredictable. Consider this confusing example.

```typescript
const network = new StacksMainnet();
doSomeAsyncCall(() => anotherCallRequiringNetwork({ network })); // uses new url, not default
// tens or hundreds of lines later
network.coreApiUrl = 'https://...'; // now changes target url of 
```

An immutable api mitigates this. 

Further, I think we should be explicit with our default URLs, and avoid defaultly set defaults. It's quite an invisible behaviour.
Setting defaults like this makes the functionality much more transparent.

```typescript
import { StacksMainnet, HIRO_MAINNET_DEFAULT } from '@stacks/network';

const network = new StacksMainnet({ url: HIRO_MAINNET_DEFAULT });
network.coreApiUrl = 'http://'; // Error: Cannot change readonly property
```

asimm241:
## Description
The PR addresses issue #895.
closes #895 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

The `coreApiUrl` in `StacksNetwork , StacksMainnet, StacksTestnet, StacksMocknet`  is now `readonly`.  Now, this field can only be initialized in the constructor. 

## Testing information
added a test case for this scenario 
`cd packages/network && npm run test `

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

@reedrosenbluth @kyranjamie should we include `bnsLookupUrl` as well?

### DIFF
--- a/packages/network/src/index.ts
+++ b/packages/network/src/index.ts
@@ -1,9 +1,17 @@
 import { TransactionVersion, ChainID, fetchPrivate } from '@stacks/common';
 
+export const HIRO_MAINNET_DEFAULT = 'https://stacks-node-api.mainnet.stacks.co';
+export const HIRO_TESTNET_DEFAULT = 'https://stacks-node-api.testnet.stacks.co';
+export const HIRO_MOCKNET_DEFAULT = 'http://localhost:3999';
+
+export interface NetworkUrl {
+  url: string;
+}
+
 export interface StacksNetwork {
   version: TransactionVersion;
   chainId: ChainID;
-  coreApiUrl: string;
+  readonly coreApiUrl: string;
   bnsLookupUrl: string;
   broadcastEndpoint: string;
   transferFeeEstimateEndpoint: string;
@@ -39,13 +47,17 @@ export interface StacksNetwork {
 export class StacksMainnet implements StacksNetwork {
   version = TransactionVersion.Mainnet;
   chainId = ChainID.Mainnet;
-  coreApiUrl = 'https://stacks-node-api.mainnet.stacks.co';
+  readonly coreApiUrl;
   bnsLookupUrl = 'https://stacks-node-api.mainnet.stacks.co';
   broadcastEndpoint = '/v2/transactions';
   transferFeeEstimateEndpoint = '/v2/fees/transfer';
   accountEndpoint = '/v2/accounts';
   contractAbiEndpoint = '/v2/contracts/interface';
   readOnlyFunctionCallEndpoint = '/v2/contracts/call-read';
+
+  constructor(networkUrl: NetworkUrl = { url: HIRO_MAINNET_DEFAULT }) {
+    this.coreApiUrl = networkUrl.url;
+  }
 
   isMainnet = () => this.version === TransactionVersion.Mainnet;
   getBroadcastApiUrl = () => `${this.coreApiUrl}${this.broadcastEndpoint}`;
@@ -99,11 +111,17 @@ export class StacksMainnet implements StacksNetwork {
 export class StacksTestnet extends StacksMainnet implements StacksNetwork {
   version = TransactionVersion.Testnet;
   chainId = ChainID.Testnet;
-  coreApiUrl = 'https://stacks-node-api.testnet.stacks.co';
+
+  constructor(networkUrl: NetworkUrl = { url: HIRO_TESTNET_DEFAULT }) {
+    super(networkUrl);
+  }
 }
 
 export class StacksMocknet extends StacksMainnet implements StacksNetwork {
   version = TransactionVersion.Testnet;
   chainId = ChainID.Testnet;
-  coreApiUrl = 'http://localhost:3999';
+
+  constructor(networkUrl: NetworkUrl = { url: HIRO_MOCKNET_DEFAULT }) {
+    super(networkUrl);
+  }
 }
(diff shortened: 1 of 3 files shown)


packages/network/src/index.ts
```diff
@@ -1,9 +1,17 @@
 import { TransactionVersion, ChainID, fetchPrivate } from '@stacks/common';
 
+export const HIRO_MAINNET_DEFAULT = 'https://stacks-node-api.mainnet.stacks.co';
+export const HIRO_TESTNET_DEFAULT = 'https://stacks-node-api.testnet.stacks.co';
+export const HIRO_MOCKNET_DEFAULT = 'http://localhost:3999';
+
+export interface NetworkUrl {
```
kyranjamie:
```suggestion
export interface NetworkConfig {
```

I think this'd be a better name, as it could contain other properties.

packages/network/tests/network.test.ts
```diff
@@ -1,4 +1,23 @@
...
+  const testnet = new StacksTestnet();
+  expect(testnet.coreApiUrl).toBe(HIRO_TESTNET_DEFAULT);
+
+  const mocknet = new StacksMocknet();
+  expect(mocknet.coreApiUrl).toBe(HIRO_MOCKNET_DEFAULT);
+
+  const customURL = 'customeURL';
+  const customNET = new StacksMainnet({ url: customURL });
+  expect(customNET.coreApiUrl).toBe(customURL);
+});
```
kyranjamie:
```suggestion
describe('Setting coreApiUrl', () => {
  test('it sets mainnet default url', () => {
    const mainnet = new StacksMainnet();
    expect(mainnet.coreApiUrl).toEqual(HIRO_MAINNET_DEFAULT);
  });

  test('it sets testnet url', () => {
    const testnet = new StacksTestnet();
    expect(testnet.coreApiUrl).toEqual(HIRO_TESTNET_DEFAULT);
  });
  test('it sets mocknet url', () => {
    const mocknet = new StacksMocknet();
    expect(mocknet.coreApiUrl).toEqual(HIRO_MOCKNET_DEFAULT);
  });

  test('it sets custom url', () => {
    const customURL = 'customeURL';
    const customNET = new StacksMainnet({ url: customURL });
    expect(customNET.coreApiUrl).toEqual(customURL);
  });

  test('it prevents changing url after initialisation', () => {
    const network = new StacksMainnet({ url: 'https://legiturl.com' });
    // @ts-ignore
    expect(() => network.coreApiUrl = 'https://dodgyurl.com').toThrowError();
    expect(network.coreApiUrl).toEqual('https://legiturl.com')
  })
});

```

packages/network/src/index.ts
```diff
@@ -39,14 +47,18 @@ export interface StacksNetwork {
...
   bnsLookupUrl = 'https://stacks-node-api.mainnet.stacks.co';
   broadcastEndpoint = '/v2/transactions';
   transferFeeEstimateEndpoint = '/v2/fees/transfer';
   accountEndpoint = '/v2/accounts';
   contractAbiEndpoint = '/v2/contracts/interface';
   readOnlyFunctionCallEndpoint = '/v2/contracts/call-read';
 
+  constructor(networkUrl: NetworkUrl = { url: HIRO_MAINNET_DEFAULT }) {
+    this.coreApiUrl = networkUrl.url;
+  }
```
kyranjamie:
```suggestion
  private _coreApiUrl: string;
  get coreApiUrl() {
    return this._coreApiUrl;
  }
  set coreApiUrl(_url: string) {
    throw new Error('Cannot modify property `coreApiUrl` after object initialization');
  }

  constructor(networkUrl: NetworkUrl = { url: HIRO_MAINNET_DEFAULT }) {
    this._coreApiUrl = networkUrl.url;
  }
```

kyranjamie (commented):
Thanks for taking this on @asimm241—

I checked this out and found that `readonly` doesn't have quite the behaviour I expected. It throws a type error only, not a runtime one.

I've suggested some changes that actively prohibit changing the URL after initialisation, and added a test case to cover this.

kyranjamie (approved):
Thanks for updating @asimm241 ✅

asimm241:
@kyranjamie should we add `bnsLookupUrl` to the `Networkconfig` as well?

kyranjamie:
@asimm241 I noticed that too, I think we can probably change it to inherit from the coreApiUrl prop, rather than it being set separately

asimm241:
@kyranjamie 


> @asimm241 I noticed that too, I think we can probably change it to inherit from the coreApiUrl prop, rather than it being set separately

so both the URL would always be the same? is it okay?  sorry, I just want to understand why they were different in the first place.

kyranjamie:
@yknl @aulneau input here? Any reason why `bnsLookupUrl` should differ from `coreApiUrl`?

asimm241:
@kyranjamie should we merge this for now?

kyranjamie:
I think we, can yes. Let's merge.

Issue #981, opened 2021-03-23
[transactions] validateAbi requires same list length


friedger:
**Describe the bug**
When submitting a transaction with flag `validateAbi` the validation fails for lists that do not have the same length as the function parameter

**What version of Blockstack.js are you using?**
stacks/transactions 1.3.3

**Is the bug present in the Blockstack CLI, Gaia hub, Connect or Blockstack Browser?**
neither, it is in stacks/transactions

**To Reproduce**

const tx = await makeContractCall({
    contractAddress: "SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE",
    contractName: "send-many-memo",
    functionName: "send-many",
    functionArgs,
    validateWithAbi: true,
    senderKey,
    network,
  });

**Expected behavior**
Pass validation even if list is shorter than what the function defines as max length (200 in the case of send-many).

PR #1046, opened 2021-06-25
Fix transactions validateAbi requires same list length

ahsan-javaid:
## Description
This PR fixes the issue of transaction with flag validateAbi (true), the validation fails for lists that do not have the same length as the function parameter. This can be reproduced with below snippet:
```
const tx = await makeContractCall({
contractAddress: "SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE",
contractName: "send-many-memo",
functionName: "send-many",
functionArgs,
validateWithAbi: true,
senderKey,
network,
}); 
```

For PR closes the issue #981

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

No

## Testing information
Steps to test:
- Make a contract call with aforementioned snippet by passing less arguments than function expects.
- makeContractCall will not throw a exception and validation will be passed

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/transactions/src/contract-abi.ts
+++ b/packages/transactions/src/contract-abi.ts
@@ -328,7 +328,7 @@ function matchType(cv: ClarityValue, abiType: ClarityAbiType): boolean {
     case ClarityType.List:
       return (
         union.id == ClarityAbiTypeId.ClarityAbiTypeList &&
-        union.type.list.length === cv.list.length &&
+        union.type.list.length >= cv.list.length &&
         cv.list.every(val => matchType(val, union.type.list.type))
       );
     case ClarityType.Tuple:
--- a/packages/transactions/tests/abi.test.ts
+++ b/packages/transactions/tests/abi.test.ts
@@ -60,6 +60,90 @@ test('ABI validation', () => {
   validateContractCall(payload, TEST_ABI);
 });
 
+test('ABI validation should accept list length less than specified in arguments type', () => {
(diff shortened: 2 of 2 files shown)


reedrosenbluth (changes requested):
@ahsan-javaid I think you may have misunderstood this issue. The issue doesn't concern the length of the list of function args, which should always be equal to what is included in the ABI. Rather, the issue is when a Clarity list is passed as an argument to a function. The ABI validation should accept lists that are less than or equal to the max size specified in the arguments type, but currently the validation is checking for equality.

The line that will need to be changed is here: https://github.com/blockstack/stacks.js/blob/4ce29cb46552721dca446fd1e23e011cd0a5ac52/packages/transactions/src/contract-abi.ts#L331

reedrosenbluth (dismissed):
Looks good! Can you include some unit tests in `abi.test.ts` that test for this?

ahsan-javaid:
> Looks good! Can you include some unit tests in `abi.test.ts` that test for this?

@reedrosenbluth unit tests added.

zone117x (approved):
LGTM!

ahsan-javaid:
@zone117x Is this ready for merge. cc: @diwakergupta

diwakergupta (approved):

kyranjamie (approved):

PR #1048, opened 2021-06-29
Transactions: broadcastTransaction to use network set in options #828

### LINKED ISSUE #828: Transactions: broadcastTransaction to use network set in options
The `broadcastTransaction` method requires a transaction and a network param. The transaction is usually created with a network property already, so we could use the network set and avoid requiring a second parameter for the method call.

```
  const txOptions = {
    contractAddress,
    contractName,
    functionName,
    functionArgs,
    senderKey,
    validateWithAbi,
    network, // THIS IS SET
  };

  const transaction = await makeContractCall(txOptions);

  const contractCall = await broadcastTransaction(transaction, network); // SETTING IT AGAIN HERE
  // could be just: broadcastTransaction(transaction)
```

ahsan-javaid:
## Description
The `broadcastTransaction` method requires a transaction and a network param. The transaction is usually created with a network property already, so we could use the network set and avoid requiring a second parameter for the method call.

```
  const txOptions = {
    contractAddress,
    contractName,
    functionName,
    functionArgs,
    senderKey,
    validateWithAbi,
    network, // THIS IS SET
  };

  const transaction = await makeContractCall(txOptions);

  const contractCall = await broadcastTransaction(transaction, network); // SETTING IT AGAIN HERE
  // could be just: broadcastTransaction(transaction)
```

For PR closes the issue #828

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

Yes. `broadcastTransaction` function signature is changed as per aforementioned description. That's why its a breaking change.

No

## Testing information
Steps to test:
- `npm run test` inside transaction package
- Other way of testing this PR is to broadcast a transaction. Now `broadcastTransaction` will only take one parameter which is transaction.

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -585,7 +585,7 @@ async function sendTokens(network: CLINetworkAdapter, args: string[]): Promise<s
     return Promise.resolve(tx.serialize().toString('hex'));
   }
 
-  return broadcastTransaction(tx, txNetwork)
+  return broadcastTransaction(tx)
     .then((response: TxBroadcastResult) => {
       if (response.hasOwnProperty('error')) {
         return response;
@@ -645,7 +645,7 @@ async function contractDeploy(network: CLINetworkAdapter, args: string[]): Promi
     return Promise.resolve(tx.serialize().toString('hex'));
   }
 
-  return broadcastTransaction(tx, txNetwork)
+  return broadcastTransaction(tx)
     .then(response => {
       if (response.hasOwnProperty('error')) {
         return response;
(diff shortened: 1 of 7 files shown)


reedrosenbluth (changes requested):
I don't think it makes sense to include a `network` with a `StacksTransaction` since a `StacksTransaction` can potentially be broadcast to multiple different networks. For example, a `StacksTransaction` constructed with `TransactionVersion.Testnet` can be broadcast on `StacksTestnet` or `StacksMocknet`.

Unfortunately, even though it's redundant to provide the `network` both when constructing the transaction, and when ultimately broadcasting it, I don't think we can get rid of it as an argument to `broadcastTransaction`.

If anything, I think that `network` should be removed from `ContractCallOptions` (and other similar constructor option interfaces) and be replaced by `version: TransactionVersion`.

@zone117x @aulneau @agraebe any thoughts on this?

ahsan-javaid:
@zone117x I would like to know your thoughts on this pull request after @reedrosenbluth comment above in terms of how i should proceed with further changes in this pull request.

reedrosenbluth:
closing this, will create an issue for refactoring txOptions such that it doesn't require the network param

PR #1049, opened 2021-06-30
fix: broken types following diverged network config changes

kyranjamie:
Two changes merged parallel into `master` broke our types. Not sure how this could happen, as we have "branch must be up to date" rule for `master`.

This PR makes @agraebe's regtest changes fit the new format.


beguene (approved):
The format is matching StackMainnet from @stacks/network. Looks good !

PR #1050, opened 2021-06-30
fix: remove console.log

aulneau:
During some test writing in `stacks-wallet-web`, I noticed this was logging. must have accidentally been added and commited.


reedrosenbluth (approved):
Good catch! Thanks!

Issue #784, opened 2020-05-28
encryptECIES: verify that the public key is a secp256k1 point


jcnelson:
The `encryptECIES` method in `src/encryption/ec.ts` does not verify that the given public key string encodes a point on the secp256k1 curve.  While this doesn't pose a security risk in our case -- the worst an attacker can do with this is recover the `ephemeralSK` by using a specially-crafted public key -- it would still be better from a correctness standpoint to verify that the `publicKey` string is indeed a secp256k1 point.  This can be achieved by either checking that `publicKey` is a compresssed secp256k1 public key, or if it isn't, compressing it and decompressing it and verifying that the re-decompressed public key is equal to the original public key (and that the compress and decompress steps don't throw an exception).

jcnelson:
h/t @dantrevino :grin: for sharing: https://github.com/christianlundkvist/blog/blob/master/2020_05_26_secp256k1_twist_attacks/secp256k1_twist_attacks.md

gdstewart:
Gonna tackle this. What should be the desired result be if the key is not valid? Should it throw an exception? Or return something else?

hstove:
Hey @gdstewart , thank you for reaching out and expressing interest. I know it's been a while, but it would be great to hear if you made any progress. As you can probably see, this repo has been refactored into a monorepo with multiple packages. This code is now in the `@stacks/encryption` NPM package, and the code referenced in this issue is now in [`packages/encryption/src/ec.ts`](https://github.com/blockstack/stacks.js/blob/master/packages/encryption/src/ec.ts).

gdstewart:
I have not, but I'll take a look again this coming week when I get a chance

ahsan-javaid:
I am working on this issue. Assigning it to myself.

ahsan-javaid:
@agraebe This needs to be closed as PR is merged.

PR #1052, opened 2021-07-06
EncryptECIES: verify that the public key is a secp256k1 point 

ahsan-javaid:
## Description
This PR fixes the issue to verify that the public key is a secp256k1 point. The `encryptECIES` method in `src/packages/encryption/ec.ts` does not verify that the given public key string encodes a point on the secp256k1 curve. So in the PR i have added a method to validate the public key and added test cases as well. 

For PR closes the issue #784

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No
No

## Testing information
Steps to test:
- npm run test
- Other way of testing this PR is to use the `encryptECIES` method with some bad pubic key and it will reject the public key by throwing an exception.

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/encryption/src/ec.ts
+++ b/packages/encryption/src/ec.ts
@@ -48,6 +48,14 @@ export type SignedCipherObject = {
 /**
  * @ignore
  */
+export enum InvalidPublicKeyReason {
+  InvalidFormat = 'InvalidFormat',
+  IsNotPoint = 'IsNotPoint',
+}
+
+/**
+ * @ignore
+ */
 export async function aes256CbcEncrypt(
   iv: Buffer,
   key: Buffer,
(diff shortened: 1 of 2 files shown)


packages/encryption/src/ec.ts
```diff
@@ -101,6 +101,48 @@ function sharedSecretToKeys(sharedSecret: Buffer): { encryptionKey: Buffer; hmac
...
+  return maybe.match(/^[0-9a-f]+$/i) !== null;
+}
+
+/**
+ * @ignore
+ */
+function isValidPublicKey(pub: string): { result: boolean; reason: string | null } {
+  const invalidFormat = {
+    result: false,
+    reason: 'Invalid public key format',
```
kyranjamie: Maybe it'd be good to use a reason `enum` here, such as

```ts
enum InvalidPublicKeyReason {
  InvalidFormat,
  IsNotPoint
}
```

This'd make it easier to programatically check which one it is. Atm you'd have to compare strings. The text as it is could still be returned, but perhaps as a different property?

kyranjamie (approved):

packages/encryption/src/ec.ts
ahsan-javaid: @kyranjamie Added these changes. Let me know if it's fine or need further improvement.

packages/encryption/src/ec.ts
```diff
@@ -101,6 +109,58 @@ function sharedSecretToKeys(sharedSecret: Buffer): { encryptionKey: Buffer; hmac
...
+  };
+  if (pub.length !== 66 && pub.length !== 130) return invalidFormat;
+
+  const firstByte = pub.slice(0, 2);
+
+  // uncompressed public key
+  if (pub.length === 130 && firstByte !== '04') return invalidFormat;
+
+  // compressed public key
+  if (pub.length === 66 && firstByte !== '02' && firstByte !== '03') return invalidFormat;
```
zone117x, line 145: This looks right to me. Could you add test cases for all these branches? Well known valid keys for each version/prefix, and invalid keys for each?

ahsan-javaid: @zone117x Test cases added for all these branches.

zone117x (approved):

ahsan-javaid:
@zone117x Is this ready for merge. cc: @diwakergupta

PR #1056, opened 2021-07-17
Fix generated explorer URL for faucet command

rswol:
## Description

When running faucet command from cli, the URL generated has an extra "0x"

Example:
```
> stx faucet BLAH -t
{
  "txid": "...",
  "transaction": "https://explorer.stacks.co/txid/0x0x6a4d9...?chain=testnet"
}
ac":"e73..."}
```
## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

no

no

## Testing information
```
> stx faucet ST1HP8ZEGKKRD8P5XS6B5EK0JGE19SG62M5KR0S4Z  -t 
```

## Checklist
- [X] `npm run test` passes
- [] Changelog is updated

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -1673,7 +1673,7 @@ function faucetCall(_: CLINetworkAdapter, args: string[]): Promise<string> {
     .then((faucetTx: any) => {
       return JSONStringify({
         txid: faucetTx.txId!,
-        transaction: generateExplorerTxPageUrl(faucetTx.txId!.substr(2), new StacksTestnet()),
+        transaction: generateExplorerTxPageUrl(faucetTx.txId!.replace(/^0x/, ''), new StacksTestnet()),
       });
     })
     .catch((error: any) => error.toString());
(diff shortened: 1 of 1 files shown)


packages/cli/src/cli.ts
```diff
@@ -1673,7 +1673,7 @@ function faucetCall(_: CLINetworkAdapter, args: string[]): Promise<string> {
     .then((faucetTx: any) => {
       return JSONStringify({
         txid: faucetTx.txId!,
-        transaction: generateExplorerTxPageUrl(faucetTx.txId!, new StacksTestnet()),
+        transaction: generateExplorerTxPageUrl(faucetTx.txId!.substr(2), new StacksTestnet()),
```
zone117x: It would be safer to check for the prefix before trimming, e.g. `faucetTx.txId.replace(/^0x/, '')`

rswol: done!
```Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.```

zone117x (approved):

PR #1058, opened 2021-07-19
v2.0.0

reedrosenbluth:
Version bump for v2


zone117x (commented):
Hmm the changelogs look off, like missing auto-generated changes from 0.1.x to 2.x-beta. I'm not sure how exactly to fix it

reedrosenbluth:
@zone117x can you share an example of one of these changes? I used the `lerna version --conventional-commits` command to auto-generate the changelog. It only includes commits that include semver tags that match the patterns "feature", "fix", "performance improvement" or "breaking changes".

zone117x:
> @zone117x can you share an example of one of these changes? I used the `lerna version --conventional-commits` command to auto-generate the changelog. It only includes commits that include semver tags that match the patterns "feature", "fix", "performance improvement" or "breaking changes".

It looks like the changelogs are missing this range https://github.com/blockstack/stacks.js/compare/v1.4.1...v2.0.0-beta.1

PR #1062, opened 2021-07-20
GH action for generating updated docs on release

### LINKED ISSUE #1047: GH action for generating updated docs on release
There should be a github action in this repo to generate up to date docs whenever a new release is cut

ahsan-javaid:
## Description
There should be a github action in this repo to generate up to date docs whenever a new release is cut

This PR closes #1047

## Type of Change
- [X] New feature
- [ ] Bug fix
- [X] API reference/documentation update
- [ ] Other

No

No

## Testing information

For each new commits added to one of the release branches (for example master, next, beta), with git push or by merging a pull request or merging from another branch, a CI build is triggered and runs the semantic-release command to make a release if there are codebase changes since the last release that affect the package functionalities
Reference: `https://github.com/semantic-release/semantic-release`
## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,43 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - '**'
+    tags-ignore:
+      - '**'
+    paths-ignore:
+      - '**/CHANGELOG.md'
+      - '**/package.json'
+  pull_request:
+  workflow_dispatch:
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set Node Version
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16
+      - name: Install dependencies
+        run: npm i
+
+      - name: Bootstrap
+        run: npm run bootstrap
+
+      - name: Generate docs
+        run: npm run build:docs
+
+      - name: Semantic Release
+        uses: cycjimmy/semantic-release-action@v2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NODE_OPTIONS: --max-old-space-size=4096
+        with:
+          extra_plugins: |
+            @semantic-release/changelog
+            @semantic-release/git
+            @semantic-release/exec
--- a/package.json
+++ b/package.json
@@ -10,6 +10,26 @@
   "workspaces": [
     "packages/**"
   ],
+  "release": {
+    "branches": [
+      "master",
+      "next"
+    ],
+    "plugins": [
+      "@semantic-release/commit-analyzer",
+      "@semantic-release/release-notes-generator",
+      "@semantic-release/changelog",
+      [
+        "@semantic-release/git",
+        {
+          "assets": [
+            "docs"
+          ],
+          "message": "chore(release docs): ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
+        }
+      ]
+    ]
+  },
   "devDependencies": {
     "@stacks/eslint-config": "^1.0.9",
     "@stacks/prettier-config": "0.0.8",
--- a/tsconfig.typedoc.json
+++ b/tsconfig.typedoc.json
@@ -1,7 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
-    "target": "ES2017",
+    "target": "ES2020",
     "module": "ES6"
   },
   "typedocOptions": {


zone117x:
I'd like @CharlieC3 to take a look at this one

CharlieC3:
Thanks @ahsan-javaid, the doc updates are a great start to automated releases!

@zone117x Considering how many packages are being maintained in this repo, it may require a more complex configuration. Besides the doc updates, what else needs to happen during a release for this repo? Some general questions:

1. Do the version properties in the `package.json` files need to be updated for each package during a release?
1. Do the packages need to be versioned separately, or versioned together?
1. Do the packages need to be published to `npmjs.org`?
1. Do Github releases need to be created for new tags? (Usually this is wanted)
1. Does anything else need to happen during a release?

CharlieC3:
I have a meeting with Reed on Friday to discuss the release process reqs for this repo, which may result in changes needed for this PR.

package.json
```diff
@@ -10,6 +10,26 @@
...
+      "@semantic-release/commit-analyzer",
+      "@semantic-release/release-notes-generator",
+      "@semantic-release/changelog",
+      [
+        "@semantic-release/git",
+        {
+          "assets": [
+            "docs"
+          ],
+          "message": "chore(release docs): ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
```
kyranjamie, line 28: Do we want it to skip CI, or should this be removed? We set message like this in the stacks wallet specifically to remove `[skip ci]`

package.json
```diff
@@ -10,6 +10,26 @@
   "workspaces": [
     "packages/**"
   ],
+  "release": {
+    "branches": [
+      "master",
+      "next"
```
CharlieC3, line 16: This repo doesn't have a `next` branch, so this can be dropped. If the `master` branch is the only branch where releases should occur, then the `branches` block can be removed since that's the default behavior.

package.json
CharlieC3: The current message is the default one I believe. So if the default is desired, this `message` property can be removed, but it will skip CI on the subsequent push like Kyran said.

.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: cycjimmy/semantic-release-action@v2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NODE_OPTIONS: --max-old-space-size=4096
+        with:
+          extra_plugins: |
+            @semantic-release/changelog
+            @semantic-release/git
+            @semantic-release/exec
```
CharlieC3, line 43: I think you can remove this `exec` plugin, because I don't see it being used in the `package.json`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Bootstrap
+        run: npm run bootstrap
+
+      - name: Generate docs
+        run: npm run build:docs
+
+      - name: Semantic Release
+        uses: cycjimmy/semantic-release-action@v2
+        env:
```
CharlieC3, line 35: I think you might need the `SEMANTIC_RELEASE_PACKAGE` env var here too. See here: https://github.com/hirosystems/gh-actions-example/blob/main/.github/workflows/ci.yml#L134

CharlieC3 (commented):
Sorry for the late feedback! Also, it's worth mentioning that we have an example repository for setting up `semantic-release`, along with detailed explanations for certain configurations. I also recently updated it, so it should be helpful.
https://github.com/hirosystems/gh-actions-example

Furthermore, this repo is a bit tricky because it's modeled to be a monorepo, and the current configuration is set to only release the root `package.json`, but nothing under the `packages` directory. Getting this to work as desired could be in a separate PR, while this PR could just handle the doc updates. I believe @reedrosenbluth intends to handle this larger work piece.

package.json
ahsan-javaid: The `next` branch is removed. Removing branches block entirely results in error because it is required.
Error message that occurs after removing branches block: 
`A minimum of 1 and a maximum of 3 release branches are required in the branches configuration (https://github.com/semantic-release/semantic-release/blob/master/docs/usage/configuration.md#branches).`
So i think we should keep the branches block with minimum one branch.

.github/workflows/release.yml
ahsan-javaid: Removed `exec` plugin.

.github/workflows/release.yml
ahsan-javaid: Added `SEMANTIC_RELEASE_PACKAGE ` env variable as mentioned.

package.json
ahsan-javaid: Removed `message ` property as mentioned.

kyranjamie (approved):

package.json
```diff
@@ -10,6 +10,24 @@
...
+      [
+        "@semantic-release/git",
+        {
+          "assets": [
+            "docs"
+          ]
+        }
+      ]
+    ]
+  },
```
CharlieC3: I think in this current configuration, `semantic-release` will by default attempt to publish the NPM package found at the root to npmjs.org, a.k.a. [this package](https://github.com/blockstack/stacks.js/blob/master/package.json#L2). If you intend to have that package published on release, you'll need to add more items to the `assets` block to ensure updated files get pushed back to the repo properly after `semantic-release` runs on the GH runner. [See example here.](https://github.com/hirosystems/gh-actions-example/blob/main/package.json#L60-L71)

If you _don't_ want this root package to be published to npmjs.org, you'll need to [add an extra configuration like so](https://github.com/hirosystems/gh-actions-example/blob/main/package.json#L48-L54) (`pkgRoot` can be left out in this case).

Also note, you currently have `docs` listed as an asset to be pushed back up to the `master` branch upon a release. The `docs` dir is not listed as an [ignored path in the workflow file](https://github.com/blockstack/stacks.js/pull/1062/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R9), so pushing changes to this dir on a release will cause another workflow to be ran. Probably harmless, but not sure if that's what you intended. At most it would just be unnecessary to run the workflow again.

.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
...
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
CharlieC3, line 37: This will fail on PRs triggered from forks because forks do not have access to secrets other than `secrets.GITHUB_TOKEN`. You might want to add an `if` conditional to this step to only run on pushes to `master`.

CharlieC3 (changes requested):

package.json
ahsan-javaid: - Added config auto publish false for npm . I would suggest to do it in separate PR once we have detailed requirements as per your questions asked in this thread 
- docs path is ignored

.github/workflows/release.yml
ahsan-javaid: - Bypassed forks
- Added if condition to skip the job

CharlieC3 (approved):
:shipit: Thanks!

ahsan-javaid:
cc: @reedrosenbluth (if ready for merge or need further changes)

ahsan-javaid:
cc: @reedrosenbluth (if ready for merge or need further changes)

reedrosenbluth:
@ahsan-javaid thanks for this! I have one question: does this action also generate changelogs?

ahsan-javaid:
> @ahsan-javaid thanks for this! I have one question: does this action also generate changelogs?

@reedrosenbluth The current configuration does not generate `changelogs`. Let me know if i should add `changelog` option as well ? Would like to know your thoughts (Yes/No) if it needs to be added.

PR #1064, opened 2021-07-26
fix: update readme for broadcastTransaction

### LINKED ISSUE #1063: Update docs with new broadcasting response
With the breaking change introduced [here](https://github.com/blockstack/stacks.js/commit/31055902be7a8123f35c7c81d6c10747b1ceb6ae), we should update the README

@pgray-hiro this would also impact our docs, specifically this page: https://github.com/blockstack/docs/search?q=broadcastTransaction

agraebe:
Fixes https://github.com/blockstack/stacks.js/issues/1063

### DIFF
--- a/packages/transactions/README.md
+++ b/packages/transactions/README.md
@@ -50,8 +50,9 @@ const txOptions = {
   senderKey: 'b244296d5907de9864c0b0d51f98a13c52890be0404e83f273144cd5b9960eed01',
   network,
   memo: 'test memo',
-  fee: 200n, // set a tx fee if you don't want the builder to estimate
-  anchorMode: AnchorMode.Any
+  nonce: new BigNum(0), // set a nonce manually if you don't want builder to fetch from a Stacks node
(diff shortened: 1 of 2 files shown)


packages/transactions/README.md
```diff
@@ -62,7 +62,8 @@ const transaction = await makeSTXTokenTransfer(txOptions);
 const serializedTx = transaction.serialize().toString('hex');
 
 // broadcasting transaction to the specified network
-broadcastTransaction(transaction, network);
+const broadcastResponse = await broadcastTransaction(transaction, network);
+const txId = broadcastResponse.txid;
```
reedrosenbluth, line 65: I think we should omit parsing out the `txId` in this and the following examples

```suggestion
```

reedrosenbluth (approved):
LGTM! left a minor comment

packages/transactions/README.md
agraebe: why? I think it adds valuable info (the return is a JSON object with `txid` as a property)

PR #1067, opened 2021-07-28
Fix coreNode error when fetching public data with no user session 

### LINKED ISSUE #802: Fix coreNode error when fetching public data with no user session
As first encountered with https://github.com/markmhx/humans/issues/12, I'm getting `TypeError: Cannot read property 'coreNode' of undefined` upon calling `getFile` and reaching: https://github.com/blockstack/blockstack.js/blob/950dc4fc838099835ce0592f245453c770663b84/src/storage/index.ts#L253

`sessionData.userData` is undefined and throws an error instead of getting to use `caller.appConfig.coreNode`

cc @zone117x since he appears to be the last person to touch this part of the code

ahsan-javaid:
## Description
This PR fixes the issue of coreNode error when fetching public data with no user session in storage package.

This PR closes the issue #802

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No

oN

## Testing information
- Added test case to verify the coreNode utilized from app config, not from user session as userData is undefined
- `npm run test`

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/storage/tests/storage.test.ts
+++ b/packages/storage/tests/storage.test.ts
@@ -257,6 +257,115 @@ test('getFile unencrypted, unsigned', async () => {
   expect(JSON.parse(<string>file)).toEqual(fileContent);
 });
 
+test('getFile without user session', async () => {
+  const path = 'file.json';
+  const appConfig = new AppConfig(['store_write'], 'http://localhost:8080');
+  const appSpecifiedCoreNode = 'https://app-specified-core-node.local';
+  const userSession = new UserSession({ appConfig });
+  userSession.appConfig.coreNode = appSpecifiedCoreNode;
+  // userData is undefined in userSession so that userSession.appConfig.coreNode is used
+  const nameRecord = {
+    status: 'registered',
+    zonefile:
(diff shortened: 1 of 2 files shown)


packages/storage/tests/storage.test.ts
```diff
@@ -257,6 +257,25 @@ test('getFile unencrypted, unsigned', async () => {
...
+    app: 'http://localhost:8080',
+    decrypt: false,
+  };
+  userSession.appConfig.coreNode = appSpecifiedCoreNode;
+  let storage = new Storage({ userSession });
+  try {
+    //Get file will utilize cordNode url from app config not from userSession
+    await storage.getFile(path, options);
+  } catch {}
+  expect(true).toEqual(true);
```
zone117x: Does this really test the behavior? Seems like this test would pass regardless of the above fix

markmhendrickson:
This PR is great to see! 🙏

packages/storage/tests/storage.test.ts
ahsan-javaid: Test case is updated to test the behaviour.

zone117x (approved):
Thanks @ahsan-javaid!

reedrosenbluth (approved):
Thanks!

PR #1074, opened 2021-08-05
Fixes the incorrect private key derivation path config

andresgalante:
Wallet config private key uses incorrect BIP44 derivation path. BIP44 states `m / purpose' / coin_type' / account' / change / address_index`.

We use here (`m/44/5757'/0'/1`) while it should be `m/44'/5757'/0'/1` (44 should be hardened)

This change will allow us to conform to the standard.

### DIFF
--- a/packages/wallet-sdk/src/derive.ts
+++ b/packages/wallet-sdk/src/derive.ts
@@ -8,7 +8,7 @@ import { Account } from './models/account';
 import { WalletKeys } from './models/wallet';
 
 const DATA_DERIVATION_PATH = `m/888'/0'`;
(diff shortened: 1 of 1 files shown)


aulneau:
Just a note, this needs more work before being merged

agraebe:
if this needs more work, should this be in draft mode? please add reviewers (eg Matt, Reed) once you're ready. thanks!

markmhendrickson:
@andresgalante should we keep this in the backlog or remove until we have clear next steps?

diwakergupta:
@andresgalante this PR is quite stale: unless you're planning to push this forward, I recommend closing it out for now.

andresgalante:
I'll close it until we can come back to this

PR #1078, opened 2021-08-11
feat: construct TransactionSigner with partially signed multi-sig tx

### LINKED ISSUE #1077: Enhance multi-sig ux
It's currently fairly difficult to sign a multi-sig transaction unless using a single instance of `TransactionSigner`. This means that the transaction can't easily be signed by multiple parties. This is the case because the constructor for `TransactionSigner` sets its `sighash` field assuming the transaction hasn't already been signed by anyone.

**Proposal:** upgrade constructor for `TransactionSigner` such that it reconstructs the `sighash` from the signatures that already exist.

_Note: I'm not sure if this is possible..._

reedrosenbluth:
issue: #1077

This PR makes it possible to construct a `TransactionSigner` from a partially signed multi-sig transaction, enabling multiple parties to sign a transaction with different instances of `TransactionSigner`.

### DIFF
--- a/packages/transactions/src/signer.ts
+++ b/packages/transactions/src/signer.ts
@@ -2,7 +2,7 @@ import { StacksTransaction } from './transaction';
 
 import { StacksPrivateKey, StacksPublicKey } from './keys';
 import { isSingleSig, nextVerification, SpendingConditionOpts } from './authorization';
-import { cloneDeep } from './utils';
+import { cloneDeep, txidFromData } from './utils';
 import { AuthType, PubKeyEncoding, StacksMessageType } from './constants';
 import { SigningError } from './errors';
 
@@ -26,10 +26,14 @@ export class TransactionSigner {
     if (spendingCondition && !isSingleSig(spendingCondition)) {
       spendingCondition.fields.forEach(field => {
         if (field.contents.type === StacksMessageType.MessageSignature) {
+          if (!transaction.auth.authType) {
+            throw Error('"transaction.auth.authType" not defined');
+          }
+
           const signature = field.contents;
           const nextVerify = nextVerification(
             this.sigHash,
-            transaction.auth.authType!,
+            transaction.auth.authType,
             spendingCondition!.fee,
             spendingCondition!.nonce,
             PubKeyEncoding.Compressed, // always compressed for multisig
--- a/packages/transactions/src/signer.ts
+++ b/packages/transactions/src/signer.ts
@@ -24,6 +24,11 @@ export class TransactionSigner {
     // auth fields and reconstruct sigHash
     let spendingCondition = transaction.auth.spendingCondition;
     if (spendingCondition && !isSingleSig(spendingCondition)) {
+
+      if (spendingCondition.fields.length >= spendingCondition.signaturesRequired) {
+        throw new Error('SpendingCondition has more signatures than are expected');
+      }
+
       spendingCondition.fields.forEach(field => {
         if (field.contents.type === StacksMessageType.MessageSignature) {
           if (!transaction.auth.authType) {
--- a/packages/transactions/src/signer.ts
+++ b/packages/transactions/src/signer.ts
@@ -1,9 +1,9 @@
 import { StacksTransaction } from './transaction';
 
 import { StacksPrivateKey, StacksPublicKey } from './keys';
-import { isSingleSig, SpendingConditionOpts } from './authorization';
+import { isSingleSig, nextVerification, SpendingConditionOpts } from './authorization';
 import { cloneDeep } from './utils';
-import { AuthType, StacksMessageType } from './constants';
+import { AuthType, PubKeyEncoding, StacksMessageType } from './constants';
 import { SigningError } from './errors';
 
 export class TransactionSigner {
@@ -19,6 +19,26 @@ export class TransactionSigner {
     this.originDone = false;
     this.checkOversign = true;
     this.checkOverlap = true;
+
+    // If multi-sig spending condition exists, iterate over
+    // auth fields and reconstruct sigHash
+    let spendingCondition = transaction.auth.spendingCondition;
+    if (spendingCondition && !isSingleSig(spendingCondition)) {
+      spendingCondition.fields.forEach(field => {
+        if (field.contents.type === StacksMessageType.MessageSignature) {
+          const signature = field.contents;
+          const nextVerify = nextVerification(
+            this.sigHash,
+            transaction.auth.authType!,
+            spendingCondition!.fee,
+            spendingCondition!.nonce,
+            PubKeyEncoding.Compressed, // always compressed for multisig
+            signature,
+          );
+          this.sigHash = nextVerify.nextSigHash;
+        }
+      }); 
+    }
   }
 
   static createSponsorSigner(
--- a/packages/transactions/src/signer.ts
+++ b/packages/transactions/src/signer.ts
@@ -2,7 +2,7 @@ import { StacksTransaction } from './transaction';
 
 import { StacksPrivateKey, StacksPublicKey } from './keys';
 import { isSingleSig, nextVerification, SpendingConditionOpts } from './authorization';
-import { cloneDeep, txidFromData } from './utils';
+import { cloneDeep } from './utils';
 import { AuthType, PubKeyEncoding, StacksMessageType } from './constants';
 import { SigningError } from './errors';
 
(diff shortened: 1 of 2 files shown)


packages/transactions/tests/builder.test.ts
```diff
@@ -250,11 +250,23 @@ test('Make Multi-Sig STX token transfer', async () => {
   expect(deserializedPayload.amount.toString()).toBe(amount.toString());
 
   const signer = new TransactionSigner(deserializedTx);
+  // sign once
```
gregorycoppola: Are you modifying an existing test?


Seems like you should be striclty adding tests unless the old tests are not going to pass anymore, which I doubt/ hope not.

Also, I would guess you should add multiple tests.

packages/transactions/src/signer.ts
```diff
@@ -19,6 +19,26 @@ export class TransactionSigner {
     this.originDone = false;
     this.checkOversign = true;
     this.checkOverlap = true;
+
+    // If multi-sig spending condition exists, iterate over
+    // auth fields and reconstruct sigHash
+    let spendingCondition = transaction.auth.spendingCondition;
```
gregorycoppola, line 25: I'm curious, can this loop ever fail?

gregorycoppola (commented):
Nice idea, @reedrosenbluth .

I had a couple questions...

Also, I wonder if people with more experience on the JS client side have any thoughts.

packages/transactions/src/signer.ts
```diff
@@ -19,6 +19,26 @@ export class TransactionSigner {
...
+    if (spendingCondition && !isSingleSig(spendingCondition)) {
+      spendingCondition.fields.forEach(field => {
+        if (field.contents.type === StacksMessageType.MessageSignature) {
+          const signature = field.contents;
+          const nextVerify = nextVerification(
+            this.sigHash,
+            transaction.auth.authType!,
+            spendingCondition!.fee,
+            spendingCondition!.nonce,
```
zone117x: Maybe better to explicitly check for these optionals and throw a useful error if they are undefined?

packages/transactions/src/signer.ts
```diff
@@ -19,6 +19,26 @@ export class TransactionSigner {
     this.originDone = false;
     this.checkOversign = true;
     this.checkOverlap = true;
+
+    // If multi-sig spending condition exists, iterate over
+    // auth fields and reconstruct sigHash
+    let spendingCondition = transaction.auth.spendingCondition;
+    if (spendingCondition && !isSingleSig(spendingCondition)) {
+      spendingCondition.fields.forEach(field => {
```
zone117x, line 35: Should this have a similar check to the other signing function? like 
```ts
      if (
        spendingCondition.fields.length >= spendingCondition.signaturesRequired
      ) {
        throw new Error('Origin would have too many signatures');
      }
```

reedrosenbluth: Good idea! I wonder if perhaps this check should happen instead (or also) in the `deserializeMultiSigSpendingCondition` function...

packages/transactions/src/signer.ts
reedrosenbluth: Agreed for `authType`, but for the other two, the bang operator is asserting that `spendingCondition` isn't null, which is checked by the `if` on line `26`. I am honestly not sure why these further checks are necessary, but the compiler complains if I remove them.

I really want to refactor `Authorization` so it doesn't have these optional fields.

packages/transactions/src/signer.ts
reedrosenbluth: yes, it can fail if the auth fields are invalid

packages/transactions/src/signer.ts
reedrosenbluth: went ahead and implemented the `Authorization` refactor I mentioned above: https://github.com/blockstack/stacks.js/pull/1082

Was able to get rid of these null checks (in that PR) as a result

packages/transactions/src/signer.ts
gregorycoppola: Why is it `>=`? is the `==` case also an error? Can we test one of these error cases?

packages/transactions/tests/builder.test.ts
```diff
@@ -267,6 +267,90 @@ test('Make Multi-Sig STX token transfer', async () => {
...
+    nonce,
+    memo: memo,
+    numSignatures: 2,
+    publicKeys: pubKeyStrings,
+    anchorMode: AnchorMode.Any
+  });
+
+  const serializedTx = transaction.serialize();
+
+  const tx =
```
gregorycoppola, line 343: I'm curious why we need to test the form of the serialization? is this in case we change the serialization algorithm? It seems to me enough to test the recovered values.

packages/transactions/src/signer.ts
gregorycoppola: Can you test the error cases? And document what errors this constructor will throw in the constructor comment?

On that note, does this codebase have any policy of not failing in a constructor?

packages/transactions/tests/builder.test.ts
```diff
@@ -267,6 +267,90 @@ test('Make Multi-Sig STX token transfer', async () => {
...
+  const partiallySigned = deserializeTransaction(bufferReader2);
+
+  // finish signing with new TransactionSigner
+  const signer2 = new TransactionSigner(partiallySigned);
+  signer2.signOrigin(privKeys[1]);
+  signer2.appendOrigin(pubKeys[2]);
+
+  const serializedSignedTx = partiallySigned.serialize();
+
+  const signedTx =
```
gregorycoppola, line 423: is there some other way to test the success of signing other than checking the serialized output?

I'm guessing you got this output by just printing the result that you got, so none of us human readers can know if this is really expected or not.

gregorycoppola (commented):
Hey Reed.. Nice test case added. I had a few questions about the error cases and the testing of them.

packages/transactions/tests/builder.test.ts
zone117x: I think it's helpful to check output blob, but agreed with @gregorycoppola, would be helpful if individual properties relevant to signing could be explicitly checked (I think you could test the public keys, ordering, and count?)

packages/transactions/tests/builder.test.ts
kantai: It is generally pretty helpful to have these kinds of explicit vector tests. The benefits are basically:

1. They provide vectors to test other implementations. Imagine a hardware wallet implementation that can't use the TS/JS builds.
2. The tests become something of a canary for non-backwards-compatible changes in the serialization format.

packages/transactions/tests/builder.test.ts
kantai: +1 -- the test should check both against a blob, and against the reconstructed object properties.

packages/transactions/src/signer.ts
```diff
@@ -19,6 +19,35 @@ export class TransactionSigner {
...
     this.checkOverlap = true;
+
+    // If multi-sig spending condition exists, iterate over
+    // auth fields and reconstruct sigHash
+    let spendingCondition = transaction.auth.spendingCondition;
+    if (spendingCondition && !isSingleSig(spendingCondition)) {
+
+      if (spendingCondition.fields.length >= spendingCondition.signaturesRequired) {
+        throw new Error('SpendingCondition has more signatures than are expected');
+      }
```
kantai: I'd be careful with this check -- the fix from #1081 would also need to be applied here.

kantai (commented):
This LGTM, @reedrosenbluth, but I had some comments about the signature over-sign check and the test cases.

packages/transactions/src/signer.ts
zone117x: > And document what errors this constructor will throw in the constructor comment

This is not very common in js. I'd consider it a nice-to-have.

> On that note, does this codebase have any policy of not failing in a constructor?

No.

packages/transactions/src/signer.ts
```diff
@@ -19,6 +19,38 @@ export class TransactionSigner {
...
+          if (!transaction.auth.authType) {
+            throw Error('"transaction.auth.authType" not defined');
+          }
+
+          const signature = field.contents;
+          const nextVerify = nextVerification(
+            this.sigHash,
+            transaction.auth.authType,
+            spendingCondition!.fee,
+            spendingCondition!.nonce,
```
zone117x, line 46: Are these force assertions still needed? If so, should this instead use some undefined checks and explicit errors instead?

reedrosenbluth: these are removed in the followup PR that I have open (the Authorization refactor)

zone117x: Ah thats right, thanks!

zone117x (approved):
LGTM!

PR #1081, opened 2021-08-12
Improve signOrigin oversign check.

### LINKED ISSUE #1080: Appending pubs before Signing returns an error for multi sig
Describe the bug
When submitting a P2SH multisig transaction to the mainnet/testnet and trying to append before signing you get an error of "Origin would have too many signatures" which is incorrect and its caused by this line https://github.com/blockstack/stacks.js/blob/master/packages/transactions/src/signer.ts#L59
It should only count the one that are type signature

Code to Reproduce

```javascript
import {
  TransactionSigner,
  getPublicKey,
  createStacksPrivateKey,
  publicKeyToString,
  standardPrincipalCV,
  broadcastTransaction,
  makeUnsignedSTXTokenTransfer,
import BigNum from 'bn.js';

// Private key strings with compression byte appended
const privKeyStrings = [
];

// Compressed public keys
const pubKeys = privKeys.map(getPublicKey);

const recipient = standardPrincipalCV('SN3C9PPBTKJNWJ5HZBBWERRAF4E3W6M39AN54S92');
const amount = new BigNum(74);
const fee = new BigNum(278);
const nonce = new BigNum(0);
const memo = 'test';

(async () => {
    recipient,
    amount,
    network,
    fee,
    nonce,
    memo,
    numSignatures: 2,
    publicKeys: pubKeyStrings,
  });

  signer.signOrigin(privKeys[0]);
  signer.appendOrigin(pubKeys[1]);
  

  const txId = await broadcastTransaction(transaction, network);
  console.log('Transaction ID: ' + JSON.stringify(txId, null, 2).replace(/^"|"$/, ''));
})().catch((err) => {
  console.error('Error:' + err);
  process.exit(1);
});
```

Expected Behavior
Sign or Append in any order as long  number of signatures are right

alebusse:
Let's appendOrigin works before signing . Fields which are not a MeesageSignature don't count in Overlap check

issue: https://github.com/blockstack/stacks.js/issues/1080

### DIFF
--- a/packages/transactions/src/signer.ts
+++ b/packages/transactions/src/signer.ts
@@ -1,4 +1,4 @@
-import { StacksTransaction } from './transaction';
+import { StacksTransaction, StacksMessageType } from './transaction';
 
 import { StacksPrivateKey, StacksPublicKey } from './keys';
 import { isSingleSig, SpendingConditionOpts } from './authorization';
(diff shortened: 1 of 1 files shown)


kantai (approved):
LGTM! Thanks @alebusse

PR #1082, opened 2021-08-12
Refactor: change Authorization from a class to functional interface

reedrosenbluth:
Refactor `Authorization` class to be a more functional typescript interface without nullable fields.

This is a BREAKING CHANGE

### DIFF
--- a/packages/transactions/src/authorization.ts
+++ b/packages/transactions/src/authorization.ts
@@ -559,14 +559,20 @@ export function verifyOrigin(auth: Authorization, initialSigHash: string): strin
   }
 }
 
-export function setFee(auth: Authorization, amount: IntegerType) {
+export function setFee(auth: Authorization, amount: IntegerType): Authorization {
   switch (auth.authType) {
     case AuthType.Standard:
-      auth.spendingCondition.fee = intToBigInt(amount, false);
-      break;
+      let spendingCondition = {
+        ...auth.spendingCondition,
+        fee: intToBigInt(amount, false),
+      };
+      return { ...auth, spendingCondition };
     case AuthType.Sponsored:
-      auth.sponsorSpendingCondition.fee = intToBigInt(amount, false);
-      break;
+      let sponsorSpendingCondition = {
+        ...auth.sponsorSpendingCondition,
+        fee: intToBigInt(amount, false),
+      };
+      return { ...auth, sponsorSpendingCondition };
   }
 }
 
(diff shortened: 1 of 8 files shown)


packages/transactions/src/authorization.ts
```diff
@@ -503,177 +493,141 @@ function verifySingleSig(
...
+    spendingCondition,
+  };
+}
+
+export function createSponsoredAuth(
+  spendingCondition: SpendingCondition,
+  sponsorSpendingCondition?: SpendingCondition
+): Authorization {
+  if (!sponsorSpendingCondition) {
+    sponsorSpendingCondition = createSingleSigSpendingCondition(
```
kyranjamie: I'd write this in a way where you don't reassign the parameter. Perhaps do this conditional as a ternary in the return object itself?

packages/transactions/src/authorization.ts
```diff
@@ -503,177 +493,141 @@ function verifySingleSig(
...
 
-  serialize(): Buffer {
-    const bufferArray: BufferArray = new BufferArray();
-    if (this.authType === undefined) {
-      throw new SerializationError('"authType" is undefined');
-    }
-    bufferArray.appendByte(this.authType);
+export function setNonce(auth: Authorization, nonce: IntegerType) {
+  auth.spendingCondition.nonce = intToBigInt(nonce, false);
+}
```
kyranjamie: Can this be done in an immutable fashion? If it must mutate the value directly, I'd add warning with a tsdoc comment.

kyranjamie (approved):

packages/transactions/src/transaction.ts
```diff
@@ -102,13 +110,14 @@ export class StacksTransaction {
     if (!tx.auth.spendingCondition) {
       throw new Error('Cannot set signature on transaction without spending condition');
     }
-    (tx.auth.spendingCondition as SingleSigSpendingCondition).signature =
-      createMessageSignature(parsedSig);
+    (tx.auth.spendingCondition as SingleSigSpendingCondition).signature = createMessageSignature(
```
zone117x: Rather than casting (`as SingleSigSpendingCondition`) which can potentially be type-unsafe, is there a runtime type guard that would inform typescript that this `tx.auth.spendingCondition` is already an instance of `SingleSigSpendingCondition`?

reedrosenbluth: this function is only used in a test, so I'm actually going to move it there, in addition to making the suggested change

zone117x (approved):
Other than the CI issues, LGTM!

kyranjamie (commented):
What does the `!` in front of your commit messages mean?

Doesn't a breaking change require `BREAKING CHANGE` on a new line of one of the commit messages, per conventional commits/semantic release?

reedrosenbluth:
@kyranjamie according to https://www.conventionalcommits.org/en/v1.0.0/ it's another way to indicate a breaking change.

agraebe:
fixing the CI issue is the last step here, right @reedrosenbluth ?

PR #1095, opened 2021-08-25
Export getUserAppFileUrl

### LINKED ISSUE #1019: Export getUserAppFileUrl
**Describe the feature**

As we get user-owned storage back, it is important to lookup the read url from the profile. This is what `getUserAppFileUrl` does. This function should be exported by `@stacks/storage`

ahsan-javaid:
## Description
As we get user-owned storage back, it is important to lookup the read url from the profile. This is what `getUserAppFileUrl` does. This function should be exported by `@stacks/storage`

This PR closes #1019

## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Other

No

No

## Testing information

1. Execute `npm run test` within `storage` package
2. Use below snippet to test the functionality
```
  import { getUserAppFileUrl } from '@stacks/storage';
  const zoneFileLookupURL = 'https://stacks-node-api.mainnet.stacks.co/v1/names';
  const path = 'file.json';
  const username = 'yukan.id';
  const appOrigin = 'http://localhost:8080';
  const url = await getUserAppFileUrl({ 
    path,
    username,
    appOrigin,
    zoneFileLookupURL
  });
  console.log(url);
```

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/storage/src/storage.ts
+++ b/packages/storage/src/storage.ts
@@ -833,6 +833,21 @@ export class Storage {
   }
 }
 
+type FileUrlOptions = {
+  path: string;
+  username: string;
+  appOrigin: string;
+  zoneFileLookupURL: string;
+};
+export function getUserAppFileUrl(options: FileUrlOptions): Promise<string | undefined> {
+  return new Storage({}).getUserAppFileUrl(
+    options.path,
+    options.username,
+    options.appOrigin,
+    options.zoneFileLookupURL
+  );
+}
+
 /**
  * @param {Object} [options=null] - options object
  * @param {String} options.username - the Blockstack ID to lookup for multi-player storage
--- a/packages/storage/tests/storage.test.ts
+++ b/packages/storage/tests/storage.test.ts
@@ -7,12 +7,13 @@ import {
   getBucketUrl,
   deleteFromGaiaHub,
   GaiaHubConfig,
+  getUserAppFileUrl,
 } from '../src';
 
 import { Storage } from '../src';
 
 import { UserSession, AppConfig, UserData } from '@stacks/auth';
-import { DoesNotExist, getAesCbcOutputLength, getBase64OutputLength } from '@stacks/common';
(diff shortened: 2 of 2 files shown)


packages/storage/src/storage.ts
```diff
@@ -833,6 +833,26 @@ export class Storage {
...
+ * blockstack.js's [[getNameInfo]] function instead.
+ * @return {Promise<string>} that resolves to the public read URL of the file
+ * or rejects with an error
+ */
+export function getUserAppFileUrl(
+  path: string,
+  username: string,
+  appOrigin: string,
+  zoneFileLookupURL?: string
+): Promise<string | undefined> {
```
kyranjamie: Would promote using an object as the argument when requiring this many parameters.

ahsan-javaid: Added these changes.

packages/storage/tests/storage.test.ts
```diff
@@ -2118,6 +2119,104 @@ test('getUserAppFileUrl', async () => {
...
+            }
+          }
+        },
+        'signature': 'UyQNZ02kBFHEovbwiGaS-VQd57w9kcwn1Nt3QbW3afEMArg1OndmeplB7lzjMuRCLAi-88lkpQLkFw7LwKZ31Q'
+      }
+    }
+  ]
+  /* eslint-enable */
+
+  fetchMock.once(JSON.stringify(nameRecord)).once(JSON.stringify(profileContent));
```
zone117x: Can this test confirm that the expected URLs are being fetched, maybe by specifying it in this fetch mock?
And URLs like `zoneFileLookupURL` and `nameRecord.zonefile`  should be specified as something other than the real endpoints.

ahsan-javaid: Added these changes.

kyranjamie (approved):

zone117x (approved):

PR #1101, opened 2021-09-06
CLI add missing clarity types in contract call

### LINKED ISSUE #1100: Stacks CLI only supports limited subset of Clarity types for contract function calls
While developing an example application for the Stacks docs, I wanted to demonstrate to a developer how to use the Stacks CLI to call a smart contract function on the local Clarinet DevNet. The function I was trying to call requires a `string-utf8` argument, which the Stacks CLI apparently does not support.

Talking with @zone117x, it appears that the Stacks CLI only supports `int`, `uint`, `bool`, `principal`, and `buffer`. Relevant source can be found [here](https://github.com/blockstack/stacks.js/blob/6d9ff4df00469bb1015e031bb04cc3423357de5b/packages/cli/src/utils.ts#L805-L840).

Ideally, the Stacks CLI should be updated to support all Clarity types. At the minimum, it would be nice to get string types supported immediately, and add other types in the future.

ahsan-javaid:
## Description
This PR adds support for missing clarity types `string-ascii`, `string-utf8` and `optional` types in cli for contract calls.

For details refer to issue #1100

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No

No
## Testing information

1. `npm run test`
2. Calling `call_contract_func` command using cli will allow the passing of string and optional types without any errors.

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -1964,5 +1964,6 @@ export const testables =
   process.env.NODE_ENV === 'test'
     ? {
         addressConvert,
+        contractFunctionCall,
       }
     : undefined;
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -11,6 +11,8 @@ import {
   getTypeString,
   ClarityAbiType,
   isClarityAbiPrimitive,
+  isClarityAbiStringAscii,
+  isClarityAbiStringUtf8,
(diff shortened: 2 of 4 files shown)


zone117x (approved):
LGTM!

reedrosenbluth (approved):

PR #1102, opened 2021-09-21
Some etags in sessionData might not be deleted in deleteFile

### LINKED ISSUE #888: Some etags in sessionData might not be deleted in deleteFile
**Describe the bug**
When delete multiple files at once, some etags in sessionData might not be deleted. As await is not guaranteed ordering, there might be a race condition. For example, deleting 2 files, calling deleteFile('path1') in the first line and calling deleteFile('path2') in the second line. At storage.ts:line 637, the first call will get a copied of sessionData with etags: ['path1', 'path2'] and send a request to the server. With await keyword, this is queued up and the second call is processed. The second call will get a copied of sessionData with etags: ['path1', 'path2'] and send a request to the server. When the request1 get responded, 'path1' is deleted from the copied and the rest: ['path2'] get saved to sessionData. And when the request2 get responded, 'path2' is deleted from the copied and the rest: ['path1'] get saved to sessionData.

I think the way to fix this is to get a copied of sessionData after awaiting.

Also, this might happen at storage.ts:line 555 and 599 as well. At that time, the copied of sessionData and the real sessionData might be different and some etags can be missing due to overriding of the obsolete copied.

I found this bug in 21.1.1 and I believe it also happens in the latest version

I think not

**To Reproduce**
Delete 2 files with deleteFile('path1') and deleteFile('path2')

**Expected behavior**
Etags in sessionData should be deleted

 - OS: [e.g. iOS] All
 - Browser [e.g. chrome, safari] All
 - Version [e.g. 22] found in 21.1.1 and believe also in the latest version

ahsan-javaid:
## Description
This PR fixes the issue: `Some etags in sessionData might not be deleted in deleteFile`

This PR closes the issue #888

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No
## Testing information
* `npm run test` in `storage`
* `Use the below attached snippet to test the behaviour`
```
import { UserSession, AppConfig, LOCALSTORAGE_SESSION_KEY } from './packages/auth';
import { Storage } from '@stacks/storage';
import * as jsdom from 'jsdom';

void(async () => {
  const dom = new jsdom.JSDOM('', { url: 'https://example.org/' }).window;
  const globalAPIs: { [key: string]: any } = {
    localStorage: dom.localStorage,
    location: dom.location,
    self: dom,
  };
  for (const globalAPI of Object.keys(globalAPIs)) {
    (global as any)[globalAPI] = globalAPIs[globalAPI];
  }
  const privateKey = '896adae13a1bf88db0b2ec94339b62382ec6f34cd7e2ff8abae7ec271e05f9d8';
  const appConfig = new AppConfig();
  const userSession = new UserSession({ appConfig });
  const session = userSession.store.getSessionData();
  session.userData = <any> {
    appPrivateKey: privateKey,
  };
  userSession.store.setSessionData(session);
  const storage = new Storage({ userSession });

  const files = ['aa.json', 'bb.json', 'cc.json', 'dd.json', 'ee.json'];

  for (let i = 0; i < files.length; i++) {
    const myData = JSON.stringify({
      // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
      hello: 'hello: ' + i,
      num: i,
    });
    await storage.putFile(files[i], myData, { encrypt: false });
  }
  const promises = [];
  for (let i = 0; i < files.length; i++) {
    promises.push(storage.deleteFile(files[i]));
  }
  Promise.all(promises).then(() => {
    const se = userSession.store.getSessionData();
    console.log('Final session', se);
    console.log('localStorage: ', localStorage.getItem(LOCALSTORAGE_SESSION_KEY));
  });
})();


```

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/storage/src/storage.ts
+++ b/packages/storage/src/storage.ts
@@ -636,29 +636,32 @@ export class Storage {
   ) {
     const gaiaHubConfig = await this.getOrSetLocalGaiaHubConnection();
     const opts = Object.assign({}, options);
-    const sessionData = this.userSession.store.getSessionData();
     if (opts.wasSigned) {
       // If signed, delete both the content file and the .sig file
       try {
         await deleteFromGaiaHub(path, gaiaHubConfig);
         await deleteFromGaiaHub(`${path}${SIGNATURE_FILE_SUFFIX}`, gaiaHubConfig);
+        const sessionData = this.userSession.store.getSessionData();
         delete sessionData.etags![path];
         this.userSession.store.setSessionData(sessionData);
       } catch (error) {
         const freshHubConfig = await this.setLocalGaiaHubConnection();
         await deleteFromGaiaHub(path, freshHubConfig);
         await deleteFromGaiaHub(`${path}${SIGNATURE_FILE_SUFFIX}`, gaiaHubConfig);
+        const sessionData = this.userSession.store.getSessionData();
         delete sessionData.etags![path];
         this.userSession.store.setSessionData(sessionData);
       }
     } else {
       try {
         await deleteFromGaiaHub(path, gaiaHubConfig);
+        const sessionData = this.userSession.store.getSessionData();
         delete sessionData.etags![path];
         this.userSession.store.setSessionData(sessionData);
       } catch (error) {
         const freshHubConfig = await this.setLocalGaiaHubConnection();
         await deleteFromGaiaHub(path, freshHubConfig);
+        const sessionData = this.userSession.store.getSessionData();
         delete sessionData.etags![path];
         this.userSession.store.setSessionData(sessionData);
       }
--- a/packages/storage/tests/storage.test.ts
+++ b/packages/storage/tests/storage.test.ts
@@ -12,7 +12,7 @@ import {
 
 import { Storage } from '../src';
 
-import { UserSession, AppConfig, UserData } from '@stacks/auth';
+import { UserSession, AppConfig, UserData, LOCALSTORAGE_SESSION_KEY } from '@stacks/auth';
 import { DoesNotExist, getAesCbcOutputLength, getBase64OutputLength, fetchPrivate } from '@stacks/common';
 import { StacksMainnet } from '@stacks/network';
 import * as util from 'util';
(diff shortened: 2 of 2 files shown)


packages/auth/src/sessionStore.ts
```diff
@@ -93,17 +96,22 @@ export class LocalStorageStore extends SessionDataStore {
       throw new NoSessionDataError('No session data was found in localStorage');
     }
     const dataJSON = JSON.parse(data);
-    return SessionData.fromJSON(dataJSON);
+    const session = SessionData.fromJSON(dataJSON);
+    Object.assign(this.sessionData, session);
```
kyranjamie: Is it necessary to do this object assign here? Seem unusual that this function would have mutation/side effects.

packages/auth/src/sessionStore.ts
```diff
@@ -66,6 +66,9 @@ export class InstanceDataStore extends SessionDataStore {
  */
 export class LocalStorageStore extends SessionDataStore {
   key: string;
+  // sessionData points to localStorage instance for sharing session between parallel calls of put or delete file
+  // To prevent the conflicting issue of etags deletion or updation
+  sessionData: SessionData = new SessionData({});
```
zone117x: It's not clear to me that storing/caching more state in memory is the solution here, seems like it would lead to more kinds of race conditions.

It makes more sense to me that the functions modifying session data (like mutating etags) should have fixes to avoid caching and writing old `SessionData` between network calls. E.g. the deleteFile function should get the latest session data before updating and writing:
```ts
await deleteFromGaiaHub(path, gaiaHubConfig);
// get latest session data which could have been modified while the above `await` was resolving
const sessionData = this.userSession.store.getSessionData();
delete sessionData.etags![path];
this.userSession.store.setSessionData(sessionData);
```

Any other functions that modify `sessionData.etags` should be looked at as well for similar race conditions.

ahsan-javaid: The `deleteFile` function call is already getting the latest `session ` values. I don't think we can even touch the deleteFile function to solve this issue. It's a concurrency issue that only happens when you call deleteFile without await statement and in my opinion making session a shared variable solves the problem. And i also provided `script` in the description of this pr to prove the correct behaviour.

Secondly other functions that are modifying the etags are transparent similar to `deleteFile` as i did not make a change in `deleteFile` function.  Hope it make sense.
cc: @zone117x

packages/auth/src/sessionStore.ts
ahsan-javaid: The objective is to keep the same reference to make it a shared variable as similar to `InstanceDataStore`. So that between the parallel calls to deleteFile should be able to get same reference to session variable. 
I also mentioned the `testing` script that proves the correct behaviour. 
cc: @kyranjamie

packages/auth/src/sessionStore.ts
zone117x: > The deleteFile function call is already getting the latest session values

That isn't correct, here's the code:
https://github.com/blockstack/stacks.js/blob/9816b3c7ae4e9a8f7fcff219239f8b6952f70471/packages/storage/src/storage.ts#L656-L658

ahsan-javaid: @zone117x  Yeah that's what i fixed in this PR.

zone117x: I'd just re-iterate my first comment above that I don't think this is a good approach, and that a more robust fix would be ensuring the localStorage `sessionData` is retrieved, updated, and saved _after_ the async network calls.

ahsan-javaid: @zone117x I did full research in terms of approach to get this issue fixed. You may mention any other good approach if you would like. Secondly i would like you to reconsider the behaviour in this issue. The user is facing this issue because he want to deleteFiles in a parallel fashion without `await` statement. Considering that the best approach is to make a shared variable that will remain in `sync` with localstorage all time, that's what i did to solve the issue. I also provided the `testcases` and `scripts` to verify the results.

zone117x: But it won't remain in sync at all times. Create multiple instances of `userSession`, or open an app in multiple tabs, etc. There are many ways caching the localStorage data in memory like this is can create more issues. 

The solution I suggested seems pretty straightforward.

ahsan-javaid: @zone117x Alright
- Would you mind sharing many ways of caching localStorage data in memory
- Can you produce what issues this solution will create
- Can you share the solution that you mentioned as pretty straightforward 

I am ready to implement if there are better solutions. Please guide me in the right direction.

zone117x: Yes, test something like the code snippet I shared in my first comment:
```ts
await deleteFromGaiaHub(path, gaiaHubConfig);
// get latest session data which could have been modified while the above `await` was resolving
const sessionData = this.userSession.store.getSessionData();
delete sessionData.etags![path];
this.userSession.store.setSessionData(sessionData);
```

ahsan-javaid: Tested and added this approach. Thanks.

zone117x: Looks good, thanks!

zone117x (approved):
LGTM. It seems like a similar issue would happen with concurrent calls to `putFile`, although that could be addressed in a separate PR.

reedrosenbluth (approved):

PR #1106, opened 2021-09-28
Domain names lost capacity to access remote json file from GAIA storage

### LINKED ISSUE #1086: Domain names lost capacity to access remote json file from GAIA storage
**Describe the bug**
In the legacy version for Storage (GAIA) it was possible to access a json file from a remote user using the username.  Now using domain names does not have the capacity to read the GAIA space.

Latest

**To Reproduce**
1. Before in the legacy version
imagen
import { UserSession } from '@stacks/auth';
import { Storage } from '@stacks/storage'
….
this.state = {  userSession: new UserSession()}
….
const { userSession } = this.state
….
const storage = new Storage({ userSession });
const options = { username: userremote.id, decrypt: false, verify: false }
storage.getFile(file.json`, options)
          .then((fileContents) => {console.log(‘Good’)}
One specified the username in options and manage to get the file.

2. Now
imagen

import { UserSession } from '@stacks/auth';
import { Storage } from '@stacks/storage'
….
this.state = {  userSession: new UserSession()}
….
const { userSession } = this.state
….
const storage = new Storage({ userSession });
const options = { username: userremote.id.blockstack, decrypt: false, verify: false }
storage.getFile(file.json`, options)
          .catch((err) => {console.log(‘Bad)}

Using the new domain name the he username specified in options is not anymore in the zone file, goes as an error through the catch.
Now the name instead of username is located in a different place, retrievable by the stacks address associate to the name, so the storage cannot get the file.

**Expected behavior**
The storage should be able to get a file from a remote user.

ahsan-javaid:
## Description
This PR fixes the issue #1086. 

1. The main issue was in `resolveZoneFileToProfile` function which was unable to detect the profile url from zonefile field. I have fixed this issue by adding a check if zonefile is a valid profile url then use this url to fetch the profile. 

2. Second change is in the `lookupProfile` function call to skip the `verifyProfileToken` flow because in case of subdomain 
the address received from `zonefile` and `issuer public` key can never match. 

Code snippet to reproduce the issue:
```typescript
import { UserSession, AppConfig } from '@stacks/auth';
import { Storage } from '@stacks/storage';

void(async () => {
  const appConfig = new AppConfig();
  const userSession = new UserSession({ appConfig });
  const storage = new Storage({ userSession });
  const options = {
    username: 'jfontirroig.xck.app',
    app: 'https://xck.app',
    decrypt: false,
    verify: false,
  };
  const fileContent = await storage.getFile('contractlist.json', options);
  console.log(fileContent);

})();

```

This PR closes the issue #1086

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No
## Testing information
* `npm run test` in `storage`
* `Use the below attached snippet to test the behaviour`
```
import { UserSession, AppConfig } from '@stacks/auth';
import { Storage } from '@stacks/storage';

void(async () => {
  const appConfig = new AppConfig();
  const userSession = new UserSession({ appConfig });
  const storage = new Storage({ userSession });
  const options = {
    username: 'jfontirroig.xck.app',
    app: 'https://xck.app',
    decrypt: false,
    verify: false,
  };
  const fileContent = await storage.getFile('contractlist.json', options);
  console.log(fileContent);

})();

```
* Running above script should fetch the file contents
## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/profile/src/profile.ts
+++ b/packages/profile/src/profile.ts
@@ -338,10 +338,17 @@ export function resolveZoneFileToProfile(zoneFile: any, publicKeyOrAddress: stri
     } catch (e) {
       reject(e);
     }
-
+    let isProfileUrl: boolean = false;
+    try {
+      // Check if zoneFile is only a profile url as in case of subdomains it can be be a url instead of zoneFile object
+      new URL(zoneFile);
+      isProfileUrl = true;
+    } catch (_) {}
     let tokenFileUrl: string | null = null;
     if (zoneFileJson && Object.keys(zoneFileJson).length > 0) {
       tokenFileUrl = getTokenFileUrl(zoneFileJson);
+    } else if (isProfileUrl) {
+      tokenFileUrl = zoneFile;
     } else {
       let profile = null;
       try {
--- a/packages/auth/src/profile.ts
+++ b/packages/auth/src/profile.ts
@@ -31,7 +31,8 @@ export function lookupProfile(options: ProfileLookupOptions): Promise<Record<str
   }
   return lookupPromise.then((responseJSON: any) => {
     if (responseJSON.hasOwnProperty('zonefile') && responseJSON.hasOwnProperty('address')) {
-      return resolveZoneFileToProfile(responseJSON.zonefile, responseJSON.address);
+      const address = responseJSON.status === 'registered_subdomain' ? '' : responseJSON.address;
+      return resolveZoneFileToProfile(responseJSON.zonefile, address);
     } else {
       throw new Error(
         'Invalid zonefile lookup response: did not contain `address`' + ' or `zonefile` field'
--- a/packages/storage/tests/storage.test.ts
+++ b/packages/storage/tests/storage.test.ts
@@ -367,6 +367,90 @@ test('getFile without user session', async () => {
   });
 });
 
+test('getFile username registered as subdomain', async () => {
+  const path = 'file.json';
+  const appConfig = new AppConfig();
+  const userSession = new UserSession({ appConfig });
+  const nameRecord = {
+    address: 'SP31C1RGYVJVWNYWJNRVSZQA2XN9HDEA08FB0G03S',
+    blockchain: 'stacks',
+    last_txid: '0x0b80143fae8a52f47c970c5cf935ae37cd13236887e2a470d8a6891091219986',
+    resolver: '',
+    status: 'registered_subdomain',
+    zonefile: 'https://gaia.blockstack.org/hub/1BhQxPY3N9ABXzkbr4U8J6gkvAdr9Acr8L/profile.json',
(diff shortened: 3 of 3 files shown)


packages/profile/src/profile.ts
```diff
@@ -338,10 +338,17 @@ export function resolveZoneFileToProfile(zoneFile: any, publicKeyOrAddress: stri
     } catch (e) {
       reject(e);
     }
-
+    let isProfileUrl: boolean = false;
+    try {
+      // Check if zoneFile is only a profile url
```
friedger: This always happens for sub domains, correct? Should be in the comment because it is not clear from the naming.

friedger (commented):
Looks good, but the code is not easy to understand why the zone file can be the profile url.

packages/profile/src/profile.ts
ahsan-javaid: > Looks good, but the code is not easy to understand why the zone file can be the profile url.
Update: 
**Explanation**:  Below is the examples where you can see that zone file is only a profile url.
**Url**: https://stacks-node-api.mainnet.stacks.co/v1/names/jfontirroig.xck.app
**Response**:
```
{
  "address": "SP31C1RGYVJVWNYWJNRVSZQA2XN9HDEA08FB0G03S",
  "blockchain": "stacks",
  "last_txid": "0x0b80143fae8a52f47c970c5cf935ae37cd13236887e2a470d8a6891091219986",
  "resolver": "",
  "status": "registered_subdomain",
  "zonefile": "https://gaia.blockstack.org/hub/1BhQxPY3N9ABXzkbr4U8J6gkvAdr9Acr8L/profile.json",
  "zonefile_hash": "fdccd9ca23ec81f8d5ef0057eb2f72fb"
}
```
That's why i added this check to ensure that if its a profile url then do not run JSON.parse as it fails because its not a zoneFile object. 
So this check will skip the JSON.parse and retrieve the profile using the url mentioned in the zonefile attribute.  I also updated the comment to explain this point in code.
cc: @friedger

friedger (approved):
LGTM!

zone117x (commented):
I don't fully understand why subdomain zonefiles can sometimes be a URL rather than a zonefile. Is this because of changes in how the subdomain registrar works with Stacks 2.0?

asimm241:
> I don't fully understand why subdomain zonefiles can sometimes be a URL rather than a zonefile. Is this because of changes in how the subdomain registrar works with Stacks 2.0?

@zone117x No, `subdomain-registrar` does no such thing, as far as `subdomain-registrar` is concerned it takes the zone file from the user as an input so the user is responsible for an invalid zone file. https://github.com/blockstack/subdomain-registrar#sample-curl-scripts

ahsan-javaid:
Closing this pull request to avoid user specific checks. For detailed reasoning refer to https://github.com/blockstack/stacks.js/issues/1086

PR #1116, opened 2021-10-14
Support custom derivation paths with CLI

### LINKED ISSUE #918: Support custom derivation paths with CLI

ahsan-javaid:
## Description
This PR fixes the issue #918 by introducing custom derivation path support in `make_keychain` and `get_stacks_wallet_key` commands.
Refer to #918 for more details.

This PR closes the issue #918

## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Other

No
## Testing information
* `npm run test` in `cli`
* Use `make_keychain` and `get_stacks_wallet_key` commands in cli to apply custom derivation paths

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -331,8 +331,9 @@ async function getPaymentKey(network: CLINetworkAdapter, args: string[]): Promis
  */
 async function getStacksWalletKey(network: CLINetworkAdapter, args: string[]): Promise<string> {
   const mnemonic = await getBackupPhrase(args[0]);
+  const derivationPath: string | undefined = args[1] || undefined;
   // keep the return value consistent with getOwnerKeys
-  const keyObj = await getStacksWalletKeyInfo(network, mnemonic);
+  const keyObj = await getStacksWalletKeyInfo(network, mnemonic, derivationPath);
   const keyInfo: StacksKeyInfoType[] = [];
   keyInfo.push(keyObj);
   return JSONStringify(keyInfo);
@@ -355,7 +356,8 @@ async function makeKeychain(network: CLINetworkAdapter, args: string[]): Promise
     );
   }
 
-  const stacksKeyInfo = await getStacksWalletKeyInfo(network, mnemonic);
+  const derivationPath: string | undefined = args[1] || undefined;
+  const stacksKeyInfo = await getStacksWalletKeyInfo(network, mnemonic, derivationPath);
   return JSONStringify({
     mnemonic: mnemonic,
     keyInfo: stacksKeyInfo,
@@ -1965,5 +1967,7 @@ export const testables =
     ? {
         addressConvert,
         contractFunctionCall,
+        makeKeychain,
+        getStacksWalletKey,
       }
     : undefined;
--- a/packages/cli/tests/cli.test.ts
+++ b/packages/cli/tests/cli.test.ts
@@ -8,11 +8,12 @@ import {ClarityAbi} from '@stacks/transactions';
 import {readFileSync} from 'fs';
 import path from 'path';
 import fetchMock from 'jest-fetch-mock';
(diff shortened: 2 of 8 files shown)


packages/cli/src/cli.ts
```diff
@@ -331,8 +331,16 @@ async function getPaymentKey(network: CLINetworkAdapter, args: string[]): Promis
  */
 async function getStacksWalletKey(network: CLINetworkAdapter, args: string[]): Promise<string> {
   const mnemonic = await getBackupPhrase(args[0]);
+  const derivationPathPrompt = await prompt([
+    {
+      type: 'input',
+      name: 'derivationPath',
+      message: 'Enter custom derivation path or press enter to use the default: ',
```
friedger: The default path should be shown in the prompt

packages/cli/tests/cli.test.ts
```diff
@@ -179,3 +179,60 @@ describe('Contract function call', () => {
...
+    // Unmock TTY
+    process.stdin.isTTY = false;
+    process.env.password = undefined;
+  });
+
+  test('Get stacks wallet key custom derivation path', async () => {
+    const encrypted = 'vim+XrRNSm+SqSn0MyWNEi/e+UK5kX8WGCLE/sevT6srZG+quzpp911sWP0CcvsExCH1M4DgOfOldMitLdkq1b6rApDwtAcOWdAqiaBk37M=';
+    const args = [encrypted];
+    // @ts-ignore
+    inquirer.prompt = jest.fn().mockResolvedValue({derivationPath: "m/44'/5757'/0'/0/0"});
```
friedger: add a test for a custom path like `m/888'/0'/0`

friedger (commented):
Looks good!

packages/cli/src/cli.ts
ahsan-javaid: Done

packages/cli/tests/cli.test.ts
ahsan-javaid: Done

kyranjamie (commented):
Can the user pass this a an argument `--derivationPath m/44/xxx` or does it only open an interactive prompt?

ahsan-javaid:
> Can the user pass this a an argument `--derivationPath m/44/xxx` or does it only open an interactive prompt?


It only opens an interactive prompt. Also displays default derivationPath if user want to skip. 

Also well thought on your original suggestion to use some external arg parsing library to parse the argument like you mentioned. IMO, i thought not to overload stacks.js with one more external library. 

Second reason was that currently we are using index based args parsing and if we introduce arg parsing mechanism this way we will have two mechanisms in codebase (index based + arg parsing) as existing commands are using index based args. 

So these were the few thoughts for choosing prompt based approach for better experience.

kyranjamie (approved):
Accepting it as a flag argument would be good cli practice, but this is still an improvement. Thanks

ahsan-javaid:
> Accepting it as a flag argument would be good cli practice, but this is still an improvement. Thanks

Alright, i will implement `flag argument would be good cli practice` improvement in this PR. Thanks for the suggestion.

ahsan-javaid:
> Accepting it as a flag argument would be good cli practice, but this is still an improvement. Thanks

Added flag argument based cli approach. Thanks for suggestion.

diwakergupta:
@kyranjamie can you do another quick pass and then we can merge, thanks!

kyranjamie (approved):
Thanks @ahsan-javaid 👍🏼

PR #1117, opened 2021-10-15
Fix/1105 to handle usernames registered with wallet key

### LINKED ISSUE #1105: `lookupProfile` throwing error with `.btc` usernames
**Describe the bug**

Calling lookupProfile is throwing the following error when trying to access `.btc` usernames.

```
[ERROR] resolveZoneFileToProfile: error fetching token file https://gaia.blockstack.org/hub/1KvMJEjCMjwLxK4VHJn3wyYCcEuzB8qjZH/profile.json: Error: Token issuer public key does not match the verifying value
```

- @stacks/auth - v2.0.1

`@stacks/auth`

**To Reproduce**

```
import { lookupProfile } from '@stacks/auth';

await lookupProfile({ username: 'leopradel.btc' });
```

**Expected behavior**

The profile should be returned, it's working fine for `*.id.blockstack` and `*.id.stx` usernames.

**Additional context**

I already faced the same error for `*.id.stx` usernames before in https://github.com/blockstack/stacks.js/issues/1016 and submitted a pull request to fix it.
For this one, I tried to debug but didn't manage to make it work.

friedger:
## Description

The BNS protocol defines a way to discover profile data by other users' names. The current implementation is broken for BNS names registered through the stacks wallet extension as described in #1105, as well as in https://github.com/blockstack/stacks-wallet-web/issues/1412 . Thank you to @pradel for filing the issue and working on a solution with me!

In the BNS protocol, the profile data are stored as a signed JWT at a location given in the zonefile of the username. When calling `lookupProfile` with a username the validity of the token is verified by comparing the signer key to the owner key of the BNS name.

There are two key that belongs to an account: 
* the key for stx address (for short "wallet key")
* the key for gaia access (for short "data key")

"Blockstack" names (Stacks 1.0 mainly `.id` and `.id.blockstack`) are owned by the data key (BTW, also .btc or other names can be owned by the data key in theory)

"Stacks" names (Stacks 2.0 registered via Stacks Extension) are owned by the wallet key because the txs are signed by the wallet.

Until now, the stacks.js library uses the data key to sign the profile token.

### Change

This PR changes in the packages `wallet-sdk`
* the signing key used for signing the profile token to the compressed wallet key (`stxPrivateKey`) 
* the way how accounts are restored: the stxPrivateKey is choosen based on the username in the config or based on newly registered usernames.
* that `restoreWalletAccounts` now requires a network

A similar change would be required in the `keychain` package and should be submitted in a separate PR.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

**Before** this change:
* profile tokens of Blockstack names are signed with the data key that owns the name -> `lookupProfile` works.
* profile tokens of Stacks names are signed with the data key that does not own the name -> `lookupProfile` fails.

**After** this change:
* profile tokens of Blockstack names are signed with the key that owns the name -> `lookupProfile` will work.
* profile tokens of Stacks names are signed with the wallet key that owns the name -> `lookupProfile` will work.

The Hiro Stacks Wallet needs to provide a StacksNetwork when in `storeSeed`.

### Required Actions by Users
Users with Stacks names do not take any actions

Users with Blockstack names do not take any actions.

Users with Stacks names and Blockstacks names on the same account (e.g. they bought a .btc name with their old secret key) can only use the Stacks name for sharing data. e.g. on sigle the address is app.sigle.io/new-name.btc instead of app.sigle.io/old-name.id. No data needs to be transferred.

Users with Blockstack names can optionally transfer their name for the address of the data key to the address of the stacks key. This requires 1 tx by the user. Fortunately, the data on gaia is still accessible in the same way, because access to the gaia storage is not impacted by the ownership of the name as long as the new address is derivate from the same root node.

In general the documentation of the BNS protocol needs to be created and ratified as a SIP. This is out of the scope of this issue.

## Testing information
As provided in #1105 

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated


zone117x:
@friedger thank you for figuring out this bug! Can you rebase onto latest `master` branch? Just merged a fix so unit tests run on PRs.

zone117x:
@markmhx @kyranjamie can you take a look at the breaking changes mentioned in this PR, and see if it makes sense for us to merge this as is?

packages/wallet-sdk/src/models/profile.ts
```diff
@@ -88,7 +88,7 @@ export const fetchAccountProfileUrl = async ({
 };
 
 export function signProfileForUpload({ profile, account }: { profile: Profile; account: Account }) {
-  const privateKey = account.dataPrivateKey;
+  const privateKey = account.stxPrivateKey.slice(0, 64);
```
kyranjamie: What is the relevance of 64 here?
```suggestion
  const descriptiveNameOfWhatNumberIs = 64
  const privateKey = account.stxPrivateKey.slice(0, descriptiveNameOfWhatNumberIs);
```

kyranjamie (commented):
Makes sense and code is fine.

@markmhx @agraebe there are product questions to answer here. 

- Can we just go ahead with this?
- Are there many Blockstack name owners this will effect?
- Do we need a migration path in place prior to releasing these changes?

pradel:
@kyranjamie you probably want to read this https://github.com/blockstack/stacks-wallet-web/issues/1412#issuecomment-944882345 for the migration.

packages/wallet-sdk/src/models/profile.ts
friedger: @kyranjamie It is the part required for compressed public key. 
There are 8 places in the repo that uses this expression `slice(0, 64)`. Should I replace all of them? Might be better as a separate PR.

friedger:
Currently, all .btc name owners are affected, and all stacks web apps sharing data like sigle

agraebe:
Great to see this PR and thanks for flagging the product questions @kyranjamie!
 
>     * Can we just go ahead with this?

I'd like to discuss this with Splat (cc @diwakergupta). It seems like a change we should make for compatibility reasons for Stacks 2 BNS naming. Merging the PR will not cut a new release automatically, so we would still have a bit of time orchestrating the release efforts. However, I wonder if there's any chance we could implement backward compatibility here? Is there a chance to check for Stacks 1 names and cover that case?

>     * Are there many Blockstack name owners this will effect?

I think we should review this before pulling the trigger. I'll bring this up in the next Splat meeting to see how we can get the numbers.

>     * Do we need a migration path in place prior to releasing these changes?

Yes, I'd advocate for a release process that includes the following:
- a forum post outlining the breaking change with the timeline of the release (including a set date at which the new change will be available to developers)
- an upgrade guide and messaging for developers that would explain the change (and actions required) to users with old Blockstack names
- an outreach to teams we know use BNS to explain the details

friedger:
> I think we should review this before pulling the trigger. I'll bring this up in the next Splat meeting to see how we can get the numbers.

Currently, 17k bns names have been registered since launch of 2.1

On the stacks node api db you can run:
```
select asset_identifier, asset_event_type_id, count(*), min(block_height)   from nft_events ne
where canonical = true and microblock_canonical = true
and asset_identifier = 'SP000000000000000000002Q6VF78.bns::names'
group by asset_identifier, asset_event_type_id order by asset_event_type_id , count(*) desc, min(block_height)  
```

However, this only returns the number of names that are **likely** to have profiles signed with the other key. Old and new names do not indicate whether their profile was signed with one or the other key.

friedger:
> I wonder if there's any chance we could implement backward compatibility here?

What does backwards compatibility mean here? Verify the signature against one or the other key? The library could be release as two versions: Stacks 1-compatible and Stacks 2-compatible. Wallet developers would need to make a choice or support both.

diwakergupta:
@friedger thanks for the PR, and apologies for the slow progress.

In terms of next steps, I'd like to hear from @markmhx . I'm copying relevant bits from the PR description:

> **Before this change:**
> 
> profile tokens of Blockstack names are signed with the data key that owns the name -> lookupProfile works.
> profile tokens of Stacks names are signed with the data key that does not own the name -> lookupProfile fails.
> 
> **After this change:**
> 
> profile tokens of Blockstack names are signed with the wallet key that does not own the name -> lookupProfile will fail.
> profile tokens of Stacks names are signed with the wallet key that owns the name -> lookupProfile will work.

and

> **Required Actions by Users**
> Users with Stacks names do not take any actions
> 
> Users with Blockstack names need to transfer their name for the address of the data key to the address of the stacks key.

My sense is that most users will legacy names will just not do this (unless forced and/or the migration path is as easy as clicking a button).

@markmhx can you add some color on the following to help figure out next steps:

* How many apps are currently using / relying on Stacks 1.0 names? Basically trying to assess the impact of merging this and depending on that, figure out what sort of release process / comms / extra work might be required.
* Are there plans to provide a migration path in the wallet (c.f. blockstack/stacks-wallet-web#1412)? If so, would such a path depend on this PR?

Also cc/ @andresgalante

markmhendrickson:
I have the same question as @agraebe here as to whether we could set up `lookupProfile` to be backwards compatible.

Could the function essentially check for a profile associated with the _wallet_ key first and if not found, check for one associated with the _data_ key, returning null only if not found on either key?

There should remain only a few active apps (e.g. Sigle, BlockSurvey, Envelop) that rely on this type of query for Stacks 1.0 / Blockstack names as far as I can tell. I presume they'll want a seamless experience for their users without the need for user action, and that they'll want to support both 1.0 and 2.0 names without distinction (e.g. to load user profiles using either type all the same).

If we can't wrap such backwards compatibility into `lookupProfile` itself, perhaps we could add `lookupProfileLegacy` or similar for those few apps who need to handle this on the app side?

IIRC https://github.com/blockstack/stacks-wallet-web/issues/1412 is actually due to a superficially similar but different underlying issue, in that we changed how the web wallet / extension applies derivation paths for addresses vs. the paths used by Blockstack Connect in the Stacks 1.0 era. So, I don't suspect that issue depends on this PR per se though @friedger or @pradel (or @hstove or @aulneau) should correct me if I'm wrong here and they're actually related.

friedger:
`lookupProfile` is a client side function, therefore, the client can only use the key received from the authenticator and the data received from the blockchain. 

For backwards compatibility, the authenticator could choose to return two different identities (using the different derivation paths) and then the client can choose to verify against one or the other.

@markmhx It is not about `lookupProfile` vs `lookupProfileLegacy` (in @stacks/auth used by the client) but about `signProfile` vs `signProfileLegacy` (in @stacks/keychain used by the authenticator). Therefore, it is related to https://github.com/blockstack/stacks-wallet-web/issues/1412

friedger:
> How many apps are currently using / relying on Stacks 1.0 names?

@diwakergupta It is not about Stacks 1.0 names vs Stacks 2.0 names. What does it mean? The data for all names is stored on Stacks 2.0 blockchain. It is about the authenticator. How many apps use the authenticator on https://app.blockstack.org/? Isn't that sunset already?

Merging this PR only affects the keychain package used by wallet developers.

packages/wallet-sdk/src/models/profile.ts
aulneau: if i'm not mistaken, it's the difference between a compressed and uncompressed key. compressed keys have extra bytes at the end that show it's compressed, typically it's `01`. this removes that extra byte. uncompressed keys don't have it, so they are already at length 64. as @friedger said, it's used all over in the repo, but could be made into a helper function that shows why this is done. 

@zone117x could probably correct me if i'm wrong

friedger:
I have updated the description to match the implementation of support for signing both cases: usernames owned by data keys and usernames owned by wallet keys.

packages/wallet-sdk/src/models/profile.ts
```diff
@@ -111,6 +117,21 @@ export async function uploadProfile({
   );
 }
 
+export const selectPrivateKey = async (account: Account) => {
+  let usernameOwner: string | undefined;
+  if (account.username) {
+    const nameInfo = await new StacksMainnet().getNameInfo(account.username);
```
kyranjamie: Can we assume mainnet here?

friedger: Has been removed

friedger:
The current solution leave the lookup of the name to the creator of the Account object

pradel:
I tested the latest changes with a dev version of the wallet and everything worked 👌  
Don't see any blocker for the merge of this one, I found this issue which will prevent Sigle to migrate to Hiro wallet after this is merged, but should be fixed in a separate pr https://github.com/blockstack/stacks.js/issues/1144

packages/wallet-sdk/src/usernames.ts
```diff
@@ -0,0 +1,18 @@
+import { fetchPrivate } from '@stacks/common';
+import { StacksNetwork } from '@stacks/network';
+
+export const fetchFirstName = async (
```
friedger, line 4: Should this be moved to bns package?

packages/wallet-sdk/src/derive.ts
```diff
@@ -71,19 +73,182 @@ export const deriveSalt = async (rootNode: BIP32Interface) => {
...
+        return DerivationType.Unknown;
+      } else {
+        return DerivationType.Data;
+      }
+    } else {
+      return DerivationType.Wallet;
+    }
+  } else {
+    // no network to determine the derivation path
+    return DerivationType.Unknown;
```
kyranjamie, line 165: Might it be better to just throw an error here?

kyranjamie (approved):
Great work @friedger

packages/wallet-sdk/src/usernames.ts
markmhendrickson: That works for me if @kyranjamie and @aulneau are in favor here.

packages/wallet-sdk/src/derive.ts
kantai: I agree an error is probably better than having a `DerivationType.Unknown`, especially because a lot of the uses of `DerivationType` just switch on whether it is equal to `Wallet`.

kantai (approved):
This looks good to me -- I think the only change should be throwing an error rather than having an Unknown enum variant.

packages/wallet-sdk/src/derive.ts
friedger: I disagree, because we are processing a correctly created and registered name. The function should leave it to the user/developer of the library to handle this case. Maybe they have a fallback handling trying more derivation paths.

friedger:
@kantai The error is thrown in the restore function: https://github.com/blockstack/stacks.js/pull/1117/files#diff-16f84c8c7e6458f1af2ba7d3c77e2adcb21835834c511336e5ac5733de784d38R64

Throwing an error during the select function would close down use cases for the library.

kantai:
> Throwing an error during the select function would close down use cases for the library.

Okay, that's fine with me.

pradel:
Hey, any update?

andresgalante:
@reedrosenbluth if this looks ok for you, let's merge it, please

friedger:
Thank you!

Issue #1120, opened 2021-10-19
Bundle size of the stacks package is really huge


pradel:
**Is your feature request related to a problem? Please describe.**

I have a super simple next.js app, and the bundle size of the js files is in total `295 kB` gzipped. Without stacks.js dependencies, the size is around `65 kB`. So most of my bundle size is coming from the stacks dependencies as you can see in the following screenshot.

![Capture d’écran 2021-10-19 à 13 24 25](https://user-images.githubusercontent.com/5749437/137900584-912a2a55-5c64-47da-a052-f0140a0ff0e6.png)

**Describe the solution you'd like**

Using stacks.js libs shouldn't make my app slow to load for my users.

Opening the issue to start the discussion around improvements that can be made in stacks.js

rafaelcr:
We are aware of this issue, we'll get back to you ASAP

ahsan-javaid:
@zone117x As a starting point on this issue, to get rid of some of dependencies like bn.js and few others, is it allowed to use some source code from `micro-stacks` repo. I can see `micro-stacks` has `GNU GENERAL PUBLIC LICENSE` and stacks js is using `The MIT License (MIT)`. I am asking in terms of copyright issue / permissions to use `micro-stacks` code, will it be allowed or not.

ahsan-javaid:
@zone117x Any thoughts on above question 🙏

pradel:
Can you reopen the issue? I don't think this one is solved yet.

pradel:
@reedrosenbluth

janniks:
_Moving this to the relevant milestone and sub-issues_
https://github.com/hirosystems/stacks.js/milestone/56

PR #1121, opened 2021-10-19
feat: upgrade bn.js from 4.12 to 5.2

pradel:
## Description

At the moment @stacks/connect is using bn.js version 5.2 and the packages of this repo the version 4.12. This means that when we use both connect and stacks.js packages in our app, we will end up having 2 versions of bn.js in our bundle.

What I did was to upgrade bn.js dependency from 4.12 to the next major 5.2 so both connect and stacks.js use the same version.

stacks-wallet-web is already using 5.2 too https://github.com/blockstack/stacks-wallet-web/blob/main/package.json#L71

Related to issue #1120, I noticed that bn.js is included twice in my bundle.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

All the tests are passing so not sure if this qualifies as a breaking change as users won't be impacted directly by this change.

I don't think so.

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated


packages/encryption/package.json
```diff
@@ -31,11 +31,11 @@
   },
   "dependencies": {
     "@stacks/common": "^2.0.1",
-    "@types/bn.js": "^4.11.6",
+    "@types/bn.js": "^5.1.0",
     "@types/node": "^14.14.43",
     "bip39": "^3.0.2",
     "bitcoinjs-lib": "^5.2.0",
```
zone117x, line 37: IIRC it was the `bitcoinjs` family of libs (this one and some of the `bip` libs) that were including `bn.js` v4. Have you tested that these dependencies don't end up requiring `bn.js` v4 to be bundled anyway?

pradel: Other packages are depending of v4, my idea was to do the upgrade gradually, here is the list of deps still using it (might have missed some):

- @stacks/keychain
  - @blockstack/rpc-client 0.3.0-alpha.11
    - @blockstack/stacks-transactions 0.5.1
- Used by multiple packages
  - jsontokens 3.0.0
    - asn1.js 5.4.1
- @stacks/cli 2.0.1
  - blockstack 19.3.0
- Used by multiple packages
  - elliptic 6.4.5
- bitcoin-lib 5.2.0
  - tiny-secp256k1 1.1.6

The issue can be that some of them seem unmaintained, like https://github.com/indutny/asn1.js/pull/118 where the pr was opened almost 2 years ago.

zone117x: I'd be more concerned with getting the `elliptic` package and bitcoinjs-lib packages upgraded to bn.js v5. Is there anything in bn.js v5 that is needed? And would be more feasible to change projects using v5 to use v4 instead?

pradel: Looks like bitcoin-js lib is working on removing the dependency to bn.js as they are dropping tiny-secp256k1 https://github.com/bitcoinjs/bitcoinjs-lib/pull/1734. For elliptic it looks more challenging.
I have no idea if you are using v5 specific features on the other packages.

junderw: We are only dropping bn.js from the main package, ecpair and bip32 will still use it until we can upgrade them to use WASM.

diwakergupta:
Hi @pradel -- based on the comments, it doesn't look like this PR would actually help reduce the bundle size. I'm going to close it for now, but please reopen if you disagree. Meanwhile, we _are_ looking into all the different pain points with stacks.js, including bundle sizes, and hope to have a plan we can share with the community in the coming weeks. Thanks!

PR #1124, opened 2021-10-21
Unsigned equivalent of makeContractDeploy

### LINKED ISSUE #1104: unsigned equivalent of makeContractDeploy
there are unsigned version for contract call as well as STX transfer in the transactons package for the use case that I may construct the transaction on machine A and send the serialized version(non-signed) to machine B(or say some mobile wallet in the future via some WalletConnect like mechanism) for signing.

But there is no such thing for contract deployment, only makeContractDeploy() which expect the senderKey as private key. While it is less likely to have this form of usage(frequent deployment of contract), but for consistent sake, would like to see similar one so I can designed my application to use this form of workflow.

### LINKED ISSUE #1140: Export `makeUnsignedContractDeploy` method
We have this already for STX token transfers and contract calls, but not contract deploys.

We'll need this method to support contract deploys on hardware wallets.

ahsan-javaid:
## Description
This PR introduces `Unsigned equivalent of makeContractDeploy` in the `transaction` package to support the workflow request in #1104
This PR closes the issue #1104

## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Other

No
No
## Testing information
1. `npm run test` passed
2. Include the test case to test the `makeUnsignedContractDeploy`

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/transactions/src/builders.ts
+++ b/packages/transactions/src/builders.ts
@@ -23,6 +23,7 @@ import {
   publicKeyToAddress,
   pubKeyfromPrivKey,
   publicKeyFromBuffer,
+  createStacksPublicKey,
 } from './keys';
 
 import { TransactionSigner } from './signer';
@@ -603,12 +604,10 @@ export async function makeSTXTokenTransfer(
 /**
  * Contract deploy transaction options
  */
-export interface ContractDeployOptions {
+export interface BaseContractDeployOptions {
   contractName: string;
   /** the Clarity code to be deployed */
   codeBody: string;
-  /** a hex string of the private key of the transaction sender */
-  senderKey: string;
   /** transaction fee in microstacks */
   fee?: IntegerType;
   /** the transaction nonce, which must be increased monotonically with each new transaction */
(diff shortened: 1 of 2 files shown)


reedrosenbluth (changes requested):
Thanks for this!

There's a lot of duplicate code in this new `makeUnsignedContractDeploy` function. Could you possibly refactor the `makeContractDeploy` function such that it utilized `makeUnsignedContractDeploy` to reduce the amount of duplicate code?

packages/transactions/src/builders.ts
```diff
@@ -690,6 +699,23 @@ export async function estimateContractDeploy(
  */
 export async function makeContractDeploy(
   txOptions: ContractDeployOptions
+): Promise<StacksTransaction> {
+  const privKey = createStacksPrivateKey(txOptions.senderKey);
+  const stacksPublicKey = getPublicKey(privKey);
+  const publicKey = publicKeyToString(stacksPublicKey);
+  const unsignedTxOptions = { ...txOptions, publicKey } as UnsignedContractDeployOptions;
```
zone117x: It's generally safer to define the object type rather than cast to it, i.e.
```ts
const unsignedTxOptions: UnsignedContractDeployOptions = { ...txOptions, publicKey };
```
Casting allows things like this to pass the typechecker:
```ts
const unsignedTxOptions = { ...txOptions, publicKey, wrong: 124 } as UnsignedContractDeployOptions;
```

zone117x (approved):
Left a comment around a minor type safety improvement, otherwise LGTM!

packages/transactions/src/builders.ts
ahsan-javaid: Done

kyranjamie (approved):
Just what I was after, thanks!

kyranjamie:
@reedrosenbluth can we merge?

PR #1129, opened 2021-11-02
[Browser support] Fix type error can't convert BigInt to number

### LINKED ISSUE #1096: Support environments without native `bigint`
I am trying to update a project to the latest libraries, and have the following error:

Screenshot 2021-08-26 at 16 30 51

`
 "dependencies": {
     ...
    "@stacks/auth": "^2.0.1",
    "@stacks/connect": "^5.0.5",
    "@stacks/profile": "^2.0.1",
    "@stacks/storage": "^2.0.1",
    "@stacks/transactions": "^2.0.1",
    ...
}
`

We were using node: 12.3.0
Updating only this libraries resulted in a error telling us we needed to use a higher version >= 12.13.0

But even inside the node JS 14.17.5
` Math.pow(BigInt(2), BigInt(128)) - BigInt(1); `
 will result in the error:
 
 **Cannot convert a BigInt value to a number**
 
 Can someone assist?

### LINKED ISSUE #1119: App with react-script throws TypeError: can't convert BigInt to number
**Describe the bug**
When running my app "pool registry" (https://github.com/friedger/stacks-pool-registry)  with the 2.0.1 in production, I get the error: 
`TypeError: can't convert BigInt to number`. The app is build using `react-script`

**What version of stack.js are you using?**
2.0.1

**To Reproduce**
1. clone the repo
2. yarn
3. yarn build
4. serve -s build

**Expected behavior**
localhost:5000/hodl should show

**Screenshots**
Screenshot from 2021-10-18 18-11-42

Maybe related to #1096 

**Additional Context**
https://github.com/facebook/create-react-app/issues/6907

ahsan-javaid:
## Description
The `transpilers` coverts the lines like `const MAX_U128 = BigInt(2) ** BigInt(128) - BigInt(1);` to `Math.pow(BigInt(2), BigInt(128)) - BigInt(1);`.

Definitely `Math.pow` will throw exception `can't convert BigInt to number`.

A lot of users are facing this issue and merging this pr will unblock them.

This PR closes the issue #1119 and #1096

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No
## Testing information
* `npm run test`
* Merging this pr will fix `can't convert BigInt to number` exception

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/transactions/src/clarity/types/intCV.ts
+++ b/packages/transactions/src/clarity/types/intCV.ts
@@ -1,10 +1,11 @@
 import { IntegerType, intToBigInt } from '@stacks/common';
 import { ClarityType } from '../clarityValue';
 
-const MAX_U128 = BigInt(2) ** BigInt(128) - BigInt(1);
+const MAX_U128 = BigInt('0xffffffffffffffffffffffffffffffff'); // (2 ** 128 - 1)
 const MIN_U128 = BigInt(0);
-const MAX_I128 = BigInt(2) ** BigInt(127) - BigInt(1);
-const MIN_I128 = BigInt(-2) ** BigInt(127);
+const MAX_I128 = BigInt('0x7fffffffffffffffffffffffffffffff'); // (2 ** 127 - 1)
+// no signed (negative) hex support in bigint constructor
+const MIN_I128 = BigInt('-170141183460469231731687303715884105728'); // (-2 ** 127)
 
 interface IntCV {
   readonly type: ClarityType.Int;


packages/transactions/src/clarity/types/intCV.ts
```diff
@@ -1,10 +1,10 @@
-import { IntegerType, intToBigInt } from '@stacks/common';
+import { IntegerType, intToBigInt, bigIntPow } from '@stacks/common';
 import { ClarityType } from '../clarityValue';
 
-const MAX_U128 = BigInt(2) ** BigInt(128) - BigInt(1);
+const MAX_U128 = bigIntPow(BigInt(2), BigInt(128)) - BigInt(1);
```
zone117x: Clever solution! But, because these are constants (module inlines), it would better to define them with as little overhead as possible. These were only math expressions in the first place for clarity to the reader, which could just be comments. So something like:
```ts
const MAX_U128 = BigInt('0xffffffffffffffffffffffffffffffff'); // (2 ** 128 - 1)
const MIN_U128 = BigInt(0);
const MAX_I128 = BigInt('0x7fffffffffffffffffffffffffffffff'); // (2 ** 127 - 1)
// no signed (negative) hex support in bigint constructor
const MIN_I128 = BigInt('-170141183460469231731687303715884105728'); // (-2 ** 127)
```

ahsan-javaid: Thanks

zone117x (approved):
Note that this will fix the issue with the bugged bundlers injecting a `Math.pow` statement but it will not close https://github.com/blockstack/stacks.js/issues/1096
Environments like react-native (and older browsers) don't support `bigint` at all.

kyranjamie (approved):

PR #1134, opened 2021-11-05
Use new fee estimation endpoint

### LINKED ISSUE #1133: Support new fee estimation endpoint
Stacks.js should be using the new `/v2/fees/transaction` endpoint:

https://github.com/blockstack/stacks-blockchain/pull/2872

reedrosenbluth:
fixes: #1133


packages/transactions/src/builders.ts
```diff
@@ -137,6 +140,78 @@ export async function estimateTransfer(
...
+ * and signatures
+ * @param {StacksNetwork} network - the Stacks network to estimate transaction fees for
+ *
+ * @return a promise that resolves to FeeEstimate
+ */
+export async function estimateTransaction(
+  transaction: StacksTransaction,
+  estimatedLen?: number,
+  network?: StacksNetwork
+): Promise<FeeEstimate> {
```
kantai: Should the caller pass in a `StacksTransaction` or a `StacksTransactionPayload` here? Would they need to have already applied the tx fee, etc., to get the `StacksTransaction` object, or can they just supply a partial? (And how would they get the partial transaction -- aren't most of the builder methods one-shot?)

packages/transactions/src/builders.ts
```diff
@@ -137,6 +140,78 @@ export async function estimateTransfer(
...
+    middle: bigint
+  },
+  estimated_fees: FeeEstimate
+}
+
+interface FeeEstimate {
+  high: bigint,
+  middle: bigint,
+  low: bigint,
+}
```
kyranjamie: Isn't the endpoint an array, not an object with high/middle/low? (see comments in blockchain repo pr)

reedrosenbluth: not according to this PR description:

https://github.com/blockstack/stacks-blockchain/pull/2872

packages/transactions/src/builders.ts
```diff
@@ -150,18 +150,12 @@ interface FeeEstimateResponse {
...
-interface FeeEstimate {
-  high: bigint,
-  middle: bigint,
-  low: bigint,
+  estimated_fee_rates: [
+    bigint, bigint, bigint
+  ],
+  estimated_fees: [
+    bigint, bigint, bigint
+  ]
```
kyranjamie: https://github.com/blockstack/stacks-blockchain/pull/2872/files#diff-f77826b61e33c610fb326b65534d0d94cae05967bdad98eca951264ff8422995

it's one array of objects

packages/transactions/src/builders.ts
reedrosenbluth: I think a `Payload` input would be better, will make this change!

packages/transactions/src/builders.ts
```diff
@@ -1280,13 +1356,10 @@ export async function sponsorTransaction(
...
       case PayloadType.TokenTransfer:
-        txFee = await estimateTransfer(options.transaction, network);
-        break;
       case PayloadType.SmartContract:
-        txFee = await estimateContractDeploy(options.transaction, network);
-        break;
       case PayloadType.ContractCall:
-        txFee = await estimateContractFunctionCall(options.transaction, network);
+        const estimatedLen = options.transaction.serialize().byteLength;
+        txFee = (await estimateTransaction(options.transaction.payload, estimatedLen, network))[1].fee;
```
kyranjamie: Is this safe? Should we check for undefined? 🤔

kyranjamie (approved):

packages/transactions/src/builders.ts
```diff
@@ -137,6 +141,71 @@ export async function estimateTransfer(
   return feeRate * txBytes;
 }
 
+interface FeeEstimation {
+  fee: bigint,
```
zone117x: Are these `bigint` or `number`? The unit tests have `number`s

reedrosenbluth: changing them to `number`

PR #1146, opened 2021-11-30
fix: jsdom failing with init bigint

kyranjamie:
Tests failing with jsdom


zone117x (approved):
LGTM assuming tests pass

PR #1150, opened 2021-12-16
Offload `bn.js`

### LINKED ISSUE #1120: Bundle size of the stacks package is really huge

ahsan-javaid:
## Description
This PR removes `bn.js` dependency from stacks.js packages.  
`bn.js` minified size is around `43kb`. This will help to reduce the bundle size. In the subsequent pr, further dependencies can be removed to reduce the bundle size further. 

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No

No

## Testing information

1. `npm run test`

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated


pradel:
@ahsan-javaid amazing to see some work to improve the bundle size, much needed 🙏

saralab:
@kyranjamie @zone117x

ahsan-javaid:
@zone117x Share your thoughts on this pr as well. Thanks 🙏

packages/common/src/utils.ts
```diff
@@ -397,3 +395,116 @@ export function intToBigInt(value: IntegerType, signed: boolean): bigint {
...
+  byteToHexCache[n] = n.toString(16).padStart(2, '0');
+}
+
+export function bytesToHex(uint8a: Uint8Array) {
+  const hexOctets = new Array(uint8a.length);
+  for (let i = 0; i < uint8a.length; ++i) hexOctets[i] = byteToHexCache[uint8a[i]];
+  return hexOctets.join('');
+}
+
+export class BN {
```
zone117x: Is a goal of this PR to avoid breaking changes? If so, this new class is a breaking change. It prevents other libraries from passing a `bn.js` lib instance of `BN` because the class types are different, and the functionality is different.

If we don't care about breaking changes, then I'm not sure why we'd try to replicate the bn.js structure when everything can use a `bigint` instance, and call any of these helper functions when needed.

ahsan-javaid: Done!

ahsan-javaid:
@zone117x Requesting the review on this pr 🙏.

packages/common/src/utils.ts
```diff
@@ -331,16 +329,11 @@ export function getGlobalObjects<K extends Extract<keyof Window, string>>(
   return result;
 }
 
-export type IntegerType = number | string | bigint | Uint8Array | BN;
+export type IntegerType = number | string | bigint | Uint8Array;
```
zone117x: Seems like it would be pretty easy to maintain bn.js support by creating an `interface` that matched bn.js class instances, like this handful of properties:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ea0e0de908b17dd697e373da9cd0fca8d7e41c66/types/bn.js/index.d.ts#L112-L145

Then copy over a bit of code from `BN.isBN` to use in `intToBigInt(...)`, like this line: https://github.com/indutny/bn.js/blob/db57519421f0c47c9f68c05fa6fc12273dcca2c2/lib/bn.js#L66-L67

That would avoid creating a lot of breaking changes for all the projects still passing in a bn.js instance to various functions.

ahsan-javaid: Done!

ahsan-javaid:
@zone117x Any further thoughts on this pr 🙏

zone117x (approved):
LGTM

packages/common/src/utils.ts
```diff
@@ -330,17 +328,13 @@ export function getGlobalObjects<K extends Extract<keyof Window, string>>(
   }
   return result;
 }
-
+// After removing bn.js library provide backward compatibility for users passing bn.js instance
+type BN = import('bn.js'); // Type only import from @types/bn.js
```
janniks, line 332: Should this be `@types/bn.js` ?
Not sure, but is this not just dynamic importing the whole `BN`?

janniks (commented):
Looks good so far — but I'll need some more time to examine in detail 😊

packages/common/src/utils.ts
ahsan-javaid: **Test Case**: Verify dist/build should not contain bn.js dependency after removing bn.js 
- I verified the dist folder does not contains bn.js dependency / imports
- Compared it with master branch generated build which contains bn.js dependency

**Answer**
- Importing  like @types/bn.js is incorrect and gives error and suggest to import like import('bn.js')
- ** I repeat type BN = import('bn.js') is // Type only import from @types/bn.js ** 
- When  generate a ts build these lines get removed in generated build folder
- This can be verified by running test case as i mentioned above

janniks: Interesting! Good to know — thanks 🙏

janniks (approved):

yknl:
What's the latest on this PR? The switch to BigInt from bn.js prevents stacks.js from being used in environments like React Native due to the lack of BigInt support in the JavascriptCore version used by RN. 

As a result we're currently unable to update to the latest version of this library.

I propose switching back to bn.js for all packages in stacks.js if the only concern is bundle size.

janniks:
@yknl good point! Do you know if solutions like https://github.com/facebook/react-native/issues/28492#issuecomment-824698934 would work for you in this case?

yknl:
> @yknl good point! Do you know if solutions like [facebook/react-native#28492 (comment)](https://github.com/facebook/react-native/issues/28492#issuecomment-824698934) would work for you in this case?

Unfortunately the suggested polyfill package does not work for stacks.js. There are differences between the polyfill and the BigInt type. For example, the polyfill cannot instantiate from a `0x` prefixed hex string which something inside stacks.js uses. This causes a crash.

janniks:
@yknl Got it.

Have you tried this https://github.com/react-native-community/jsc-android-buildscripts by any chance? From what I'm reading, jsc is the problem (which is up-to-date on iOS, as it uses the built-in jsc), but android uses a crazy old bundled jsc. This buildscript allows shipping with an up-to-date jsc on android.

zone117x:
Note that this specific PR is a relatively minor move towards native `bigint` compared to many changes that have already been pushed. For example, the new and much improved cryptography dependencies are heavily dependent on native `bigint`, see https://github.com/paulmillr/noble-secp256k1/blob/main/index.ts (totally dependent on bigint: exponentiation operator `**`, bigint literals `123n`, etc). 

We should not hold back from merging this PR due to react native support, as that's already been thrown out the window.

Facebook's react native js runtimes are embarrassingly outdated, but looks like they are making progress towards `bigint` support at least https://github.com/facebook/hermes/issues/510.

Alternatively, @aulneau and I have had discussions around this topic, and found offloading certain operations to native webviews as a potential viable solution, e.g. https://github.com/webview-crypto/react-native-webview-crypto.

@aulneau do you have any updates around this issue?

aulneau:
If packages are moving towards more native bigint usage -- they simply cannot be used in react-native as it stands currently. I've explored ways to create a messaging bridge between the native webview and react native which then makes it possible to use something like micro-stacks or noble-secp256k1 in react native. This however has a large overhead and burden of complexity. 

If stacks.js is already using the noble packages, this change in this PR will not make it so react native users can't use stacks.js, it would have already been the case because of the heavy usage of bigint in the noble packages.

> I propose switching back to bn.js for all packages in stacks.js if the only concern is bundle size.

An alternative would be to use a different big int package that has a smaller footprint than bn.js. 

I've thought a lot about how this would be possible in the context of micro-stacks. What I've landed on currently is to take an approach where consumers of the lib could pass in their own versions of things like a secp256k1 package or other encoding libs, which have the same signature. That way you can pass a more performant secp256k1 package that is built for react native, rather than using what is best for the web. Additionally, I think using native `Bigint` is fine as long as there is no math being done. That is easy enough to polyfill in react native. It's much more tricky when it uses `**` and the like.

Hope that helps. I'm planning on having an example implementation for react native and micro-stacks in the coming months.

zone117x:
Thanks for the insight and research @aulneau!

We should consider a similar approach in stacks.js _if_ we prioritize react-native support. Keep in mind one of their maintainers mentioned upcoming support for bigints just a week or so ago.

packages/common/src/utils.ts
```diff
@@ -380,17 +374,26 @@ export function intToBigInt(value: IntegerType, signed: boolean): bigint {
...
+  if (
+    parsedValue != null &&
+    typeof parsedValue === 'object' &&
+    parsedValue.constructor.name === 'BN'
+  ) {
     return BigInt(parsedValue.toString());
   }
   throw new TypeError(
-    `Invalid value type. Must be a number, bigint, integer-string, hex-string, BN.js instance, or Buffer.`
+    `Invalid value type. Must be a number, bigint, integer-string, hex-string, BigInt instance, or Buffer.`
```
zone117x: Nit: message already mentions `bigint`, no need to specify the constructor name `BigInt`.

janniks: 🙏

packages/stacking/src/utils.ts
```diff
@@ -140,7 +139,7 @@ export function poxAddressToBtcAddress(...args: PoxAddressArgs): string {
 }
 
 export function getBTCAddress(version: Buffer, checksum: Buffer) {
-  const btcAddress = address.toBase58Check(checksum, new BN(version).toNumber());
+  const btcAddress = address.toBase58Check(checksum, Number(version.toString()));
```
zone117x: This is a bit confusing. The `version` variable should always be a single byte right? In that case, wouldn't something like `address.toBase58Check(checksum, version[0])` make more sense?

janniks: Great catch!
Added a [fix](https://github.com/hirosystems/stacks.js/pull/1150/commits/a1b65de96e104cdb328039e68907a8f972449b94) that also accepts plain number as `version`

PR #1151, opened 2021-12-19
wallet-sdk / fix stx address derivation instead of stxPrivateKey in name fetching

ma-aboelenein:
## Description

This Pull Request fixes the latest release of stacks JS (3.1.0) where names resolution was adjusted in this [PR](https://github.com/blockstack/stacks.js/pull/1117)

1. How does this impact application developers
 Non Applicable
2. Link to relevant issues and documentation
 https://github.com/blockstack/stacks.js/pull/1117

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

 Non Applicable

### DIFF
--- a/packages/wallet-sdk/src/derive.ts
+++ b/packages/wallet-sdk/src/derive.ts
@@ -176,14 +176,16 @@ const selectUsernameForAccount = async ({
   network?: StacksNetwork;
 }): Promise<{ username: string | undefined; derivationType: DerivationType }> => {
   // try to find existing usernames owned by stx derivation path
-  const address = deriveStxPrivateKey({ rootNode, index });
+  const stxPrivateKey = deriveStxPrivateKey({ rootNode, index });
+  const address = getAddressFromPrivateKey(stxPrivateKey);
   if (network) {
     let username = await fetchFirstName(address, network);
     if (username) {
       return { username, derivationType: DerivationType.Wallet };
     } else {
       // try to find existing usernames owned by data derivation path
-      const address = deriveDataPrivateKey({ rootNode, index });
+      const stxPrivateKey = deriveDataPrivateKey({ rootNode, index });
+      const address = getAddressFromPrivateKey(stxPrivateKey);
       username = await fetchFirstName(address, network);
       if (username) {
         return { username, derivationType: DerivationType.Data };
--- a/packages/wallet-sdk/src/models/wallet.ts
+++ b/packages/wallet-sdk/src/models/wallet.ts
@@ -70,15 +70,6 @@ export async function restoreWalletAccounts({
             salt: wallet.salt,
             stxDerivationType,
           });
-        } else {
-          existingAccount = {
-            ...existingAccount,
-            stxPrivateKey: derivePrivateKeyByType({
-              rootNode,
-              index,
-              derivationType: stxDerivationType,
-            }),
-          };
         }
         return {
           ...existingAccount,
@@ -117,15 +108,6 @@ export async function restoreWalletAccounts({
             salt: wallet.salt,
             stxDerivationType,
           });
-        } else {
-          existingAccount = {
-            ...existingAccount,
-            stxPrivateKey: derivePrivateKeyByType({
-              rootNode,
-              index,
-              derivationType: stxDerivationType,
-            }),
-          };
         }
         return {
           ...existingAccount,


ma-aboelenein:
@kyranjamie @kantai @friedger Hello Guys can you take a look at this i am trying the latest version of StacksJS (3.1.0) and it fails to restore the wallet accounts primarily  because: 
1- restoreWalletAccount now requires the network instance to be present 
2- there was a bug when fetching names that was trying to fetch it using the private key instead of the account address
I Hope this fix is sufficient and can get published ASAP

ma-aboelenein:
@kyranjamie @kantai @friedger Changing the stxPrivateKey that was introducd in the mentioned PR  has been removed in this [commit](https://github.com/blockstack/stacks.js/pull/1151/commits/caedea84dade24210a4058178bb0671ede9f5d47) the change caused inconsistency between the returned value from generateWallet function and restoreWalletAccounts function therefor producing a different stxAddress and can't resolve the account's holdings.

kyranjamie:
cc @friedger @pradel

friedger:
@ma-aboelenein Changing the stx address was the whole point of the PR. Can you please add a test that establish your findings?

ma-aboelenein:
@friedger i don't think that is necessary, the problem is the following after i generate the wallet from an existing seed phrase calling `restoreWalletAccounts` returns the same address for all the accounts instead the one containing a `.id.stx` name the address is different due to the wallet SDK replacing the `stxPrivateKey` of the account and thus resulting in a different `stxAddress` that doesn't contain the assets that i have in the account since it's registered under the old address, when i remove the part where the sdk regenerate the private key the name is fetched correctly due to the new introduced logic and the old `stxAddress` is returned as usual along with the assets.

ma-aboelenein:
@friedger Also the new way we resolve the names of the accounts should also be introduced in the auth library specifically in `makeAuthResponse` because it fails to verify the auth response in `doPublicKeysMatchUsername` for names that are not signed by the `walletKey`

ma-aboelenein:
@markmhx can you tag along in this?

friedger:
@ma-aboelenein correct, that is why the lookup of .btc names fails. This needs to be fixed by the Hiro wallet, the SDK provides the tools to do this in a flexible way

ma-aboelenein:
> @ma-aboelenein correct, that is why the lookup of .btc names fails. This needs to be fixed by the Hiro wallet, the SDK provides the tools to do this in a flexible way

Can you elaborate please where do that fall under the two comments or maybe hop on a call tomorrow to discuss this?

> @friedger i don't think that is necessary, the problem is the following after i generate the wallet from an existing seed phrase calling `restoreWalletAccounts` returns the same address for all the accounts instead the one containing a `.id.stx` name the address is different due to the wallet SDK replacing the `stxPrivateKey` of the account and thus resulting in a different `stxAddress` that doesn't contain the assets that i have in the account since it's registered under the old address, when i remove the part where the sdk regenerate the private key the name is fetched correctly due to the new introduced logic and the old `stxAddress` is returned as usual along with the assets.

> @friedger Also the new way we resolve the names of the accounts should also be introduced in the auth library specifically in `makeAuthResponse` because it fails to verify the auth response in `doPublicKeysMatchUsername` for names that are not signed by the `walletKey`

friedger:
@ma-aboelenein restoreWalletAccounts is an implementation of restoring where the derivation path for the stx address is selected depending on ownership of a username. If the user has a name registered with the default wallet address then that address is used. If the user has a name registered with the default data address then that address is used as wallet address and data address. If you want to use a different selection algorithm for the derivation paths, e.g. based on the balances of the wallet address and the data address, you need to implement your own `restoreWalletAccounts`. Maybe we should pass in the `selectStxDerivation` function as a parameter, if you think that wallet builders want to make different choices. You can also overwrite the wallet address by setting the private key based on your choice for the derivation path.
```
 stxPrivateKey: derivePrivateKeyByType({
              rootNode,
              index,
              derivationType: stxDerivationType,
            }),
```

Re profile signatures, all profiles need to be signed by the walletKey, i.e. by the key that owns the username, that is part of the protocol. The Hiro Wallet did sign with the data wallet in the past, therefore the verification method correctly rejects most of the profiles for .btc names. The Hiro Wallet needs to fix that by resigning the profiles.

packages/wallet-sdk/src/derive.ts
```diff
@@ -176,14 +176,16 @@ const selectUsernameForAccount = async ({
...
+  const stxPrivateKey = deriveStxPrivateKey({ rootNode, index });
+  const address = getAddressFromPrivateKey(stxPrivateKey);
   if (network) {
     let username = await fetchFirstName(address, network);
     if (username) {
       return { username, derivationType: DerivationType.Wallet };
     } else {
       // try to find existing usernames owned by data derivation path
-      const address = deriveDataPrivateKey({ rootNode, index });
+      const stxPrivateKey = deriveDataPrivateKey({ rootNode, index });
```
friedger, line 187: should be `dataPrivateKey`

packages/wallet-sdk/src/derive.ts
```diff
@@ -176,14 +176,16 @@ const selectUsernameForAccount = async ({
   network?: StacksNetwork;
 }): Promise<{ username: string | undefined; derivationType: DerivationType }> => {
   // try to find existing usernames owned by stx derivation path
-  const address = deriveStxPrivateKey({ rootNode, index });
+  const stxPrivateKey = deriveStxPrivateKey({ rootNode, index });
+  const address = getAddressFromPrivateKey(stxPrivateKey);
```
friedger, line 180: should respect the network

friedger:
@ma-aboelenein Here are the improved tests that would have catch this error: https://github.com/friedger/stacks.js/commit/f51614c4eff2ad4a00edcc34c5ef10ffb3cbd3ae

kyranjamie:
There are outstanding actions here based on @friedger's feedback

ma-aboelenein:
> There are outstanding actions here based on @friedger's feedback

@kyranjamie where do you just we move on from here, the private key changing procedure is a breaking change that i think we need to discuss a little bit more

friedger:
@ma-aboelenein The first part of your PR about the private key was fixed by https://github.com/hirosystems/stacks.js/commit/6bd5fd9bddf1b277a021a8b15cf82f5a336a9fa6

The second part of your PR (removing the stxPrivateKey) should not be merged because it would not respect usernames owned by other keys than the data key. With 3.1.1, we have a way that wallets can select which derivation path the user wants to use.

I think this PR can be closed.

ma-aboelenein:
> @ma-aboelenein The first part of your PR about the private key was fixed by [6bd5fd9](https://github.com/hirosystems/stacks.js/commit/6bd5fd9bddf1b277a021a8b15cf82f5a336a9fa6)
> 
> The second part of your PR (removing the stxPrivateKey) should not be merged because it would not respect usernames owned by other keys than the data key. With 3.1.1, we have a way that wallets can select which derivation path the user wants to use.
> 
> I think this PR can be closed.

@friedger i agree with you this is no longer relevant but can you elaborate more on the way wallets can select the derivation path

janniks:
Are there any issues remaining or can this be closed?

PR #1154, opened 2021-12-20
[Fix/1144] Populate username field in auth response

### LINKED ISSUE #1144: makeAuthResponse username is not populated for `.btc` names
**Describe the bug**

When the user login into an application via the Hiro wallet, the username field is an empty string while it should be the .btc domain.

**To Reproduce**

1. use the Hiro wallet to login into an app with an account that does have a .btc name
2. call `userSession.loadUserData()` and see that the `username` field is an empty string

**Expected behavior**

When the user is logging in into my app I should be able to read the owner .btc username from the `userSession.loadUserData()` function. Similarly to .id.stx names.

 - OS: Mac
 - Browser: chrome
 - Version 96.0.4664.55

ahsan-javaid:
## Description
This PR fixes the issue of empty username field in the auth response.

For details refer to issue #1144

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No
No

## Testing information
1. Test case added
2. `npm run test` passed successfully

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/auth/src/userSession.ts
+++ b/packages/auth/src/userSession.ts
@@ -347,6 +347,16 @@ export class UserSession {
       userData.profile = tokenPayload.profile;
     }
 
+    const address = userData.profile?.stxAddress?.mainnet;
+    if (!userData.username && address) {
+      try {
+        const namesResponse = await fetchPrivate(`${coreNode}/v1/addresses/stacks/${address}`);
+        const namesJson = await namesResponse.json();
+        if ((namesJson.names.length || 0) > 0) {
+          userData.username = namesJson.names[0];
+        }
+      } catch (e) {}
+    }
     sessionData.userData = userData;
     this.store.setSessionData(sessionData);
 
--- a/packages/auth/tests/auth.test.ts
+++ b/packages/auth/tests/auth.test.ts
@@ -255,6 +255,47 @@ test('auth response with invalid private key', async () => {
   expect(pass).toBeCalledTimes(2);
 });
 
+test('auth response with populated username field', async () => {
+  const username = 'test.btc';
+  fetchMock.mockResponse(JSON.stringify({ names: [ username ] }));
+  const appPrivateKey = makeECPrivateKey();
+  const transitPrivateKey = makeECPrivateKey();
+  const transitPublicKey = getPublicKeyFromPrivate(transitPrivateKey);
+  const metadata = {};
+  const authResponse = await makeAuthResponse(
+    privateKey,
+    {
+      '@type': 'Person',
+      '@context': 'http://schema.org',
+      apps: {
+        'http://localhost:3000': 'https://gaia.blockstack.org/hub/1JoKiUKHYqXGktzEzxPtR9khSNRPfwwp6R/'
+      },
+      appsMeta: {},
+      stxAddress: {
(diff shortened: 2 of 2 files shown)


packages/auth/src/userSession.ts
```diff
@@ -347,6 +347,16 @@ export class UserSession {
       userData.profile = tokenPayload.profile;
     }
 
+    const address = userData.profile?.stxAddress?.mainnet;
```
friedger, line 350: how does it work with testnet?

ahsan-javaid: https://github.com/blockstack/stacks.js/blob/fe831cf28d71de9410c81cfabe256fd511e9aa4d/packages/auth/src/userSession.ts#L227-L233

The `handlePendingSignIn` function is already configured with `mainnet`. It does not have any testnet param. So the implementation is aligned with existing auth flow. User may also set `coreNode` url there. 

**Secondly**

When we try to get bnsLookupUrl from testnet it will give you bnsLookupUrl of mainnet.
Here is code sample:
```
void (async () => {
  const n = new StacksTestnet();
  console.log(n.bnsLookupUrl);
})();

```

`Output: https://stacks-node-api.mainnet.stacks.co`


**Reason**: Refer to this commit: `fix: use only mainnet URLs during auth BNS checks`
https://github.com/blockstack/stacks.js/commit/1434262dd5ff34b8ef837fa33c4219eac57d0436

This proves to use only mainet urls for auth BNS checks. 
In case if we add network parameter in `handlePendingSignIn` then it will return mainet bnsLookupUrl on both testnet and mainnet as the reason is mentioned above.

Other than that if you want to purpose different thoughts then most welcome.

ahsan-javaid:
@kyranjamie @friedger Share your thoughts on this pr for further action items. Thanks 🙏

packages/auth/src/userSession.ts
friedger: The hard coded URLs are only fallback. Here the user does not have a chance to use testnet.

friedger:
Shouldn't this issue be solved in wallet.js on the authenticator?
The app can always try to fetch more data than thrbuser/authenticator provided but the auth should not do it.

ahsan-javaid:
@friedger Thanks for the review. 
Can you please point out to wallet.js/authenticator repo that you mentioned above. 🙏

friedger:
@ahsan-javaid Somewhere around here: https://github.com/hirosystems/stacks.js/blob/3f8fa11bde3f536d0dfd38cf6fddd1363d660a0e/packages/wallet-sdk/src/models/wallet.ts#L67

Newly created accounts are not populated, this should be changed, also it is never checked whether the user has registered a name by now.

The client should be able to trust the authenticator that the authenticator knows about registered usernames. If the auth response does not contain the username, the client can savely assume that the user does not have a username or that the user does not want to the app to use it. IMHO.

friedger:
Another good place could be https://github.com/blockstack/stacks-wallet-web/blob/f1567b6812b2a2e7de1532fbe7a690fbdc9166d2/src/app/query/account/account.query.ts#L27

ahsan-javaid:
Closing based on friedger comment to handle this on wallet repo.

PR #1156, opened 2022-01-04
[Bundle Size] Offload lodash dependency

### LINKED ISSUE #1120: Bundle size of the stacks package is really huge

ahsan-javaid:
## Description
This PR removes `lodash` dependency from stacks.js packages.  
`lodash` minified size is around `69KB` out of which stacks.js is using around `10KB` of its modules. This will help to reduce the bundle size. In the subsequent pr, further dependencies can be removed to reduce the bundle size further. 

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No

No

## Testing information

1. `npm run test`

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/transactions/package.json
+++ b/packages/transactions/package.json
@@ -27,7 +27,7 @@
   "devDependencies": {
     "@types/common-tags": "^1.8.0",
     "@types/jest": "^26.0.22",
-    "@types/lodash": "^4.14.160",
+    "@types/lodash.clonedeep": "^4.5.6",
     "common-tags": "^1.8.0",
     "jest": "^26.6.3",
     "jest-fetch-mock": "^3.0.3",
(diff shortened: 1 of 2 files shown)


kyranjamie (approved):

reedrosenbluth:
I'd prefer if we simply changed the dependency to be _just_ the `cloneDeep()` method from Lodash, rather than implementing our own.

https://www.npmjs.com/package/lodash.clonedeep

reedrosenbluth (changes requested):
Good idea! I think I'd prefer if we simply changed the dependency to be _just_ the `cloneDeep()` method from Lodash, rather than implement our own.

https://www.npmjs.com/package/lodash.clonedeep

PR #1158, opened 2022-01-09
feat: add tips option to readonly call

zhigang1992:
## Description

In stack API standard, there is a tip query params that will make sure the readonlyCall runs on up to date block.
Whiteout this params, it will return outdated data.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No breaking change, only adding new optional field

Typescript field added comment

## Testing information

1. when a new block just get mined
2. call readonlycall with updated tip params
3. it should now return up to date result

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated


zhigang1992:
@yknl

sry I'm not able to edit the reviewer fields anymore, will ping you here instead.

packages/transactions/src/builders.ts
```diff
@@ -1297,10 +1299,13 @@ export async function callReadOnlyFunction(
 
   const options = Object.assign(defaultOptions, readOnlyFunctionOptions);
 
-  const { contractName, contractAddress, functionName, functionArgs, network, senderAddress } =
+  const { contractName, contractAddress, functionName, functionArgs, network, senderAddress, tip } =
     options;
 
-  const url = network.getReadOnlyFunctionCallApiUrl(contractAddress, contractName, functionName);
+  let url = network.getReadOnlyFunctionCallApiUrl(contractAddress, contractName, functionName);
+  if (tip != null) {
```
kyranjamie: I know this will check both `null` and `undefined`, but imo it'd be clearer to be explicit and check both. Behaviour of `!=` vs `!==` is a bit of a js quirk that the next developer may not be familiar with. 

Alternatively, why not just flip and condition and check `typeof tip === 'number'`?

kyranjamie (approved):

janniks (approved):

zone117x (changes requested):
> Whiteout this params, it will return outdated data

This isn't really true. When the `tip` param is unspecified, the default is to return the latest _confirmed_ (non-microblock) state.

Also, the change in this PR is looking for a `number` type for `tip` which doesn't make sense. The `tip` param expects a hex string of the `index_block_hash`, _or_ the microblock hash. As of the latest stacks-node release, it also supports the string value `latest` which will use the latest unconfirmed microblock state.

See the OpenAPI docs here:
https://github.com/stacks-network/stacks-blockchain/blob/de541f9fc9bba0f2eccc647e729cfb5402476a38/docs/rpc/openapi.yaml#L223-L229

Or the rust code here if you want to dig deeper:
https://github.com/stacks-network/stacks-blockchain/blob/de541f9fc9bba0f2eccc647e729cfb5402476a38/src/net/http.rs#L1717-L1722

packages/transactions/src/builders.ts
```diff
@@ -1293,10 +1295,14 @@ export async function callReadOnlyFunction(
...
 
-  const { contractName, contractAddress, functionName, functionArgs, senderAddress } = options;
+  const { contractName, contractAddress, functionName, functionArgs, senderAddress, tip } = options;
 
   const network = StacksNetwork.fromNameOrNetwork(options.network);
-  const url = network.getReadOnlyFunctionCallApiUrl(contractAddress, contractName, functionName);
+  let url = network.getReadOnlyFunctionCallApiUrl(contractAddress, contractName, functionName);
+
+  if (typeof tip === 'number') {
+    url = `${url}?tip=${tip}`;
```
zone117x, line 1304: Blindly appending `?` to an opaque URL is bad idea. In this case, the above function that generates the url can already add query params:
https://github.com/hirosystems/stacks.js/blob/266f473276f209a2cdd7044313a64c5e40a93455/packages/network/src/index.ts#L76-L78

So you could end up with a broken query string like `myurl.com/endpoint?thing1?thing2`.

The URL construction should probably all be contained within that existing function, or at least use the `URL` class to safely append query params.

packages/transactions/src/builders.ts
```diff
@@ -1273,6 +1273,8 @@ export interface ReadOnlyFunctionOptions {
   network?: StacksNetworkName | StacksNetwork;
   /** address of the sender */
   senderAddress: string;
+  /** tip height **/
+  tip?: number;
```
zone117x, line 1277: Type of `number` does not match the RPC docs and afaik won't work and could potentially cause problems.

janniks:
Thanks for catching this @zone117x — I'll make sure to start comparing w/ rust/api when reviewing

kyranjamie (approved):

zhigang1992:
Thanks for the context, I think this PR isn't ready anymore. Will close for now.

The problem we have is that 

when the block is mined, if we immediately request for data via readonly-call, if it's processed by some node which haven't fully synced yet.

We will get an outdated data rather from the latest block.

PR #1160, opened 2022-01-17
[Bundle Size][Transactions] Use noble-secp256k1 to replace elliptic dependency

### LINKED ISSUE #1120: Bundle size of the stacks package is really huge

ahsan-javaid:
## Description
This PR replaces the `elliptic` dependency with `@noble/secp256k1` in transaction package. 

Most of the bundle size is comming from `elliptic` library around `130kb`.
Reference: https://bundlephobia.com/package/elliptic@6.5.4

`@noble/secp256k1` is fast and small in size: https://bundlephobia.com/package/@noble/secp256k1@1.4.0

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No

No

## Testing information

1. `npm run test`

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated


saralab:
@kantai  @zone117x  Can you please take a peek at this one, involves secp256k1 changes.

packages/common/src/utils.ts
```diff
@@ -397,3 +397,39 @@ export function intToBigInt(value: IntegerType, signed: boolean): bigint {
 export function with0x(value: string): string {
   return !value.startsWith('0x') ? `0x${value}` : value;
 }
+
+export function hexToBigInt(hex: string): bigint {
+  if (typeof hex !== 'string')
+    throw new TypeError('hexToNumber: expected string, got ' + typeof hex);
+  // Big Endian
+  return BigInt(`0x${hex}`);
+}
```
kantai, line 411: Can you add a docs comment to the function specifying that the `hex` input is supplied without the `0x` prefix and in big endian?

packages/common/src/utils.ts
```diff
@@ -397,3 +397,39 @@ export function intToBigInt(value: IntegerType, signed: boolean): bigint {
...
+    throw new TypeError('hexToBytes: expected string, got ' + typeof hex);
+  if (hex.length % 2)
+    throw new Error(`hexToBytes: received invalid unpadded hex, got: ${hex.length}`);
+  const array = new Uint8Array(hex.length / 2);
+  for (let i = 0; i < array.length; i++) {
+    const j = i * 2;
+    array[i] = Number.parseInt(hex.slice(j, j + 2), 16);
+  }
+  return array;
+}
```
kantai: What happens if this method is supplied a `0x` prefixed string? What happens if it's supplied a non-hex string?

packages/transactions/src/keys.ts
```diff
@@ -108,19 +108,32 @@ export function serializePublicKey(key: StacksPublicKey): Buffer {
...
+    if (data[data.length - 1] !== 1) {
+      throw new Error(
+        'Improperly formatted private-key. 33 byte length usually ' +
+          'indicates compressed key, but last byte must be == 0x01'
+      );
+    }
+    compressed = true;
+  }
+  return compressed;
+}
```
kantai: Did the library previously default to compressed or default to uncompressed?

kantai: Nevermind -- I see this is the correct behavior. Can you update `createStacksPrivateKey` to invoke this method rather than duplicate the logic?

packages/transactions/src/keys.ts
```diff
@@ -158,26 +171,27 @@ export function createStacksPrivateKey(key: string | Buffer): StacksPrivateKey {
...
 
-export function makeRandomPrivKey(entropy?: Buffer): StacksPrivateKey {
-  const ec = new EC('secp256k1');
-  const options = { entropy: entropy || randomBytes(32) };
-  const keyPair = ec.genKeyPair(options);
-  const privateKey = keyPair.getPrivate().toString('hex', 32);
-  return createStacksPrivateKey(privateKey);
+export function makeRandomPrivKey(): StacksPrivateKey {
+  return createStacksPrivateKey(utils.bytesToHex(utils.randomPrivateKey()));
 }
```
kantai, line 174: Perhaps someone else should comment on the change in function signature here -- was the entropy argument ever used by library clients?

packages/transactions/tests/keys.test.ts
```diff
@@ -70,10 +70,10 @@ test('Retrieve public key from signature', () => {
   const compressedPubKey = '03ef788b3830c00abe8f64f62dc32fc863bc0b2cafeb073b6c8e1c7657d9c2c3ab';
 
   const message = 'hello world';
-  const sig = signWithKey(privKey, message);
-
-  const uncompressedPubKeyFromSig = publicKeyFromSignature(message, sig, PubKeyEncoding.Uncompressed)
-  const compressedPubKeyFromSig = publicKeyFromSignature(message, sig, PubKeyEncoding.Compressed)
+  const messageHash = bytesToHex(utf8ToBytes(message));
```
kantai: This isn't actually a `messageHash`, just the message encoded in hex.

packages/transactions/tests/keys.test.ts
```diff
@@ -70,10 +70,10 @@ test('Retrieve public key from signature', () => {
   const compressedPubKey = '03ef788b3830c00abe8f64f62dc32fc863bc0b2cafeb073b6c8e1c7657d9c2c3ab';
 
   const message = 'hello world';
-  const sig = signWithKey(privKey, message);
```
kantai, line 73: Did `signWithKey` used to work with non-hex encoded strings? If so, is this a change in that method's semantics?

packages/transactions/src/keys.ts
zone117x: IIRC it was used for deterministic unit tests at some point.

packages/transactions/src/keys.ts
```diff
@@ -139,45 +159,51 @@ export interface StacksPrivateKey {
...
+      return createStacksPrivateKey(utils.bytesToHex(b32));
+  }
+  throw new Error('Valid private key was not found in 8 iterations. PRNG is broken');
+}
+
+function verifyHex(input: string): string {
+  const regexp = /^[0-9a-fA-F]+$/;
+  let hex = input;
+  if (!regexp.test(input)) {
+    hex = utils.bytesToHex(utf8ToBytes(input));
```
zone117x: The name of the function is inaccurate, and the functionality is not clear to me. It appears to be a heuristic to detect hex strings, and if a non-hex is detected then it treats the input as utf8? That is surprising behavior which should be avoided.

packages/common/src/utils.ts
zone117x: This was marked as resolved but doesn't appear to be resolved

packages/transactions/src/keys.ts
zone117x: For example, if this function is being used for the utf8 functionality, then a message like `1234` will be detected as hex.

packages/transactions/src/keys.ts
```diff
@@ -139,45 +159,56 @@ export interface StacksPrivateKey {
...
+  const regexExp = /^[a-f0-9]{64}$/gi;
+  return regexExp.test(input);
+}
+
+function toHash(input: string): string {
+  return utils.bytesToHex(utf8ToBytes(input));
+}
+
+function getHash(input: string): string {
+  return isHash(input) ? input : toHash(input);
```
zone117x: This is another heuristic that looks strange to me. What is the purpose of this function? I can see what the code does, but why?

packages/transactions/src/keys.ts
```diff
@@ -139,45 +159,56 @@ export interface StacksPrivateKey {
...
+
+function isHash(input: string): boolean {
+  // Regular expression to check if string is a SHA256 hash
+  const regexExp = /^[a-f0-9]{64}$/gi;
+  return regexExp.test(input);
+}
+
+function toHash(input: string): string {
+  return utils.bytesToHex(utf8ToBytes(input));
+}
```
zone117x: Why is this called `toHash`? There's no hashing being done here.

packages/transactions/src/keys.ts
```diff
@@ -139,45 +159,56 @@ export interface StacksPrivateKey {
...
+  const bytes: Uint8Array = new Uint8Array(entropy?.length || 0);
+  if (entropy) {
+    for (let i = 0; i < entropy.length; i++) bytes[i] = entropy[i];
+  }
+  let i = 8;
+  while (i--) {
+    const b32 = bytes.length ? bytes : utils.randomBytes(32);
+    const num = BigInt(`0x${utils.bytesToHex(b32)}`);
+    // Check isWithinCurveOrder
+    if (BigInt(0) < num && num < CURVE.n && num !== BigInt(1))
```
janniks: Why are we excluding `BigInt(1)` from the range?

> Also: I believe, `BigInt(0) < num && num !== BigInt(1)` can be simplified to `BigInt(1) < num`

zone117x: This function code was pulled from the `secp256k1` lib right? There's too much potentially sensitive code here IMO and should be left in the lib. If it's only unit tests requiring the entropy arg then those tests should be changed to work around this.

ahsan-javaid:
@zone117x Let me know if you have any further thought's on this pr. 🙏

ahsan-javaid:
@zone117x Just a small reminder if you get a chance to give it a quick look. Thanks.

packages/transactions/src/keys.ts
```diff
@@ -31,6 +37,12 @@ import {
 import { BufferReader } from './bufferReader';
 import { c32address } from 'c32check';
 
+utils.hmacSha256Sync = (key: Uint8Array, ...msgs: Uint8Array[]) => {
+  const h = hmac.create(sha256, key);
+  msgs.forEach(msg => h.update(msg));
+  return h.digest();
+};
```
zone117x, line 51: Why is a lib's function being defined here? Can you add a comment?

ahsan-javaid: Done!

zone117x: Thanks!

zone117x (approved):
LGTM!

zone117x (changes requested):
Actually could you add a couple tests that use the previous lib to sign a message, the use the new lib to verify? And vice versa? Elliptic.js would be required as a devDependency

ahsan-javaid:
> Actually could you add a couple tests that use the previous lib to sign a message, the use the new lib to verify? And vice versa? Elliptic.js would be required as a devDependency

Done!

zone117x (approved):

kantai (approved):

janniks (approved):
Looks Good 👍

But, I noticed some new [versions of `secp256k1`](https://github.com/paulmillr/noble-secp256k1/releases) have been released, since the PR was opened.

ahsan-javaid:
> Looks Good 👍
> 
> But, I noticed some new [versions of `secp256k1`](https://github.com/paulmillr/noble-secp256k1/releases) have been released, since the PR was opened.

Done!

ahsan-javaid:
@janniks Do you have any further thought's on this. Thanks 🙏

PR #1168, opened 2022-01-28
[ignore] remove network package reliance

### LINKED ISSUE #1161: Build Transactions without Network object
It would be great if we could use the transaction builder functions without having to pass in a `StacksNetwork` object, so a dependency on the `network` package isn't required for building transactions.

janniks:
_moved to https://github.com/hirosystems/stacks.js/pull/1169_


janniks:
https://github.com/hirosystems/stacks.js/pull/1169

PR #1169, opened 2022-01-28
reduce network package reliance

### LINKED ISSUE #1161: Build Transactions without Network object

janniks:
## scope
- closes #1161  
- simplifies developer experience by "hiding" the network package for simple calls

---

## changes
- renames `StacksNetwork` to `IStacksNetwork` to change inheritance (previously `StacksMainnet` was parent of `StacksTestnet` and `StacksMocknet`)
  - maybe `I` prefix should be used for interfaces that also have a object-style class. we can keep name without prefix when they are just an interface for data or similar. similarly this is already being done for `IdentifyInterface`(suffix) in **keychain** and `IDAppKeys`(prefix) in **utils**.
- updates symbols search for IDE/vscode use with typescript (to ignore `dist` folder)
- adds static methods for building and deriving a Network object (e.g. from a string name, like `testnet`)
- auto-detects the wanted Network from the transaction during broadcast (to eliminate a duplicate statement) — network is still optionally specifiable for advanced purposes
- minor refactors for clearness (mostly guard clauses, etc.)

---

## tangential ideas
- typescript docstrings (via [typedoc](https://typedoc.org)) don't require `@param {type}`, as these can be inferred from code. these duplicate type definitions could be removed from all docstrings, as it requires changes to be made in multiple parts (easily missed). i tested this locally; it works for removing type but not for removing parameter name (even though this could be inferred as well). — not sure where else these additional type annotations are used, i am assuming that the ts language server always uses code over annotation...

---

## next steps
- we could further consolidate the "network hiding" in other areas of the code base
  - e.g. https://github.com/hirosystems/stacks.js/issues/1170


reedrosenbluth (approved):
LGTM! Awesome work!

packages/transactions/tests/builder.test.ts
```diff
@@ -921,7 +1014,7 @@ test('make a multi-sig contract call', async () => {
   );
 });
 
-test('Estimate transaction transfer fee', async () => {
+test.skip('Estimate transaction transfer fee', async () => {
```
janniks, line 1017: The `.skip` tests were previously not run and are now outdated/fail. https://github.com/hirosystems/stacks.js/issues/1175 addresses fixing them in a separate PR.

packages/auth/src/profile.ts
```diff
@@ -1,11 +1,11 @@
 import { resolveZoneFileToProfile } from '@stacks/profile';
 import { fetchPrivate } from '@stacks/common';
-import { StacksNetwork, StacksMainnet } from '@stacks/network';
+import { IStacksNetwork, StacksMainnet, StacksNetwork, StacksNetworkName } from '@stacks/network';
 
 export interface ProfileLookupOptions {
   username: string;
   zoneFileLookupURL?: string;
-  network?: StacksNetwork;
+  network?: StacksNetworkName | IStacksNetwork;
```
kyranjamie: Generally speaking elsewhere in Hiro's org typescript we're not following the `I` prefixed interface naming convention (keychain is as good as deprecated).

I see your comment about it, but I'd push to avoid this. It'll introduce a inconsistency, it isn't used in TS docs, and also odd when the type itself also begins with an `I` > `IIdentity`. I used to work on a project using this convention and haven't missed it.

There are bigger fish to fry, so your call, but my 2 euro cents.

packages/network/src/index.ts
```diff
@@ -8,7 +8,10 @@ export interface NetworkConfig {
   url: string;
 }
 
-export interface StacksNetwork {
+export const StacksNetworks = ['mainnet', 'testnet'] as const;
+export type StacksNetworkName = typeof StacksNetworks[number];
+
```
kyranjamie, line 13: Repeated `StacksNetworkName | IStacksNetwork;` type, maybe we could make a reusable type of it?

```suggestion

export type StacksNetwork = StacksNetworkName | IStacksNetwork;
```

packages/network/src/index.ts
```diff
@@ -62,10 +66,35 @@ export class StacksMainnet implements StacksNetwork {
...
+      case 'testnet':
+        return new StacksTestnet();
+      default:
+        throw new Error(
+          `Invalid network name provided. Must be one of the following: ${StacksNetworks.join(
+            ', '
+          )}`
+        );
+    }
+  };
```
kyranjamie: One of the gripes with stacks.js picked up on in the past has been the inconsistent mix between OO and FP styles. Network already uses class style, so adding this static function just follows this style.

Though one thing to consider is the serialisability of these objects. In `@stacks/connect` we pass a network object --> the wallet via a custom event/postMessage comms (which auto-serialises it). We've ran into bugs with this, and also makes the types less useful (there is no type for the result of `JSON.serialize(new StacksMainnet())`)

No action necessary here, but good to be aware of. Perhaps this could also work as a separate fn definition?

packages/network/src/index.ts
```diff
@@ -62,10 +66,35 @@ export class StacksMainnet implements StacksNetwork {
...
+
+  static fromNameOrNetwork = (network: StacksNetworkName | IStacksNetwork) => {
+    if (StacksNetworks.includes(network as StacksNetworkName)) {
+      // network is StacksNetworkName
+      return StacksNetwork.fromName(network as StacksNetworkName);
+    }
+
+    // network is IStacksNetwork
```
kyranjamie: this comment doesn't add any more detail than the code it describes

kyranjamie (approved):

packages/network/src/index.ts
janniks: Good point! Generally, I am trying to move to a more functional approach, especially when composition is useful.

Yes, here we already have classy code. 😉 Technically the new methods are `static` so we could separate them. Yet, it's also nice for developers to be able to benefit from the discoverability of typing `StacksNetwork.` to auto-complete builder-style methods. In general, I am also thinking about namespaces or similar TS techniques to group together functional methods — to increase discoverability.

janniks: For example, I like the way `clarinet` does this for [`types.`](https://github.com/hirosystems/clarinet/blob/8254f51ed33faa280ff5c7297b2312c2a4f5e9e6/deno/index.ts#L273) But I'm reading mixed things about namepsaces/modules/es6-modules.

packages/network/src/index.ts
janniks: Thanks for bringing this up. Sadly, TS `match`ing on types only works in funky ways. I added some of these comments to explicitly show which type we have at this point in code. Firstly, for when it's not immediately clear what the type is to the dev, but TS knows, e.g. using `'senderKey' in txOptions`. Secondly, for when even TS can't accurately detect the type, even if it's clear.

Not sure if this is a good idea or not — comments can also become stale and potentially reflect incorrect types at some point.

In any case, I'll refactor this to better utilise TS.

packages/network/src/index.ts
janniks: Here, I'd like to keep the verbose type for better DX (hover, docs, etc.) — developer will be shown both types and `StacksNetworkName` auto-expands to all its string options.

packages/auth/src/profile.ts
janniks: I went ahead and simplified this to just a single `class` now. The additional interface/class was making things too complicated here. Now we just have a class that is extended by `..Mainnet`/etc. and can also be used as a type — rather than having duplicate information with the interface+impl.

janniks:
CC @zone117x

zone117x (approved):
LGTM! Solid DX improvements 👍 
Thanks @janniks!

PR #1181, opened 2022-02-09
update release ci

### LINKED ISSUE #1057: Publishing/release automation
For our other projects, we have some process for automating releases. Looking at `stacks.js`, there isn't any automation currently (eg, merge to main -> production deployment to npm).

Does splat have interest in this, or is it there a need to have the manual publishing step?

Additionally, would it be possible to publish new versions of all packages that have changes? There are many changes on `master` that have yet to be published. cc @reedrosenbluth @zone117x

janniks:
## scope
- adds `prerelease` publishing on `master` merge
- closes https://github.com/hirosystems/stacks.js/issues/1057

## comments
might be a bit hard to "test" and require a few tries

### based on
- https://github.com/hirosystems/gh-actions-example/blob/main/.github/workflows/ci.yml
- https://github.com/actions/checkout#push-a-commit-using-the-built-in-token

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,9 +41,9 @@ jobs:
         run: |
           git config user.name github-actions
           git config user.email github-actions@github.com
-          lerna version --conventional-commits --conventional-prerelease --yes
+          lerna version --conventional-commits --conventional-prerelease --preid beta --yes
 
       - name: "Release: Publish"
-        run: lerna publish from-git --yes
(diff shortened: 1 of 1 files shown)


kyranjamie (approved):
So no more semantic release?

janniks:
Kind of — semantic-release had been disabled on stacks.js for a while, since it seemed to not work well with lerna.

I believe that lerna now supports everything we need from semantic-release out-of-the-box.

PR #1184, opened 2022-02-11
add browser polyfill

### LINKED ISSUE #1162: Utilise the `browser` field of the package.json to deliver pre-polyfilled browser modules
Webpack 5 doesn't support auto-polyfill for node.js modules like stream and crypto that are used in stacks.js
In an attempt to enable the SDK work seamlessly with both node and browser, it will be great to deliver pre-polyfilled browser modules.

janniks:
---
> well... i hate bundlers now 💢
---

## changes
- see https://github.com/hirosystems/stacks.js/issues/1162 for context
- adds a `umd` bundle with all dependencies (also internal) bundled up for plug-and-play
- adds a `esm` bundle with external dependencies bundled up for plug-and-play
- set `browser` field in package.json to point to new `polyfill` bundle
- add brief readme notes to explain types and usage with frameworks (ideally, we will add real working examples next/soon)
- switch webpack to es-loader (speedup from >200s to ~100s)

> this makes frameworks like React use the `polyfill` bundle by default instead of throwing errors on unfound node modules.

> `@stacks/encryption` need `browser: umd` for now, since there is an issue in `elliptic` that common esm bundlers can't resolve. this can be addressed as soon as elliptic is removed as a dep

## reasoning
- we are putting compatibility before bundle size (until we have removed all `fallback` requiring code)
- consumers that want to roll their own polyfills or need specific imports can use `import { ... } from '@stacks/transactions/dist/esm'` to manually access the esm modules

### steps
as a goal i wanted to keep esm compatible file structures for advanced use (which e.g. the wallet team likely needs/wants). so the polyfilled bundle should either be a second esm bundle OR replace the existing cjs/umd. it turns out the umd bundles are not currently being shipped/deployed — not quite sure why (might have to do with the build steps/scripts in lerna). after comparing cjs and umd in React (using create-react-app), umd resulted in less warnings. it also can make more sense as a "universal" bundle (usable in <script> tags), compared to cjs, which is typically the node native module system.

i also compared the different solutions used for bundling/building libraries in hirosystems repos. `tsdx` looks promising as a modern library build tool and is already being used in https://github.com/hirosystems/ui. unfortunately, i couldn't get it running smoothly with the stacks.js monorepo. it is also very constrained by its opinions on typescript code and not flexible in its output format. with more effort these kinks could be likely be fixes or worked around.

webpack and other modern alternatives are designed more for bundling, rather than library authorship and lose many needed features when dealing with typescript and esm.

rollup is promising (`tsdx` also uses rollup internally), but [isn't used a lot](https://sourcegraph.com/search?q=context:global+repo:github%5C.com/hirosystems+rollup&patternType=literal) in hirosystems repos.

**update:** i'm using rollup now for esm bundling, after seeing the imperfections of the `umd` bundle, i wasn't happy enough with it. following way to many hours of config-wordle guessing — it works! ⬜️🟨🟩

### learnings

- react respects `browser` > `module` > `main`
- next.js mixes in cjs for SSR
- vite / rollup don't always respect `browser` field and prefer esm
- no bundler can really tree-shake umd

→ greatest compatibility requires esm! 

### todo
- [x] get umd to work! 😠
- [x] analyze bundle size w/ umd module in next.js
- [x] test with vite + vue
- [x] test with svelte
- [x] ~test tree-shakeability w/ next js~ umd not shakeable 🙃
- [x] add second esm output as bundle with inline polyfills

### thoughts

- `bigint` could also be polyfilled at this point

---

⚠️ these `umd` bundles are huge. this is not a great solution, but compatibility currently takes priority

🚛 analyzing page bundles (for react, vue, and svelte) showed pages ranging from 400kb to 10mb 🤯

| package              |   size `umd` | size `umd` polyfilled | size `polyfill` |
| :------------------- | -----------: | --------------------: | --------------: |
| @stacks/common       |   `88.9 KiB` |                       |      `87.64 KB` |
| @stacks/network      |   `90.7 KiB` |                       |       `3.33 KB` |
| @stacks/bns          |    `274 KiB` |                       |       `8.38 KB` |
| @stacks/transactions |    `299 KiB` |                       |     `804.52 KB` |
| @stacks/profile      |    `437 KiB` |                       |     `709.76 KB` |
| @stacks/stacking     |  ~`549 KiB`~ |             `551 KiB` |     `520.57 KB` |
| @stacks/encryption   |  ~`606 KiB`~ |             `824 KiB` |     `654.52 KB` |
| @stacks/auth         |  ~`996 KiB`~ |            `1.17 MiB` |      `711.4 KB` |
| @stacks/storage      | ~`1.95 MiB`~ |            `2.39 MiB` |     `845.09 KB` |
| @stacks/keychain     | ~`4.47 MiB`~ |             `5.5 MiB` |       `1.13 MB` |
| @stacks/wallet-sdk   |  ~`4.5 MiB`~ |            `5.43 MiB` |       `1.16 MB` |


configs/polyfills/brorand.js
```diff
@@ -0,0 +1,75 @@
+/*
```
janniks, line 1: converted from cjs-like to esm from https://github.com/indutny/brorand (needed for rollup bundling)

janniks:
sorry, @zone117x @kyranjamie this one is a bit annoying 😉 but basically just configuration change. let me know if you have additional ideas on solving bundle problems or see any issues — would love to get this right.

so far, nothing is really perfect/great (until we get rid of more dependencies). that's going to be the next big task, while in parallel adding some better example/docs

kyranjamie (approved):
Good work @janniks, I haven't tested out this PR myself but like the approach you've taken re: creating different outputs for polyfilled bundles.

Would your configs benefit at all from using [webpack-merge](https://www.npmjs.com/package/webpack-merge)?  

I don't really understand this borand stuff. Why is this needed exactly?

janniks:
> Would your configs benefit at all from using [webpack-merge](https://www.npmjs.com/package/webpack-merge)?

Nice package! Will keep this in mind if we need to object merge configurations — right now it's fairly straightforward to merge, since it's just setting `name` and `fallback` which both don't need to exist in the base config.

> I don't really understand this borand stuff. Why is this needed exactly?

This was a bit annoying, yes. Basically, `brorand` is used in a dependency of `crypto-browserify` (and `elliptic`). It's written in a legacy cjs style that rollup can't understand. I couldn't find an existing shim/polyfill for it, so I manually converted the exports to a esm style, while keeping the logic the same. Unfortunately, in elliptic `brorand` uses `crypto`, resulting in a circular dependency (with initialization) that rollup can't get around `crypto ∖ crypto-browserify → brorand → crypto ...`. So elliptic is currently not usable in `polyfill` mode 😞 cjs is a lot more forgiving so everything works in the `umd` bundles

zone117x (approved):
The new dist outputs look great! It seems like a lot of code and dependencies were added just to deal with a handful of problematic libraries like bitcoinjs-lib, elliptic, etc. Given that we know these can be eliminated (see micro-stacks), it seems like we're possibly ordering tasks incorrectly. If we merge this as-is, are you anticipating rolling back / refactoring a lot of the changes _after_ we remove the problematic deps?

janniks:
Yes, definitively a valid point! I had similar [thoughts/doubts](https://github.com/hirosystems/stacks.js/issues/1162#issuecomment-1031626164). For now, the plan is to treat this as a short term bandaid. Ideally, for every change we make going forward we eliminate any required polyfill from the build steps (and even if we forget to at first, most polyfills wouldn't show up in the output if there's no need for them).

Issue #1195, opened 2022-02-22
Allow wallets to specify the derivation path for stx private key


friedger:
**Is your feature request related to a problem? Please describe.**
Currently, wallet accounts are restored such that the stxPrivateKey is either the wallet private key or the data private key depending on whether the user owns a BNS username on one of the corresponding addresses. 

However, wallets might want to only support one derivation path.

**Describe the solution you'd like**
The function `restoreWalletAccounts` should take an extra parameter with the derivation type for the stx private key. If Unknown derivation type is provided, this the function tries to restore the accounts as now, If a different derivation type is provided, the accounts only look at the provided derivation path.

**Describe alternatives you've considered**
This looks like the change with the least impact.

**Additional context**
Main discussion happened in https://github.com/hirosystems/stacks-wallet-web/pull/2089

@kyranjamie

kyranjamie:
Should this issue also revert the previous changes that change the derivation based on the name's lookup?

friedger:
@kyranjamie We can keep them to make it possible for other wallets to choose their own derivation path

kyranjamie:
>  We can keep them to make it possible for other wallets to choose their own derivation path 

but specifically the logic which automatically determines the derivation path, by the username, won't be needed, right? This'll be changing this to only accept derivation path as an input?

friedger:
Yes, restoreWalletAccounts would only use stxPrivateKey for account derivation.

As restoreWalletAccounts is very specific, an extra paramater does not make sense.

PR #1196, opened 2022-02-22
feat: on restoreWalletAccounts, only support stxPrivateKey

friedger:
## Description

This PR changes the way how wallet accounts are restored by default: Instead of detecting existing usernames, accounts will always use the stx private key to lookup usernames and restore the account with that name.

For details refer to issue #1195 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

The way of restoring wallet accounts has reverted to pre 3.1.

Documentation does not describe the restoring process of wallets

## Testing information
Use a wallet, e.g. Hiro Wallet with this library and 
1. restore wallets
2. check whether the username corresponds to the address of the stx private key
3. check whether a username owned by the data private key is not shown.

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x] Tag 1 of @kyranjamie  for review

### DIFF
--- a/packages/wallet-sdk/src/derive.ts
+++ b/packages/wallet-sdk/src/derive.ts
@@ -214,7 +214,7 @@ export const fetchUsernameForAccountByDerivationType = async ({
   rootNode: BIP32Interface;
   index: number;
   derivationType: DerivationType.Wallet | DerivationType.Data;
-  network?: StacksNetwork;
+  network: StacksNetwork;
 }): Promise<{
   username: string | undefined;
 }> => {
--- a/packages/wallet-sdk/src/derive.ts
+++ b/packages/wallet-sdk/src/derive.ts
@@ -87,7 +87,8 @@ export enum DerivationType {
 
 /**
  * Tries to find a derivation path for the stxPrivateKey for the account
- * defined by rootNode and index that respects the username of that account.
+ * defined by rootNode and index that respects the username of that account
+ * and that respects the given derivationType.
  *
  * The stxPrivateKey is used to sign the profile of the account, therefore,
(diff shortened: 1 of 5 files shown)


packages/wallet-sdk/src/derive.ts
```diff
@@ -197,9 +201,38 @@ const selectUsernameForAccount = async ({
...
+    });
+    const privateKey = derivePrivateKeyByType({ rootNode, index, derivationType });
+    const address = getAddressFromPrivateKey(privateKey, txVersion);
+    const username = await fetchFirstName(address, network);
+    return { username };
+  } else {
+    return { username: undefined };
+  }
+};
+
```
kyranjamie: Should we make `network` a require parameter, as it's pretty pointless to use this function without one.

kyranjamie (approved):

packages/wallet-sdk/src/derive.ts
friedger: Fixed in 9b261525dce70f4e4463eccc2b1feeed178f1458

pradel:
hey there, any update on this? As it's removing the data key derivation it should be fine to merge and release it to get it included in the stacks wallet web, or is something still missing/blocking?
We need this fix to get the profiles.json properly signed for .btc usernames :)

kyranjamie:
@janniks would you be able to help @friedger here fix the type error in tests and get this merged 🙏🏼

janniks:
Sure, I can take over 👍🏻

janniks:
forked here: https://github.com/hirosystems/stacks.js/pull/1204

PR #1219, opened 2022-03-21
Offload `bitcoinjs` in @stacks/encryption

### LINKED ISSUE #1197: Offload `bitcoinjs-lib` dependency
May be split into sub-issues to modularize over @stacks/packages

janniks:
## scope
removes the need for `bitcoinjs` as a dependency

## comments
- the added `bs58` package was already previously a nested dependency (i.e. is not really a new dependency)
- some inline-comments for context in diff


packages/keychain/src/address-derivation/index.ts
```diff
@@ -19,8 +19,7 @@ export function deriveStxAddressChain(chain: ChainID) {
     if (!childKey.privateKey) {
       throw new Error('Unable to derive private key from `rootNode`, bip32 master keychain');
     }
-    const ecPair = ECPair.fromPrivateKey(childKey.privateKey);
```
janniks, line 22: `ECPair.fromPrivateKey( ...` without options provided, would implicitly compress the passed private key — hence now the added `compressPrivateKey` wrap in many places

packages/keychain/src/identity.ts
```diff
@@ -70,15 +67,16 @@ export class Identity implements IdentifyInterface {
       if (!profile.apps) {
         profile.apps = {};
       }
-      const challengeSigner = ECPair.fromPrivateKey(Buffer.from(appPrivateKey, 'hex'));
-      const storageUrl = `${hubInfo.read_url_prefix}${ecPairToAddress(challengeSigner)}/`;
+      const publicKey = getPublicKeyFromPrivate(appPrivateKey);
```
janniks, line 70: explicit `compress` not needed here, as `getPublicKeyFromPrivate` also implicitly compresses

packages/transactions/src/keys.ts
```diff
@@ -159,13 +140,15 @@ export function deserializePublicKey(bufferReader: BufferReader): StacksPublicKe
...
+  // it actually means: should public keys be generated as "compressed" or "uncompressed" from this private key
   compressed: boolean;
+  data: Buffer;
 }
 
 export function createStacksPrivateKey(key: string | Buffer): StacksPrivateKey {
-  const data = typeof key === 'string' ? Buffer.from(key, 'hex') : key;
-  const compressed: boolean = isPrivateKeyCompressed(key);
+  const data = privateKeyToBuffer(key);
+  const compressed = data.length == PRIVATE_KEY_COMPRESSED_LENGTH;
```
kantai, line 151: Should this also check that `data[32] == 0x01`?

janniks: The previous `privateKeyToBuffer` ensures that the resulting `data` is of the correct length (either 32 or 33) and ends in a 1 byte if of 33 byte length.

janniks: We have an interesting mix of buffers, strings, and private key objects. Ideally, I would like to move to a single consistent type for internal methods — but that will be addressed in a separate PR some time in the future. This one is already touching a lot.

kantai: Yep -- that definitely makes sense.

kantai (approved):
LGTM!

zone117x (approved):
LGTM!

PR #1230, opened 2022-04-11
fix: remove username matching

janniks:
## scope
**EDIT:** now removes the `username` field completely from `auth`

~removes username matching from auth package (application-side)
currently — in applications using @stacks/connect — the @stacks/auth package will "validate" usernames which the wallet returned in the authResponse. as "validation" these usernames are matched to the response of the API's `/names` endpoint.~

~the matching is removed from [`verifyAuthResponse`](https://github.com/hirosystems/stacks.js/pull/1230/files#diff-916fca8273bf9c9b085f7a99720e76d1d5a974da0b1b237b4d9ff2d9686b48fe). in addition, the previously used `doPublicKeysMatchUsername` [seems incorrect](https://github.com/hirosystems/stacks-wallet-web/issues/2354#issuecomment-1093026523) and is removed completely (only used in `verifyAuthResponse` previously).~

## comments
~a second PR could investigate removing the `username` field from auth responses in general, as brought up [here](https://github.com/hirosystems/stacks.js/issues/1144#issuecomment-1099968763) (should be coordinated w/ wallet-web).~

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21494,7 +21494,6 @@
 				"@stacks/encryption": "^3.3.0",
 				"@stacks/network": "^3.3.0",
 				"@stacks/profile": "^3.3.0",
-				"c32check": "^1.1.3",
 				"cross-fetch": "^3.1.4",
 				"jsontokens": "^3.0.0",
 				"query-string": "^6.13.1"
@@ -24526,7 +24525,6 @@
 				"@stacks/network": "^3.3.0",
 				"@stacks/profile": "^3.3.0",
 				"@types/jest": "^26.0.22",
-				"c32check": "^1.1.3",
 				"cross-fetch": "^3.1.4",
 				"jest": "^26.6.3",
 				"jest-fetch-mock": "^3.0.3",
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -44,7 +44,6 @@
     "@stacks/encryption": "^3.3.0",
     "@stacks/network": "^3.3.0",
     "@stacks/profile": "^3.3.0",
-    "c32check": "^1.1.3",
     "cross-fetch": "^3.1.4",
     "jsontokens": "^3.0.0",
     "query-string": "^6.13.1"
--- a/packages/auth/src/index.ts
+++ b/packages/auth/src/index.ts
@@ -6,7 +6,6 @@ export {
   verifyAuthResponse,
   isExpirationDateValid,
   isIssuanceDateValid,
-  doPublicKeysMatchUsername,
   doPublicKeysMatchIssuer,
   doSignaturesMatchPublicKeys,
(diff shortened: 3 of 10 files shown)


janniks:
cc'ing @kantai and @zone117x for any potential issues with this change that haven't come up yet

janniks:
cc'ing @zone117x for any potential issues with this change that haven't come up yet

kyranjamie (approved):
Nice improvement @janniks. 

This never made sense to me: if the registrar is down, you can't verify an auth response? 🤷🏼

markmhendrickson:
@janniks Any ETA on when this PR might get released? We're hoping to incorporate it into out test app so we can QA https://github.com/hirosystems/stacks-wallet-web/pull/2303 again

kantai (approved):
This is fine to me, so long as you're all clear on what the goal is here: removing application validation of BNS names. Without this validation, applications can no longer trust that the BNS name that the authentication response provides is actually owned by the claimed user.

markmhendrickson:
Per @kantai's point here, I wonder what documentation may need updating to make this change of behavior clear. And perhaps we should pair this with removing usernames from the auth responses in general, to prevent developers from assuming incorrectly that they've been verified?

@janniks want to audit the docs for such possible updates and make the auth response change as well?

janniks:
I'm thinking a major release bump and changelog mention. But I'll go through the docs to see what should be updated.

> ... and make the auth response change as well?

Yes, good point! I'm removing these now and checking how it ties in to the user session data.

janniks:
98bff3f removes the `username` field completely. since it's just a removal i'm not sure if it warrants a change in the token.version (e.g. [see sip-auth, which isn't merged yet →](https://github.com/friedger/sips/blob/auth/sips/sip-x/sip-x-authentication-protocol.md#upgrade-from-131-to-200))
cc @kantai @zone117x

zone117x:
> [98bff3f](https://github.com/hirosystems/stacks.js/commit/98bff3f0e6acd9ab7a433801293b4d90370660e7) removes the `username` field completely

👍 this seems like a safer approach than just documentation changes. 

It seems like a token version bump would make sense with this change -- it would help apps that may want to explicitly check for this rather than having to infer it. 

Should this change be brought up in https://github.com/stacksgov/sips/pull/50? Ideally, that SIP would make a similar change or make the username optional.

markmhendrickson:
> Should this change be brought up in https://github.com/stacksgov/sips/pull/50? Ideally, that SIP would make a similar change or make the username optional.

Yep we've been in touch with @friedger about this functional change. @janniks I'd suggest leaving a comment about it on the above SIP PR whenever you have a chance as well, to help with visibility.

janniks:
✅ great input! will merge and beta-release to test

Issue #1231, opened 2022-04-15
Add encode/decode message functions for Arbitrary Message Signing


beguene:
For https://github.com/hirosystems/stacks-wallet-web/issues/1051, we need to be able to encode the message with a chain prefix specific 'Stack Message Signing:\n' and both the message and the chain prefix must be prefixed by a variable integer following [bitcoin standard](https://en.bitcoin.it/wiki/Protocol_documentation#Variable_length_integer).
Besides the encode function we need a decode and a hashMessage function.

PR https://github.com/hirosystems/stacks.js/pull/1232

PR #1232, opened 2022-04-15
feat: Add encode / decode messages to support

beguene:
This adds an `encodeMessage`, `decodeMessage` and `hashMessage` functions to support arbitrary message signing
more details in #1231

cc: @janniks 

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/packages/encryption/tests/messageSignature.test.ts
+++ b/packages/encryption/tests/messageSignature.test.ts
@@ -1,0 +1,38 @@
+import { decodeMessage, encodeMessage, hashMessage } from '../src/messageSignature';
+
+test('encodeMessage / decodeMessage', () => {
+  // array of messages and their expected encoded message
+  const messages = [
+    ['hello world', '\x18Stacks Message Signing:\n\x0bhello world'],
+    ['', '\x18Stacks Message Signing:\n\x00'],
+    // Longer message (to test a different length for the var_int prefix)
+    ['This is a really long message to test the var_int prefix This is a really long message to test the var_int prefix This is a really long message to test the var_int prefix This is a really long message to test the var_int prefix This is a really long message to test the var_int prefix',
+      Buffer.concat([
+        Buffer.from('\x18Stacks Message Signing:\n'),
+        // message length = 284 (decimal) = 011c (hex) <=> \x1c\x01 (little endian encoding)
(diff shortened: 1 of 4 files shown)


packages/encryption/tests/messageSignature.test.ts
```diff
@@ -0,0 +1,19 @@
+import { decodeMessage, encodeMessage, hashMessage } from '../src/messageSignature';
+
+test('encodeMessage / decodeMessage', () => {
```
janniks, line 3: can we add a test w/ a manually constructed "correct" encoded message?

packages/encryption/tests/messageSignature.test.ts
```diff
@@ -0,0 +1,19 @@
...
+  expect(decodedMessage1.toString()).toEqual(message1);
+
+  const message2 = '';
+  const encodedMessage2 = encodeMessage(message2);
+  const decodedMessage2 = decodeMessage(encodedMessage2);
+  expect(decodedMessage2.toString()).toEqual(message2);
+});
+
+test('hash message', () => {
+  const message1 = 'This is a new message';
```
janniks: same here, should we compare w/ a real hash?
e.g. `d6d95e464da6763ca98a91cb4087f0e1c7c7870dd94cd8d651fccafe2124bbc1` generated from the first google result for 'sha256 online' 😅

beguene: I used openssl instead of a google result 😄 . Let me know if this is fine.

packages/encryption/tests/messageSignature.test.ts
beguene: @janniks I added 3 encoded messages, let me know if it's explicit enough or if I am missing something.

packages/encryption/tests/messageSignature.test.ts
janniks: Even better 😉

packages/encryption/tests/messageSignature.test.ts
janniks: I think that's fine in combination w/ the openssl test

janniks (approved):

PR #1240, opened 2022-04-25
Support length estimation in fee estimate interface

### LINKED ISSUE #1233: Support length estimation in fee estimate interface
The transaction fee estimation API on the stacks-node supports an `estimated_len` argument to supply a length estimate for the finalized transaction. This is essential for providing accurate fee esimates: many transactions' costs, particularly STX transfers, are dominated by the byte length of the transaction authorization data (signature, fees, etc).

Providing accurate estimates for this requires knowing the number of post-conditions, whether the transaction will be sponsored, and the kind of signer.

[SIP-005](https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md) provides details on the serialization of transactions, but getting accurate estimates will likely require taking a close look at the serialization code in either the stacks-blockchain implementation of transaction serialization and signing or the implementation in stacks.js itself. Essentially, this method will be given a _payload_, and then estimate the length of these transaction fields:

* A 1-byte version number, identifying whether or not the transaction is meant as a mainnet or testnet transaction.
* A 4-byte chain ID, identifying which Stacks chain this transaction is destined for.
* The transaction authorization structure:
      - The address of the origin account.
      - The signature(s) and signature threshold for the origin account.
      - The address of the sponsor account, if this is a sponsored transaction.
      - The signature(s) and signature threshold for the sponsor account, if given.
      - The fee to pay, denominated in microSTX.
* A 1-byte anchor mode
* A 1-byte post-condition mode
* The post condition structure
      - A 4-byte length, indicating the number of post-conditions.
      - The post-conditions

And then add that estimated length to the length of the given payload.

ahsan-javaid:
## Description
This PR fixes the byte length estimation to calculate the `estimated_len ` required for `stacks-node` which is essential for providing accurate fee estimates.

### Context
1) Single-sig transaction byte length remain same due to empty message signature which allocates the space for signature
2) Multi-sig transaction byte length increases by adding signatures
     which causes the incorrect fee estimation because the fee value is set while creating unsigned transaction

Steps to reproduce :
```
import {
  AnchorMode,
  createStacksPrivateKey, makeUnsignedSTXTokenTransfer,
  pubKeyfromPrivKey, publicKeyToString,
  standardPrincipalCV, TransactionSigner,

void (async () => {

  const recipient = standardPrincipalCV('SP3FGQ8Z7JY9BWYZ5WM53E0M9NK7WHJF0691NZ159');
  const amount = 2500000;
  const fee = 100;
  const nonce = 10;
  const memo = 'test memo....';

  const privKeyStrings = [
  ];

  const pubKeys = privKeyStrings.map(pubKeyfromPrivKey);

    recipient,
    amount,
    fee,
    nonce,
    memo: memo,
    numSignatures: 3,
    publicKeys: pubKeyStrings,
    anchorMode: AnchorMode.Any
  });

  const serializedTx = transaction.serialize();
  console.log('unsigned byte length: ', serializedTx.byteLength); // 120
  signer.signOrigin(privKeys[0]);
  console.log('after sig 1: ', signer.transaction.serialize().byteLength); // 186
  signer.signOrigin(privKeys[1]);
  console.log('after sig 2: ', signer.transaction.serialize().byteLength); // 252
  console.log('after sig 3: ', signer.transaction.serialize().byteLength); // 318
})();

```

For details refer to issue #1233

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No
No

## Testing information

1. Test cases included 
2. npm run test passes 

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated
- [X] Tag @janniks and @zone117x for review

### DIFF
--- a/packages/transactions/src/builders.ts
+++ b/packages/transactions/src/builders.ts
@@ -8,6 +8,7 @@ import {
   createSponsoredAuth,
   createStandardAuth,
   SpendingCondition,
+  MultiSigSpendingCondition,
 } from './authorization';
 import { ClarityValue, PrincipalCV } from './clarity';
 import {
@@ -21,6 +22,8 @@ import {
   SingleSigHashMode,
   TransactionVersion,
   TxRejectedReason,
+  RECOVERABLE_ECDSA_SIG_LENGTH_BYTES,
+  StacksMessageType,
 } from './constants';
 import { ClarityAbi, validateContractCall } from './contract-abi';
 import {
@@ -615,7 +618,7 @@ export async function makeUnsignedSTXTokenTransfer(
   );
 
   if (txOptions.fee === undefined || txOptions.fee === null) {
-    const estimatedLen = transaction.serialize().byteLength;
+    const estimatedLen = estimateTransactionByteLength(transaction);
     const txFee = await estimateTransaction(payload, estimatedLen, options.network);
     transaction.setFee(txFee[1].fee);
   }
@@ -841,7 +844,7 @@ export async function makeUnsignedContractDeploy(
   );
 
   if (txOptions.fee === undefined || txOptions.fee === null) {
-    const estimatedLen = transaction.serialize().byteLength;
+    const estimatedLen = estimateTransactionByteLength(transaction);
     const txFee = await estimateTransaction(payload, estimatedLen, options.network);
     transaction.setFee(txFee[1].fee);
   }
@@ -1049,7 +1052,7 @@ export async function makeUnsignedContractCall(
   );
 
   if (txOptions.fee === undefined || txOptions.fee === null) {
-    const estimatedLen = transaction.serialize().byteLength;
+    const estimatedLen = estimateTransactionByteLength(transaction);
     const txFee = await estimateTransaction(payload, estimatedLen, network);
     transaction.setFee(txFee[1].fee);
   }
@@ -1376,7 +1379,7 @@ export async function sponsorTransaction(
       case PayloadType.TokenTransfer:
       case PayloadType.SmartContract:
       case PayloadType.ContractCall:
-        const estimatedLen = options.transaction.serialize().byteLength;
+        const estimatedLen = estimateTransactionByteLength(options.transaction);
         try {
           txFee = (await estimateTransaction(options.transaction.payload, estimatedLen, network))[1]
             .fee;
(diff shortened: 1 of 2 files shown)


packages/transactions/src/builders.ts
```diff
@@ -1424,3 +1426,32 @@ export async function sponsorTransaction(
...
+  // List of Multi-sig transaction hash modes
+  const multiSigHashModes = [AddressHashMode.SerializeP2SH, AddressHashMode.SerializeP2WSH];
+  // Check if its a Multi-sig transaction
+  if (multiSigHashModes.includes(hashMode)) {
+    const multiSigSpendingCondition: MultiSigSpendingCondition = transaction.auth
+      .spendingCondition as MultiSigSpendingCondition;
+    // Estimate signature bytes based on signaturesRequired value
+    // Formula: totalSignatureLength = signaturesRequired * (SIG_LEN_BYTES + 1 byte of type of signature)
+    const totalSignatureLength =
+      multiSigSpendingCondition.signaturesRequired * (RECOVERABLE_ECDSA_SIG_LENGTH_BYTES + 1);
```
zone117x: Great job on the explanatory comments here 🙏 
Could you also link or quote a SIP reference for this particular line?

zone117x: Also, is it possible to call this function with a partially signed multisig `StacksTransaction` object? For example, if 2 out of 3 required signatures are already added, does this function account for that?

ahsan-javaid: > Also, is it possible to call this function with a partially signed multisig `StacksTransaction` object? For example, if 2 out of 3 required signatures are already added, does this function account for that?

This `estimateTransactionByteLength` function is getting called while making a unsigned transaction because fee value is set there. So the purpose of this function is to estimate the transaction byte length when the transaction is in unsigned stage. 
That's why i think checking if 2 out of 3 required signatures are already added is not required.

ahsan-javaid: > Great job on the explanatory comments here 🙏 Could you also link or quote a SIP reference for this particular line?

https://github.com/hirosystems/stacks.js/issues/1233#issue-1205643906

As mentioned by Aaron SIP reference might not have this information but it will require a closer look into `stack-node` implementation. 

I tried to show case the behaviour in test cases as well as in steps to reproduce script. 

🙏 if i am missing anything.

packages/transactions/tests/builder.test.ts
```diff
@@ -1061,6 +1062,89 @@ test('Estimate transaction transfer fee', async () => {
...
+
+  const serializedTx = transaction.serialize();
+  // Unsigned transaction byte length without signatures
+  expect(serializedTx.byteLength).toEqual(120);
+
+  // Now add the required signatures and test it against the value returned by estimateTransactionByteLength
+  const signer = new TransactionSigner(transaction);
+  signer.signOrigin(privKeys[0]);
+  signer.signOrigin(privKeys[1]);
+  signer.signOrigin(privKeys[2]);
```
zone117x, line 1165: Following the convo from the last comment -- if you pass this signed tx to `estimateTransactionByteLength` (with some or all of the required signatures) will it return the correct size?

kantai (commented):
Can you add tests for the single-sig and multi-sig cases that compare the estimate to the completed transaction, something like `assertEqual(finalSerializedTx.length, estimatedLength)`?

packages/transactions/src/builders.ts
zone117x: > This `estimateTransactionByteLength` function is getting called while making a unsigned transaction because fee value is set there. So the purpose of this function is to estimate the transaction byte length when the transaction is in unsigned stage.

Ah I see. Is it still possible for devs to call this with a fully or partially signed multi-sig tx? If so, this function should either A) check that the input is unsigned and throw and error if not, or B) ensure the function works with unsigned, partial signed, and fully signed txs.

janniks: +1  🙏

> [...] throw an error if not, [...]

~furthermore, we can additionally add `Unsigned` to the function name to make it more obvious internally+externally.~ RESOLVED

ahsan-javaid:
Done

packages/transactions/tests/builder.test.ts
```diff
@@ -1061,6 +1062,118 @@ test('Estimate transaction transfer fee', async () => {
...
+  // Create a unsigned single-sig transaction
+  const unsignedTransaction = await makeUnsignedSTXTokenTransfer({
+    recipient,
+    amount,
+    fee,
+    nonce,
+    memo: memo,
+    publicKey: publicKey,
+    anchorMode: AnchorMode.Any
+  });
```
kantai, line 1090: Does this test pass if nonce and fee are set to `null` or `undefined`? Looking at what should be the typical usage in the builders module, the only case in which a caller would need to call the estimateFee method would be when fee is undefined.

kantai: Actually, I'm not sure this test will be sufficient to cover the usage for estimation: the `estimateTransaction` method is invoked within `makeUnsignedSTXTokenTransfer`, meaning that the returned transaction from this line will already have the "final" fee amount. However, if the transaction length changes when the fee goes from `null` to the final fee amount (which I believe it does), that means that the passed estimate would be wrong.

ahsan-javaid: Yes, the tests will pass.

Give it a shot:
https://github.com/hirosystems/stacks.js/blob/5be97a259eec51e582644d09f28f9b790e6621cd/packages/transactions/src/builders.ts#L553-L563

When making unsigned transaction the space for fee and nonce is allocated based on data type of fee and nonce which is IntegerType no matter what value one can set or change the fee later. 

That means transaction byte length will remains the same because it's space is allocated based on data type. No matter what value you store in it or you change anything later.

🙏

kantai: Okay, that's fine then, thanks

kantai (approved):

janniks (approved):
LGTM

zone117x (approved):

PR #1246, opened 2022-05-05
Require referrer header to be set

### LINKED ISSUE #1244: Require referrer header to be set
It's difficult to identify web requests generated from the stacks.js lib.
The `referrer`/`referer` header is a widely accepted header in many third-party services (like Cloudflare) for identifying where traffic is coming from.

The stacks.js lib should require users set this header when using functions whihc result in a web request being sent for better attribution.

https://github.com/hirosystems/stacks.js/blob/fe831cf28d71de9410c81cfabe256fd511e9aa4d/packages/common/src/fetchUtil.ts#L6-L7
https://github.com/hirosystems/stacks.js/blob/b5e4e72b4532130005636f68425a4b93f23994fe/packages/keychain/src/utils/gaia.ts#L125-L126
https://github.com/hirosystems/stacks.js/blob/c729006b8d37a18005adff146aea5f0f9aaf6d5d/packages/transactions/src/utils.ts#L204-L205

### LINKED ISSUE #1248: [test][wallet-sdk] fix wallet restore test
### background
the test [`'restore wallet with username owned by stx private key'`](https://github.com/hirosystems/stacks.js/blob/49a094b38aa9287cf4287626e5fa774138093b52/packages/wallet-sdk/tests/models/wallet.test.ts#L37) is currently incorrectly mocked and therefore results in failing tests on upcoming PRs. `fetchLegacyWalletConfig` was using the incorrect/different `fetchPrivate` and escaped mocking.

### scope
add more+better tests for `restoreWalletAccounts`

for example:
- test where only legacy config exists
- test where only current-config exists
- test where both exist
- etc.

if it helps / is necessary, we can change [this code](https://github.com/hirosystems/stacks.js/blob/49a094b38aa9287cf4287626e5fa774138093b52/packages/wallet-sdk/src/models/wallet.ts#L42-L45) to "run in band" (deterministic order)

ahsan-javaid:
## Description

 - Allow referrer header to be set by default
 - Provide a way to change default referrer as well as other options when fetch is used internally by stacks.js libraries

For details refer to issue #1244

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No

No

## Testing information

1. `npm run test`

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated
- [X] Tag @janniks and @zone117x for review


janniks:
We have some ideas on this change in connection to API tokens/keys. I'll discuss with the team and come back to this PR.

packages/common/src/fetchUtil.ts
```diff
@@ -1,12 +1,33 @@
 import 'cross-fetch/polyfill';
 
+// Define a default request options and allow modification
+// Reference: https://developer.mozilla.org/en-US/docs/Web/API/Request/Request
+const defaultFetchOpts: RequestInit = {
+  referrer: 'about:client', // This is by default value: See above reference
```
zone117x: We can just omit this property `referrer` now right?

zone117x (approved):
LGTM!
Also @janniks had a really good idea of back-porting this change to the last 1 or 2 major versions as well, if possible, it would help speed up adoption.

janniks (approved):
Looks good!

PR #1247, opened 2022-05-06
[POC] Support specifying an API key in network requests

zone117x:
This is a proof-of-concept implementation of adding a `fetch` configuration option to the existing `network` object used by almost all functions in stacks.js to perform http requests. This is also related to issue https://github.com/hirosystems/stacks.js/issues/994 -- while this PR doesn't remove the implicit fetching, it at least makes it configurable.

The property is a simple interface matching the standardized `fetch` API:
```ts
type FetchFn = (url: string, init?: RequestInit) => Promise<Response>;
```

Additionally, two proof-of-concept implementations of API key middleware are also implemented. The first one, `createApiKeyMiddleware`, is a minimalistic function that can be used to specify an API key string. Example usage:
```ts
import { createApiKeyMiddleware, getDefaultFetchFn } from '@stacks/common';
import { StacksMainnet } from '@stacks/network';
import { getNonce } from '@stacks/transactions';

const apiKey = '1234-my-api-key-example';

// Create an instance of the API key middleware
const apiKeyMiddleware = createApiKeyMiddleware({ apiKey });

// Create fetch instance using the default `fetch` lib provided by `cross-fetch`,
// also providing the middleware
const fetchFn = getDefaultFetchFn(apiKeyMiddleware);

// Create a network instance using the configured fetch function
const network = new StacksMainnet({ fetchFn });

// Provide the network instance to a stacks.js function
const fetchNonce = await getNonce('SP3FGQ8Z7JY9BWYZ5WM53E0M9NK7WHJF0691NZ159', network);

// The http request made by the above function will then include the http header:
//     `x-api-key: 1234-my-api-key-example`
```

A second included middleware implements a proof-of-concept version of dynamic per-session API key authentication. It's located in `fetchApiAuthMiddleware.ts`. This one uses the middleware `post` callback to check for `401 Unauthorized` responses, then requests a new API key, and retries the http request. The example usage is similar:
```ts
const authMiddleware = createApiSessionAuthMiddleware({
  // Optionally specify a storage interface to persist auth keys (defaults to in-memory)
  authDataStore: ...,
  // Optionally specify the endpoint path used to request API keys
  authPath: '/request_key',
  // Optionally include metadata in the auth request (free form object)
  authRequestMetadata: {
    productName: 'hiro-explorer',
    productVersion: 'v1.2.3'
  },
});

const network = new StacksMainnet({ fetchFn: getDefaultFetchFn(authMiddleware) });
const fetchNonce = await getNonce('SP3FGQ8Z7JY9BWYZ5WM53E0M9NK7WHJF0691NZ159', network);
```

Note that `createApiSessionAuthMiddleware` is an untested proof-of-concept, and the API backend is not yet ready anyway (as of creating this PR). It was implemented as a test to see if the fetch middleware approach is usable in the future for dynamic key authentication.


kyranjamie (commented):
Exceptional work, Matt 🎉    I think this is all we need for the wallets.

https://github.com/hirosystems/stacks.js/issues/994 would've been too much work up front so def the right call to add configuration before thinking about that.

 👏🏼     for ` interface SessionAuthDataStore{}`. Regarding the situation of having a client app read from the stacks.js-created session key for non-stacks.js requests, maybe it'd also be helpful to have baked-in callbacks in the middleware? `onTokenCreated`, `onTokenExpirary`? 🤔   Not entirely sure if this is necessary, was thinking that we might need to call a ReactQuery method or something in some cases, but actually if each request reads from the store, that would also be fine.

packages/common/src/fetchUtil.ts
```diff
@@ -67,46 +67,51 @@ function createDefaultMiddleware(): FetchMiddleware[] {
...
-        fetchParams = result ?? fetchParams;
-      }
+      if (typeof middleware.pre !== 'function') continue;
+      const result = await Promise.resolve(
+        middleware.pre({
+          fetch: fetchLib,
+          ...fetchParams,
+        })
+      );
+      fetchParams = result ?? fetchParams;
```
zone117x, line 99: Nit: my code style preference would be something like
```js
if (typeof middleware.pre === 'function') {
   const result = await Promise.resolve(...);
}
```
Makes things a bit more readable by avoiding the coding equivalent of a "double negative", and avoid reasoning about early returns/continues.

zone117x:
~~These changes look good! Note that the `createApiSessionAuthMiddleware` function (should all be isolated to the `fetchApiAuthMiddleware.ts` file) is preliminary/demo/POC code that won't have a final spec or backend implementation ready soon enough.~~

~~The `createApiSessionAuthMiddleware` function, however, is essentially spec-finalized and should have back-end support soon (as in, we could reasonably do a pre-release with stacks.js including it, and merge this PR sooner rather than later).~~

~~I'd advocate for either A) branching off this, deleting `fetchApiAuthMiddleware.ts`, and opening a new PR, or B) deleting `fetchApiAuthMiddleware.ts` in this branch and just ensuring it's in git history for later when we introduce it.~~

EDIT: nvm, looks like you've already taken care of this in https://github.com/hirosystems/stacks.js/pull/1255 :)

packages/common/src/fetchUtil.ts
janniks: You're probably right in this case — `continue` and negation is a bit much. Although, in general, I'm a big fan of early-returns/guards to reduce the complexity of following different nests/flows.

janniks:
split into:
- https://github.com/hirosystems/stacks.js/pull/1255
- https://github.com/hirosystems/stacks.js/pull/1261

PR #1260, opened 2022-05-19
fix(message-signing): add working message verification utility fn

kyranjamie:
In changing the signature format, the current docs & blog post has broken code examples.

See https://github.com/hirosystems/stacks-wallet-web/issues/2435

I copied `parseRecoverableSignature` from `@stacks/transactions`. What's best solution to reuse the fn? Atm encryption doesn't rely on tx package. Should we move to common?

Open to other solutions. Would be good to come up with a quick fix though, so users don't hit a wall when trying to use msg signing.

### DIFF
--- a/packages/encryption/src/ec.ts
+++ b/packages/encryption/src/ec.ts
@@ -1,6 +1,14 @@
 import { hmac } from '@noble/hashes/hmac';
 import { sha256 } from '@noble/hashes/sha256';
-import { getPublicKey, getSharedSecret, Point, signSync, utils, verify } from '@noble/secp256k1';
+import {
+  Signature,
+  getPublicKey,
+  getSharedSecret,
+  Point,
+  signSync,
+  utils,
+  verify,
+} from '@noble/secp256k1';
 import {
   Buffer,
   toBuffer,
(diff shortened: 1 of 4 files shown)


packages/encryption/src/ec.ts
```diff
@@ -502,3 +512,37 @@ export function verifyECDSA(
   const contentHash = hashSha256Sync(contentBuffer);
   return verify(signature, contentHash, publicKey);
 }
+
+export function parseRecoverableSignature(signature: string) {
+  const coordinateValueBytes = 32;
+  if (signature.length < coordinateValueBytes * 2 * 2 + 1) {
+    throw new Error('Invalid signature');
+  }
+  const recoveryParamHex = signature.substr(0, 2);
```
janniks: Maybe switch to str.slice as substr is somewhat deprecated

packages/encryption/src/ec.ts
```diff
@@ -502,3 +512,37 @@ export function verifyECDSA(
   const contentHash = hashSha256Sync(contentBuffer);
   return verify(signature, contentHash, publicKey);
 }
+
+export function parseRecoverableSignature(signature: string) {
```
janniks: Seems fine to move to @stacks/ common, we can start a new signature.ts file.
For the long-term we can think about adding another smaller signature specific package which tx+enc can rely on.

janniks (approved):
LGTM

PR #1269, opened 2022-05-23
[poc] add missing rsv functions

janniks:
## scope
- adds previously missing signing functions in RSV format (Clarity compatible)

## references (important for history/context)
- https://github.com/hirosystems/stacks-wallet-web/issues/2419
- https://github.com/hirosystems/stacks-wallet-web/pull/2429
- https://github.com/hirosystems/stacks.js/pull/1263

## usage
The original functions are marked as deprecated and the correct format methods introduced use a suffix of `Rsv` e.g. `verifyMessageSignatureRsv(...)`

I would propose this _deprecating_ path forward and let stacks-wallet-web switch to `signMessageHashRsv`. The suffix highlights explicitness. I also added `MessageHash` to the function name to further make it clear that a typical use works on a hash (compared to a message/plain-text) and the sign function does not do any previous message hashing here.

If folks are against the suffix, I'm also not opposed to `signMessageHash` or a consumer-facing `signMessage`, which would internally use more explicit methods.

In a planned overhaul (and major version bump) of the available crypto-helper methods of stacks.js, we will probably refactor this again and remove deprecated functions. Ideally, I would want to see stacks.js only export Clarity-compatible methods (to not confuse developers).

### DIFF
--- a/packages/common/src/signatures.ts
+++ b/packages/common/src/signatures.ts
@@ -1,16 +1,28 @@
 import { hexToInt } from './utils';
 
-export function parseRecoverableSignature(signature: string) {
-  const coordinateValueBytes = 32;
-  if (signature.length < coordinateValueBytes * 2 * 2 + 1) {
+const COORDINATE_BYTES = 32;
+
+/** @ignore */
+export function parseRecoverableSignatureVrs(signature: string) {
+  if (signature.length < COORDINATE_BYTES * 2 * 2 + 1) {
     throw new Error('Invalid signature');
   }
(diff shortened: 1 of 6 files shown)


janniks:
@pradel does this make sense?

zone117x (approved):
Always a fan of explicit naming 👍

kyranjamie (approved):

beguene (approved):
Really like the Rsv suffix, it will avoid confusion and prevent errors.

PR #1271, opened 2022-05-23
fix: reduce fallbacks/polyfills

### LINKED ISSUE #1259: [stacking] Offload `bitcoinjs-lib`
replace bitcoinjs-lib in @stacks/stacking with secp/noble alternatives as in previous PRs

janniks:
- closes https://github.com/hirosystems/stacks.js/issues/1259
- reduces unnecessary polyfilling


packages/encryption/src/messageSignature.ts
```diff
@@ -1,13 +1,13 @@
-import { sha256 } from 'sha.js';
-import { encode, decode, encodingLength } from 'varuint-bitcoin';
+import { sha256 } from '@noble/hashes/sha256';
```
zone117x, line 1: Is the `sha.js` dep still used? It's still in package.json

packages/transactions/src/utils.ts
```diff
@@ -66,7 +68,8 @@ export function omit<T, K extends keyof any>(obj: T, prop: K): Omit<T, K> {
   return clone;
 }
 
-export class sha512_256 extends sha512 {
+/** @deprecated use @noble/hashes instead */
+export class sha512_256 extends shaJsSha512 {
```
zone117x: Any compelling reason to deprecate this instead of just removing it?

packages/encryption/src/messageSignature.ts
```diff
@@ -1,13 +1,13 @@
...
 import { Buffer } from '@stacks/common';
+import { decode, encode, encodingLength } from 'varuint-bitcoin';
```
zone117x, line 3: Note: `varuint-bitcoin` is a good candidate for vendoring later, it has an unfriendly Nodejs buffer polyfill dependency.

zone117x (approved):

packages/encryption/src/messageSignature.ts
janniks: Nope, was intended to be removed — I missed it, thanks 🙏

packages/transactions/src/utils.ts
janniks: I'd be happy to remove and treat it as an internal (not triggering breaking change). Otherwise if we want to treat as external and used by consumers, I would keep this change to a minor/patch and remove the deprecated with the next major release.

But honestly, it's unlikely folks used this undocumented helper, even if it was `export`ed, so I'd also prefer treating this (and similar methods) as internal.

PR #1279, opened 2022-05-31
[FIX] verify in verifyECDSA is missing strict param

### LINKED ISSUE #1278: verify in verifyECDSA is missing strict param
Hi there --

I have many tests in micro-stacks that make use of stack.js methods to ensure compatibility between the two libraries, and as redundancies against regressions. I recently upgraded many of the stacks.js libs to their latest, and with stacks.js updating to use some of the noble dependencies like micro-stacks uses, the settings for verifyECDSA need to be updated in `@stacks/encryption`.

Here is the implementation in micro-stacks:
https://github.com/fungible-systems/micro-stacks/blob/main/src/crypto/encryption/sign.ts#L86-L89

There was a change in the underlying noble library: https://github.com/paulmillr/noble-secp256k1/releases/tag/1.4.0

> Important: verify() is now strict: true by default. High-s signatures are rejected, which also mirrors libsecp behavior.

which is different than legacy stacks implementations.

Here is the line that needs to change:

https://github.com/hirosystems/stacks.js/blob/master/packages/encryption/src/ec.ts#L515

my tests would only fail randomly, https://github.com/fungible-systems/micro-stacks/blob/main/src/crypto/encryption/encrypt-content.test.ts#L140-L163 this test specifically. when I changed the strict setting to false by default, it never failed again.

ahsan-javaid:
## Description
- By default `@noble/secp256k1` signature verification rejects High-s signatures.
- Support stacks legacy implementation like `verify(sig, hashedMsg, publicKey, { strict: false });` by setting strict: false
- Reference: `https://github.com/paulmillr/noble-secp256k1/releases/tag/1.4.0`

For details refer to issue #1278

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No

No

## Testing information

`npm run test`
Existing tests are robust enough so no new test is added

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated
- [X] Tag reviewers

### DIFF
--- a/packages/encryption/src/ec.ts
+++ b/packages/encryption/src/ec.ts
@@ -512,7 +512,10 @@ export function verifyECDSA(
 ): boolean {
   const contentBuffer = getBuffer(content);
   const contentHash = hashSha256Sync(contentBuffer);
-  return verify(signature, contentHash, publicKey);
+  // verify() is strict: true by default. High-s signatures are rejected, which mirrors libsecp behavior
+  // Set verify options to strict: false, to support the legacy stacks implementations
+  // Reference: https://github.com/paulmillr/noble-secp256k1/releases/tag/1.4.0
+  return verify(signature, contentHash, publicKey, { strict: false });
 }
 
 interface VerifyMessageSignatureArgs {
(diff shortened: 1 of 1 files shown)


kantai (approved):
LGTM -- thanks for patching this!

saralab:
@zone117x  Can you review this please? Waiting for your approval to merge this.

janniks:
Also @kantai & @zone117x do you have any resources I could read up on regarding high-s signatures, `strict`, etc. and what they mean? I didn't find much googling around.

zone117x (approved):

Issue #1281, opened 2022-06-01
add sip018 tests


janniks:
add tests with test vectors from sip018 https://github.com/MarvinJanssen/sips/blob/feat/signed-structured-data/sips/sip-018/sip-018-signed-structured-data.md

beguene:
@janniks FYI I am working on updating our wallet with SIP018 https://github.com/hirosystems/stacks-wallet-web/issues/2460. As part of this I am currently adding some methods to stacks.js to handle encoding/decoding + signingStructuredData functions to stacks.js
I will put up a PR soon I can also add those tests vectors if you did not work on it yet.

janniks:
perfect, thanks 🙏

PR #1284, opened 2022-06-02
feat: add SIP-018 support

### LINKED ISSUE #1283: Add new methods to support SIP018
See https://github.com/hirosystems/stacks-wallet-web/issues/2460.

We need a `signStructuredData`, `decodeStructuredDataSignature`, `encodeStructuredData` and `hashStructuredData`
that implement SIP018
https://github.com/MarvinJanssen/sips/blob/feat/signed-structured-data/sips/sip-018/sip-018-signed-structured-data.md

beguene:
closes #1283 #1281

Those functions are required to support SIP-018 here, in connect and the web-wallet
The following functions have been added
`encodeStructuredData`
`decodeStructuredDataSignature`
`signStructuredData`

## Testing information

I added some unit test and the test vectors specified in[ SIP-018](https://github.com/MarvinJanssen/sips/blob/feat/signed-structured-data/sips/sip-018/sip-018-signed-structured-data.md#test-vectors) covering 100% of the new code paths.

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

cc: @janniks @kyranjamie @MarvinJanssen @aulneau


janniks:
thoughts on moving the new files/methods to the @stacks/encryption package? currently, this is where all the additional signing related methods are (which are not necessarily needed for transactions) — this way we could keep the encrpytion dep out of transactions deps.

cc @kyranjamie 

> _in the future (maybe next major release), i'd like to move to a separate package that bundles all signing + hashing (and other crypto stuff, with the exception of encryption). this package would then be used for most things and would become a dep of both tx and encryption._

beguene:
@janniks Yes I initially did that but I needed `signMessageHashRsv` which is in `stacks/transactions`, and I did not want to include `stacks/transaction` as a dependency in `stacks/encryption`. 😞 
Indeed all the signature functions should be out of `stacks/transactions`. Any thoughts on the best way to organise this ?
cc @kyranjamie

beguene:
I could temporarily move my `signStructuredData` into the web-wallet and just have the core primitive and use
`signMessageHashRsv` in the wallet itself, until we have this new split of packages in stacks.js

beguene:
or I can just keep `signStructuredData` in `@stacks/transactions` and move everything else in `@stacks/encryption`

kyranjamie:
Not sure on best way to organise this, but signing methods in encryption package doesn't really make sense.

Either we have a generic `@stacks/crypto` package, or we make a new one for general message signing? 🤔

janniks:
🙃 semantically signing is somewhat closer to tx, as txs need signing _(not structured)_.

in hindsight, signing can only really live in tx unless we want to split structured-signing from signing _(even in the monorepo i don't want the package count to explode)_. so if structured signing needs clarity values (currently in tx, unless we wanna open the next can of worms), and signing anything else would result in circular deps. 

so for now, i think we can leave as is — and i'll try to think of how/if this can be improved

packages/transactions/tests/structuredDataSignature.test.ts
```diff
@@ -0,0 +1,190 @@
...
+          name: stringAsciiCV('Test App'),
+          version: stringAsciiCV('1.0.0'),
+          'chain-id': uintCV(1),
+        }),
+        '2538b5dc06c5ae2f11549261d7ae174d9f77a55a92b00f330884695497be5065',
+      ],
+    ];
+
+    for (const [input, expectedHash] of inputs) {
+      expect(hashStructuredData(input).toString('hex')).toEqual(expectedHash);
```
janniks: jest also provides [parameterized testing](https://jestjs.io/docs/api#testeachtablename-fn-timeout) for these `for` tests.
_(also relevant for test `'encodeStructuredData / decodeStructuredDataSignature'`)_

janniks (approved):
Looks good!

At some point we might need to think about how this type will be serialized. Currently, devs could probably cast it to a clarity buffer and use the [reference clarity verifier](https://github.com/stacksgov/sips/blob/48a8493f19702a14258db5730fece837cd783b7a/sips/sip-018/sip-018-signed-structured-data.md#clarity-implementation)

beguene:
@janniks I added `test.each` and made a few more changes
- I changed the function signature for `encodeStructuredData` it uses now 'named parameters' to avoid order mistakes
```javascript
export function encodeStructuredData({
  message,
  domain,
}: {
  message: ClarityValue;
  domain: ClarityValue;
}): Buffer {
 ```
- I added some more failure test cases
- I enforced one SIP-018 rule regarding the domain tupleCV with a `isDomain` function
https://github.com/MarvinJanssen/sips/blob/feat/signed-structured-data/sips/sip-018/sip-018-signed-structured-data.md#formal-specification

cc @aulneau @MarvinJanssen

**Review by MarvinJanssen — APPROVED** [review]
LGTM! Excited to try this out soon.

File: packages/transactions/tests/structuredDataSignature.test.ts [hunk 2]
```diff
@@ -0,0 +1,242 @@
...
+  decodeStructuredDataSignature,
+  encodeStructuredData,
+  hashStructuredData,
+  signStructuredData,
+  STRUCTURED_DATA_PREFIX,
+} from '../src/structuredDataSignature';
+
+const chainIds = {
+  mainnet: 1,
+  testnet: 2147483648,
```
janniks, line 15: minor: `testnet` is never used, but we can also keep around if we ever want tests that use it 🤷‍♂

janniks (approved):
Looks good! Should we merge+release?

packages/transactions/tests/structuredDataSignature.test.ts
beguene: Up to you but I think we can keep it.

beguene:
> Looks good! Should we merge+release?

Yes ! 🚀

PR #1306, opened 2022-06-30
fix: update version buffer size for pox addresses

janniks:
bug reported on [discord#documentation](https://discord.com/channels/621759717756370964/750467085720617031/992103351334469685)

using the example stacking function with this version fails with the following (pre-4.0.0 version works as expected):
```
 ➭ cat package.json && node index.js 
{
  "name": "t",
  "version": "1.0.0",
  "description": "",
  "keywords": [],
  "author": "",
  "license": "GPLv3",
  "type": "module",
  "dependencies": {
    "@stacks/network": "^4.3.0",
    "@stacks/stacking": "^4.3.0"

  }
}
(node:66903) ExperimentalWarning: The Fetch API is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
/Users/wileyj/Downloads/test/node_modules/@stacks/transactions/dist/contract-abi.js:265
                throw new Error(`Clarity function \`${payload.functionName.content}\` expects argument ${argNum} to be of type ${getTypeString(abiArg.type)}, not ${(0, clarity_1.getCVTypeString)(payloadArg)}`);
                      ^

Error: Clarity function `stack-stx` expects argument 2 to be of type (tuple (hashbytes (buff 20)) (version (buff 1))), not (tuple (hashbytes (buff 20)) (version (buff 16)))
    at validateContractCall (/Users/wileyj/Downloads/test/node_modules/@stacks/transactions/dist/contract-abi.js:265:23)
    at makeUnsignedContractCall (/Users/wileyj/Downloads/test/node_modules/@stacks/transactions/dist/builders.js:350:49)
    at async makeContractCall (/Users/wileyj/Downloads/test/node_modules/@stacks/transactions/dist/builders.js:395:29)
    at async StackingClient.stack (/Users/wileyj/Downloads/test/node_modules/@stacks/stacking/dist/index.js:145:20)
    at async file:///Users/wileyj/Downloads/test/index.js:14:25

Node.js v18.4.0
```
---

The bug was introduced when switching away from bn.js. Buffers from BigInt were the wrong size after the switch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -81,7 +81,7 @@
 			"devDependencies": {
 				"@changesets/cli": "^2.23.0",
 				"@janniks/typedoc-plugin-monorepo": "^0.4.2-1",
-				"@janniks/typedoc-theme-stacks": "^1.0.1",
+				"@janniks/typedoc-theme-stacks": "^1.0.3",
 				"@rollup/plugin-alias": "^3.1.9",
 				"@rollup/plugin-babel": "^5.3.0",
 				"@rollup/plugin-commonjs": "^22.0.0",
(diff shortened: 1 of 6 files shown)


kantai (approved):
This LGTM -- my only comment is on testing this: is it possible to include the PoX contract's ABI in the unit tests and check that the StackingClient behaves correctly by checking the ABI? (looking at the discord bug, it seems like that's how this bug was caught)

janniks:
> This LGTM -- my only comment is on testing this: is it possible to include the PoX contract's ABI in the unit tests and check that the StackingClient behaves correctly by checking the ABI? (looking at the discord bug, it seems like that's how this bug was caught)

Yes, good point — I did this during testing but removed again because mocking gets very tedious. But I agree — without it this type of bug is not covered by the tests. I'll explore some more...

janniks:
@kantai Added PoX Abi validation in the test suite. A bit round-about, but it's a simple fix and works well with the existing mocks.

kantai (approved):
LGTM!

PR #1320, opened 2022-07-14
fix: update make_keychain command

### LINKED ISSUE #1303: [CLI] `stx make_keychain` broken in node 17, 18
When using testing node versions `v17.9.1` and `v18.4.0` I get the following, in the latest version of the CLI. Noticed this as I changed default node version and have [a hotkey to generate testnet addresses](https://github.com/kyranjamie/dot-files/blob/master/.hammerspoon/init.lua#L48-L51).

```
❯ stx make_keychain -t
Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:67:19)
    at createHash (node:crypto:133:10)
    at Object.hash160 (/Users/xxx/.config/yarn/global/node_modules/bip32/src/crypto.js:15:16)
    at BIP32.get identifier [as identifier] (/Users/xxx/.config/yarn/global/node_modules/bip32/src/bip32.js:65:23)
    at BIP32.get fingerprint [as fingerprint] (/Users/xxx/.config/yarn/global/node_modules/bip32/src/bip32.js:68:21)
    at BIP32.derive (/Users/xxx/.config/yarn/global/node_modules/bip32/src/bip32.js:151:88)
    at BIP32.deriveHardened (/Users/xxx/.config/yarn/global/node_modules/bip32/src/bip32.js:168:21)
    at /Users/xxx/.config/yarn/global/node_modules/bip32/src/bip32.js:182:31
    at Array.reduce (<anonymous>)
    at BIP32.derivePath (/Users/xxx/.config/yarn/global/node_modules/bip32/src/bip32.js:178:26)
error:0308010C:digital envelope routines::unsupported
```

janniks:
- migrate to scure-bip32 where possible
- disable node `hashRipemd160` (broken in current stable node, requires legacy openssl flags)
- fix faulty tests (invalid mnemonic used prior)

### DIFF
--- a/packages/cli/src/common.ts
+++ b/packages/cli/src/common.ts
@@ -1,8 +1,9 @@
+import { publicKeyToAddress } from '@stacks/encryption';
+import { pubKeyfromPrivKey } from '@stacks/transactions';
+import * as bitcoinjs from 'bitcoinjs-lib';
+import { TransactionSigner } from 'blockstack';
 import { DEFAULT_MAX_ID_SEARCH_INDEX } from './argparse';
 import { CLINetworkAdapter } from './network';
-import * as blockstack from 'blockstack';
-import { TransactionSigner } from 'blockstack';
-import * as bitcoinjs from 'bitcoinjs-lib';
 
 let maxIDSearchIndex = DEFAULT_MAX_ID_SEARCH_INDEX;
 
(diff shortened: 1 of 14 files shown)


packages/encryption/src/hashRipemd160.ts
```diff
@@ -47,17 +46,17 @@ export class NodeCryptoRipemd160Digest implements Ripemd160Digest {
...
-      return nodeCrypto.createHash;
-    }
-    return false;
-  });
-  if (nodeCryptoCreateHash) {
-    return new NodeCryptoRipemd160Digest(nodeCryptoCreateHash);
-  } else {
-    return new Ripemd160PolyfillDigest();
-  }
+  // // disable node hashRipemd160, because it doesn't work with node 17,18 openSSL
```
janniks, line 49: _Commented out the node version of this hashing function._ Maybe we should switch to js version everywhere. Would simplify code a bunch. The only thing against it would be potential loss of performance, where node might switch to native openSSL.

packages/cli/src/common.ts
```diff
@@ -46,13 +47,12 @@ export function getPrivateKeyAddress(
...
-  } else {
-    const pk = privateKey;
-    const ecKeyPair = blockstack.hexStringToECPair(pk);
-    return network.coerceAddress(blockstack.ecPairToAddress(ecKeyPair));
+    return privateKey.address;
   }
+
+  const pubKey = pubKeyfromPrivKey(privateKey);
+  const btcAddress = publicKeyToAddress(pubKey.data);
+  return network.coerceAddress(btcAddress);
```
kyranjamie, line 55: 🧘🏼‍♂️  early return > else

kyranjamie (approved):

PR #1329, opened 2022-07-26
fix(encryption): change message signing prefix

### LINKED ISSUE #1328: Incorrect prefix for message signing
We use the wrong phrasing for the message signing prefix:

https://github.com/hirosystems/stacks.js/blob/5ac9385069ce48bde452520d5168c84e31b5b961/packages/encryption/src/messageSignature.ts#L5-L7

Following the [Ethereum format](https://eips.ethereum.org/EIPS/eip-712#signatures-and-hashing-overview), already followed on the Ledger app, the message should be _Stacks Signed Message_, not _Stacks Message Signing_.

This is a change of length, too, so if I'm not mistaken about how the prefixing works, should be:

```
\x17Stacks Signed Message:\n
```

`"Stacks Signed Message:\n".length.toString(16) // 17`

Useful info on prefix here https://blog.ricmoo.com/verifying-messages-in-solidity-50a94f82b2ca

kyranjamie:
> This PR was published to npm with the version `4.3.9-pr.bb49829.0`
> e.g. `npm install @stacks/common@4.3.9-pr.bb49829.0 --save-exact`

Closes #1328 

This is a potentially breaking change for message signing, but brings the stacks implementation of message signing inline with the eth precedent, and stacks ledger app implementation

cc/ @aulneau for microstacks change?

### DIFF
--- a/.github/MIGRATION.md
+++ b/.github/MIGRATION.md
@@ -3,6 +3,7 @@
 - [Stacks.js (&lt;=4.x.x) → (5.x.x)](#stacksjs-4xx--5xx)
   - [Breaking Changes](#breaking-changes)
     - [Buffer to Uint8Array](#buffer-to-uint8array)
+    - [Message Signing Prefix](#message-signing-prefix)
 - [blockstack.js → Stacks.js (1.x.x)](#blockstackjs--stacksjs-1xx)
   - [Auth](#auth)
   - [Using blockstack.js](#using-blockstackjs)
@@ -18,7 +19,8 @@
 
 ### Breaking Changes
 
-- To reduce the bundle sizes of applications using Stacks.js we are switching from Buffer (a polyfill to match Node.js APIs) to Uint8Arrays (which Buffers use in the background anyway). [Read more...](#buffer-to-uint8array)
+- To reduce the bundle sizes of applications using Stacks.js, we are switching from Buffer (a polyfill to match Node.js APIs) to Uint8Arrays (which Buffers use in the background anyway). [Read more...](#buffer-to-uint8array)
+- To allow message signing on Ledger hardware wallets, we are changing the message signing prefix. [Read more...]
 
 #### Buffer to Uint8Array
 
@@ -54,6 +56,14 @@ hexToBytes('deadbeef'); // Uint8Array(4) [ 222, 173, 190, 239 ]
 Buffer.from([222, 173, 190, 239]).toString('hex'); // 'deadbeef'
 bytesToHex(Uint8Array.from([222, 173, 190, 239])); // 'deadbeef'
 ```
+
+#### Message Signing Prefix
+
+The message signing prefix was changed from `Stacks Message Signing` to `Stacks Signed Message`.
+The change relates to the functions `verifyMessageSignature`, `encodeMessage`, `decodeMessage`, and `hashMessage`.
+The `verifyMessageSignature` functions was updated to verify against both the old and the new prefix (for unhashed message-input).
+This will generate a different hash/signature from the same input compared to previous versions of Stacks.js.
+If you have previously stored messages/signatures and compare to freshly generated ones, the messages/signatures will not match to previously stored.
 
 ---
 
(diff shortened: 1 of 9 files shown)


beguene (approved):
Good catch ! Code looks good, prefix calculation looks good too.

aulneau:
@kyranjamie i'm curious about this change, it would diverge from [SIP-018](https://github.com/stacksgov/sips/pull/57) have you brought this up there yet?

Also, what happens with any apps using this in production?

Can you say more about the rational of why this is important?

Edit: I now realize that there is nothing in the SIP associated with this lol -- but I do worry that there might be some breakage of apps with this change, what do you think?

kyranjamie:
I think it's important we're consistent and follow existing patterns, unless there's a good reason not to. (What's the rationale for deviating?) 

Using the noun, _Signed Message_ means the prefix describes the content, rather than the verb "Message Signing", which is the action required to get the result.

There may be some breakages, but I'm thinking it's better to adjust this inline with our Ledger app now while the features relatively new, than leaving as is.

cc/ @friedger @pradel for input

fbwoolf (approved):

kyranjamie:
The Latest ledger app, [`0.22.9`](https://github.com/Zondax/ledger-blockstack/releases/tag/v0.22.9), requires all payloads to begin with this prefix, rejecting those that use anything else.

janniks (approved):

kyranjamie:
@janniks can you help get this one merged? Do you reckon this should be a `BREAKING CHANGE`?

janniks:
How urgent is the timeline for this?
I would love to bundle it with some other breaking changes.

kyranjamie:
> How urgent is the timeline for this? 

Not super urgent, waiting on Ledger release for this to be absolutely necessary

janniks:
Any timeline on the ledger release?

kyranjamie:
They said "around two weeks" on the 29th August

janniks:
~When ready let's squash merge this into `next` with a `BREAKING CHANGE` commit body~

janniks:
Should we think about somehow coordinating libs and wallets to launch this at the same time? The problem is that backend/frontends will need to match the `prefix` which the wallet is using to correctly verify signatures.
I believe, this will break login for: Gamma, Sigle, Console

aulneau:
I was thinking so, it might also be of use to keep the legacy one around for a short time so both can be used as apps migrate away from it

janniks:
The only thing I can think of is letting the `verifyMessageSignature` run twice in the background and return `true` if any of the two `prefixes` can be verified. Will also only work if the `message` is passed-in directly, rather than already encoded+hashed.
_Feels a bit incorrect, but safe-enough for potentially stopping a breakage in live sign-in systems_.

**EDIT**: prototype https://github.com/hirosystems/stacks.js/pull/1356

kyranjamie:
Should we create a new method entirely `backwardsCompatibleVerifyMessageSignature` and release it with a lead on the wallet changes being released? Logging a warning message in the `verifyMessageSignature`?

janniks:
Adding a backwards(/future) compatible function sounds interesting and more intentional (can be communicated to major sign-in projects). prototype: https://github.com/hirosystems/stacks.js/pull/1356/files

aulneau:
I have implemented an update to the functions in `micro-stacks` as long as folks are on the latest, nothing should break when this feature lands

ImTheCodeFarmer:
Is upgrading the packages enough to be up to date with this, or will any other changes be required?

janniks:
**tl;dr** Yes, _more or less_  -- Basically, new messages will be signed with the new prefixes (also for the web-wallet, as soon as it upgrades). In addition, the verification method checks both the new and the old prefix.

---

For verification (`verifyMessageSignature`): the input should be a `signature`, a _trusted_ `publicKey`, a _plaintext/unhashed_ `message`. Then everything will work as expected. The method now compares against a hashed version of the `message` (twice / aka with both prefixes: old and new). If any of the two are valid, the method returns `true`.

In any case make sure to additionally check that the publicKey matches the address that is claims to be. [`publicKeyToAddress `](https://stacks.js.org/modules/transactions.html#publicKeyToAddress)

---

We'll also add more methods for these topics in the future to make them easier 🔐

PR #1330, opened 2022-07-26
docs: update readme for post conditions

### LINKED ISSUE #1326: Misleading post-condition behavior
Hi,

The post-condition codes for NFTs are misleading.  The blockchain's post-condition processor does not check whether or not an NFT is owned or not owned.  It checks whether or not a principal _sent_ or _did not send_ an NFT.  Post-conditions, in general, do not verify anything about the recipient of an asset; they only pertain the the sender of the asset.  If you want to verify conditions about asset recipients, you'll need to do so in Clarity.

I've had to correct at least ~~one~~ two developer blog posts and one security audit finding that left users vulnerable to NFT theft because they were incorrectly relying post-conditions to check the NFT recipient (and they both cite the use of the words `Owns` and `DoesNotOwn` to describe the expected behavior).

I think some kind of warning in the post-conditions section of the `transactions` package README is necessary to correct this misconception.  Perhaps it would also be prudent to change the variable names, but I don't know what the fall-out of that would be at this point in time.

EDIT: now it's two blog posts, and one security audit that got tripped

janniks:
- adds warning on post-conditions

---

I would propose changing these:
- `Owns` → `Sends`
- `DoesNotOwn` → `DoesNotSend`

The enum values stay the same, so I don't think there is anything to worry about. But I would plan this for a major release (as the old keys would disappear and break codebases).

### DIFF
--- a/packages/transactions/README.md
+++ b/packages/transactions/README.md
@@ -458,6 +458,12 @@ const contractFungiblePostCondition = makeContractFungiblePostCondition(
 
 ### Non-fungible token post condition
 
+> **Warning**
+> The post-condition codes for NFTs are currently misleading. The Stacks blockchain's post-condition processor does NOT check ownership.
+> It checks whether or not a principal **sent** or **did not send** an NFT.
+> Post-conditions can NOT verify anything about the recipient of an asset.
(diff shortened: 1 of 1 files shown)


friedger:
LGTM

PR #1341, opened 2022-08-18
feat: support `versioned-smart-contract` tx types

### LINKED ISSUE #1339: [Stacks 2.1] Support `versioned-smart-contract` transactions
See https://github.com/stacks-network/stacks-blockchain/pull/3192 which introduced the new tx type.

The encoding for the new format is fairly straightforward. Here's another PR https://github.com/hirosystems/stacks-encoding-native-js/pull/5/files that accomplishes the same change for decoding these txs. It's in rust, but the serialization format is more clear in this PR / codebase. It also includes a test vector w/ a serialized `versioned-smart-contract` tx to test against.

zone117x:
> This PR was published as an *alpha* to npm with the version `4.3.6-pr.e86b982.0` — e.g. `npm install @stacks/common@4.3.6-pr.e86b982.0`

Closes https://github.com/hirosystems/stacks.js/issues/1339

I need this change for some API testing & development for Stacks 2.1. 

This adds an optional arg `clarityVersion` to the smart contract builder functions. If specified, a `versioned-smart-contract` type will be created. If not specified, it will default to the regular `smart-contract`.

Example:
```ts
const transaction = await makeContractDeploy({
  contractName,
  codeBody,
  senderKey,
  clarityVersion: ClarityVersion.Clarity2,
});
```

I have no strong feelings on how these function signatures look. An alternative approach could be using entirely new builder functions for this tx type, rather than an optional arg to the existing functions.

I verified that the serialized payload generated in this PR is valid with the Rust deserialization code at https://github.com/hirosystems/stacks-encoding-native-js/pull/5/files

@janniks feel free to take over this PR and make any changes

### DIFF
--- a/packages/transactions/tests/payload.test.ts
+++ b/packages/transactions/tests/payload.test.ts
@@ -7,13 +7,14 @@ import {
   createCoinbasePayload,
   createContractCallPayload,
   createTokenTransferPayload,
+  VersionedSmartContractPayload,
 } from '../src/payload';
 
 import { serializeDeserialize } from './macros';
 
 import { trueCV, falseCV, standardPrincipalCV, contractPrincipalCV } from '../src/clarity';
 
-import { COINBASE_BUFFER_LENGTH_BYTES, StacksMessageType } from '../src/constants';
+import { ClarityVersion, COINBASE_BUFFER_LENGTH_BYTES, StacksMessageType } from '../src/constants';
 import { principalToString } from '../src/clarity/types/principalCV';
 
 test('STX token transfer payload serialization and deserialization', () => {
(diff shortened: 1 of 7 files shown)


packages/transactions/tests/payload.test.ts
```diff
@@ -117,6 +118,30 @@ test('Smart contract payload serialization and deserialization', () => {
   expect(deserialized.codeBody.content).toBe(codeBody);
 });
 
+test('Versioned smart contract payload serialization and deserialization', () => {
+  const contractName = 'contract_name';
+  const codeBody =
+    '(define-map store ((key (buff 32))) ((value (buff 32))))' +
```
janniks, line 124: As far as I can tell this is the came contract as `fs.readFileSync('./tests/contracts/kv-store.clar')`, we could reuse that file for this test as well...

janniks (approved):
Clean, and fast 🏎

janniks:
Maybe we could add "test vector" style tests with hardcoded serialized txs (which match the rust impl). Thoughts? 💭

This PR will be deployed to npm via github actions, is that enough for testing and development? (I'm thinking about if it makes sense to hold off releasing these additions to be closer to 2.1 launch; maybe bundle them together) 🤷‍♂

zone117x:
@janniks I'm already using this code for local API testing and development. Deployment isn't a blocker for me. But yeah I think if we do deploy this, it would make more sense if it was tagged as alpha on npm. I don't think the Stacks 2.1 changes around all of this are finalized (the serialization format could change, but unlikely).

janniks:
📦 Released under `alpha` released as @`4.4.0-stacks2.1-alpha.0`

zone117x:
@janniks I just created a new branch `stacks-2.1` (from `master`), and switched this PR to merge into that. That way, we can merge this one, then I can submit additional PRs against the `stacks-2.1` branch until we're ready for release. Are you okay with that?

PR #1343, opened 2022-08-22
feat!: switch from buffer to uint8array

janniks:
> This PR was published to npm with the version `4.3.6-pr.21ac915.0`
> e.g. `npm install @stacks/common@4.3.6-pr.21ac915.0 --save-exact`

## Remove `Buffer` and replace with `Uint8Array`

- removes `Buffer` where possible
- removes `@stacks/keychain` package
- updates MIGRATION.md to help developers replace `Buffer` (as well as previous/upcoming migrations)
- the only place `Buffer` will remain for a bit, is for using the triplesec wallet encryption with the `@stacks/wallet-sdk` package (this is mostly only used by the stacks-web-wallet, which provides a callback of choice)

---

- [x] test with the Stacks.js starters: https://github.com/hirosystems/stacks.js-starters/pull/12
- [ ] squash-merge PR (`feat!: switch from buffer to uint8array`)

```
BREAKING CHANGE:
To reduce the bundle sizes of applications using Stacks.js we are moving away from Buffer (a polyfill to match Node.js APIs) to Uint8Arrays (which Buffers use in the background anyway). To make the switch easier we have introduced a bunch of methods for converting between strings and Uint8Arrays: `hexToBytes`, `bytesToHex`, `utf8ToBytes`, `bytesToUtf8`, `asciiToBytes`, `bytesToAscii`, and `concatBytes`.
```


janniks:
This is finally ready for review 🎉 A big step in making Stacks.js more modern again...

To make sure everything works together, this is one huge PR. The resulting packages have also been smoke-tested with @stacks/connect and the starter-templates: https://github.com/hirosystems/stacks.js-starters/pull/12

It might make sense to go-over the individual commits starting with `chore: remove buffer from XYZ` and then the remaining commits together. 
The GitHub `Files changed` UI makes this fairly straight-forward:

Screen Shot 2022-09-06 at 18 25 48

...but it's still a lot of code. We could also try tools that allow for more granular approval of commits/stacks, like https://graphite.dev

For now, I'm planning on squash-merging this into a `next` branch, to collect features for our next major release.

🙏 Let me know what you think @zone117x or if there might be other reviewers who would make sense for this.

packages/encryption/tests/encryption.test.ts
```diff
@@ -518,10 +474,16 @@ test('bn-padded-to-64-bytes', () => {
...
-  const bnBuffer = getBufferFromBN(BigInt(123));
+  const bnBytes = getBytesFromBN(BigInt(123));
+
+  expect(bnBytes.byteLength).toEqual(32);
+  expect(bytesToHex(bnBytes)).toEqual(getHexFromBN(BigInt(123)));
+});
 
-  expect(bnBuffer.byteLength).toEqual(32);
-  expect(bnBuffer.toString('hex')).toEqual(getHexFromBN(BigInt(123)));
+test('something', () => {
```
zone117x: Should be removed

packages/storage/src/fileContentLoader.ts
```diff
@@ -1,28 +1,29 @@
...
 
   readonly contentByteLength: number;
 
-  private loadedData?: Promise<Buffer>;
+  private loadedData?: Promise<Uint8Array>;
 
   static readonly supportedTypesMsg =
     'Supported types are: `string` (to be UTF8 encoded), ' +
-    '`Buffer`, `Blob`, `File`, `ArrayBuffer`, `UInt8Array` or any other typed array buffer. ';
+    '`Uint8Array`, `Blob`, `File`, `ArrayBuffer`, `UInt8Array` or any other typed array buffer. ';
```
zone117x: Nit: duplicate `Uint8Array` mentions

packages/storage/src/fileContentLoader.ts
```diff
@@ -106,26 +109,30 @@ export class FileContentLoader {
...
 
-  private async loadContent(): Promise<Buffer> {
+  private async loadContent(): Promise<Uint8Array> {
     try {
-      if (Buffer.isBuffer(this.content)) {
+      if (this.content instanceof Uint8Array) {
         return this.content;
       } else if (ArrayBuffer.isView(this.content)) {
-        return Buffer.from(this.content.buffer, this.content.byteOffset, this.content.byteLength);
+        return new Uint8Array(
```
zone117x, line 117: I think this copies the underlying ArrayBuffer, not a that big of a deal, but with a large file it might cause issues. 

The constructor `new Uint8Array(buffer, byteOffset, length)` will create a view over the existing data without copying.

Something like:
```ts
return new Uint8Array(this.content.buffer, this.content.byteOffset, this.content.byteLength);'`

File: packages/storage/src/storage.ts [hunk 4]
```diff
@@ -321,61 +323,59 @@ export class Storage {
...
-            ` match gaia address (${gaiaAddress})`
+          `Signer pubkey address (${signerAddress}) doesn't match gaia address (${gaiaAddress})`
         );
-      } else if (!verifyECDSA(fileContents, publicKey, signature)) {
+
+      if (
+        !verifyECDSA(
+          typeof fileContents === 'string'
+            ? utf8ToBytes(fileContents)
+            : new Uint8Array(fileContents),
```
zone117x, line 362: What's the type of `fileContents` here? Could possibly avoid a large copy

packages/wallet-sdk/src/triplesec.ts
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable node/prefer-global/buffer */
+import * as triplesec from 'triplesec';
+
+export function triplesecDecrypt(
+  arg: { data: Uint8Array; key: Uint8Array },
+  cb: (err: Error | null, buff: Uint8Array | null) => void
+) {
+  if (!global.Buffer) throw Error('Using triplesec currently requires polyfilling `Buffer`');
```
zone117x: `global` is still Node.js specific right? I think [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) is the new-ish environment agnostic global scope, seems to be supported everywhere now. 

Although, you might be able to just do 
```ts
if (typeof Buffer === 'undefined') { ... }
```
right?

packages/storage/src/fileContentLoader.ts
janniks: As long as the input to the Uint8Array constructor is a buffer (e.g. ArrayBuffer here) is will create a view.

But I'm also slicing(/copying) the original buffer (to avoid unwanted side-effects originally, although in this case it doesn't make too much sense (for performance reasons, and also the first if case also returns the original Uint8Array).

So, yes — thanks, I will fix this! 👍

[MDN Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/Uint8Array#syntax):
`new Uint8Array(buffer, byteOffset, length)`
When called with a `buffer`, and optionally a `byteOffset` and a `length` argument, a new typed array view is created that views the specified ArrayBuffer. The `byteOffset` and `length` parameters specify the memory range that will be exposed by the typed array view. If both are omitted, all of `buffer` is viewed; if only `length` is omitted, the remainder of buffer is viewed.

packages/storage/src/storage.ts
janniks: In this case it should already be creating a view (see https://github.com/hirosystems/stacks.js/pull/1343#discussion_r964781901)

packages/wallet-sdk/src/triplesec.ts
janniks: Yes, much better!! 🎉

packages/common/src/utils.ts
```diff
@@ -561,3 +560,24 @@ export function concatBytes(...arrays: Uint8Array[]): Uint8Array {
   }
   return result;
 }
+
+export function concatArray(elements: (Uint8Array | number[] | number)[]) {
```
zone117x, line 606: This function seems a bit dicey. If any `number` values are not 8 bit (below 0 or above 255), the Uint8Array constructor silently overflows/underflows the values, which could lead to hard to catch bugs. E.g.:
```ts
const bytes = [...new Uint8Array([1, 2, 3, 300])];
// [1, 2, 3, 44]

const bytes = [...new Uint8Array([1, 2, 3, -300])]
// [1, 2, 3, 212]

const bytes = [...new Uint8Array([1, 2, 3, 3.14])]
// [1, 2, 3, 3]
```

There could be check that `number` values are in the expected range. Is anything dependent on the `number[]` type right now?

janniks: Yes, I'm not super content with this yet...
It was an effort to reduce the need for custom methods and the hidden state in the `BufferArray` class. Used for serialization.

Screen Shot 2022-09-07 at 14 44 26

_screenshot, because GitHub .md can't do diff and language highlighting at the same time._

The new `concatArray` basically flattens the input to a byte list. I'll add the same check which `appendByte` from the original `BufferArray` used. 👍

zone117x (approved):
This looks great @janniks!

PR #1350, opened 2022-09-02
chore: add deprecation notice to keychain package methods

janniks:
> This PR was published as an *alpha* to npm with the version `4.3.6-pr.e161700.0` — e.g. `npm install @stacks/common@4.3.6-pr.e161700.0`

- add deprecation docstring to methods which will be removed soon
- this should make updating easier (users can update to the latest minor; replace deprecated methods; update major without too many surprises)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22389,13 +22389,13 @@
     },
     "packages/auth": {
       "name": "@stacks/auth",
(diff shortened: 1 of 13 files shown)


rafaelcr (approved):
LGTM

PR #1354, opened 2022-09-13
[2.1] Stacking pox-2

### LINKED ISSUE #1347: [Stacks 2.1] Support Stacks 2.1 in `@stacks/stacking` (PoX-2)
A PoX-2 contract will be deployed after the Stacks 2.1 hardfork goes live. Several new functions will be available:
 * [x] `stack-extend` / `delegate-stack-extend` https://github.com/stacks-network/stacks-blockchain/pull/2755
 * [x] `stack-unlock` https://github.com/stacks-network/stacks-blockchain/issues/2534
 * [x] `stack-increase` https://github.com/stacks-network/stacks-blockchain/pull/3282
 * [x] Native segwit address support: https://github.com/stacks-network/stacks-blockchain/pull/3283
   This changes the signatures for functions (e.g. `stack-stx`, `delegate-stack-stx`, etc) to use 32 bytes rather than 20 for address hashbytes, and expands the number of valid address version bytes. For for context see https://github.com/stacks-network/stacks-blockchain/issues/2586 and https://github.com/stacks-network/stacks-blockchain/pull/3245).

It probably makes sense to create a new set of sister function calls for PoX-2, and detect if the new functions should be used via an RPC call to determine if PoX-2 is live. The library needs to detect if `Period 3` has activated -- this is after the 2.1 fork, _and_ after PoX cycle (N+1), see https://github.com/stacksgov/sips/blob/c0fb33e0fc2b62e8e6d4ed85fb4b4aa289bb6042/sips/sip-015/sip-015-network-upgrade.md#specification

It may be possible to detect if PoX-2 is ready by checking if `GET /v2/pox` returns `{"contract_id":"ST000000000000000000002AMW42H.pox-2"}`, but that appears to be unimplemented as of opening this issue: https://github.com/stacks-network/stacks-blockchain/issues/3262

In the meantime, something like this should work:
```ts
async function isPoX2Ready(): Promise<boolean> {
  const req = await fetch(`/v2/data_var/ST000000000000000000002AMW42H/pox-2/configured?proof=0`);
  const body = await req.text();
  if (req.ok) {
    return JSON.parse(body)['data'] === '0x03'; // Clarity boolean-true
  } else if (req.status === 404 && body === 'Data var not found') {
    return false;
  } else {
    throw new Error(`Unexpected response: ${req.status} - ${body}`);
  }
}
```

janniks:
> This PR was published to npm with the version `5.0.3-pr.50a78ae.0`
> e.g. `npm install @stacks/common@5.0.3-pr.50a78ae.0 --save-exact`

## Methods for the upcoming @stacks/stacking changes to PoX-v2

* adds methods new stacking methods
* supports new btc addresses
* adds helpers for pox transition
* closes #1347 
* adds period checking ~and payload modification~, to avoid failing txs
* updates some (more or less ignorable) test configs
- [ ] one test is skipped waiting for transition unlock events
- [x] squash this PR!
```
feat: Add PoX-2 methods to @stacks/stacking client

A StackingClient will now automatically detect which PoX version to use for Stacking.
Adds the following PoX Stacking methods to the StackingClient: `stackExtend`, `stackIncrease`, `delegateStackExtend`, `delegateStackIncrease`, `stackAggregationCommitIndexed`, `stackAggregationIncrease`.
Adds the following helper methods to the StackingClient: `getAccountExtendedBalances`, `getAccountBalanceLocked`, `getDelegationStatus`, `getRewardSet`.
```

---

**For Follow-Up PRs:**
- Optimize period detection to make less network requests _(semver compatible)_
- Add explicit `poxContract` parameter to stacking methods, for users to choose specifically _(semver compatible)_


janniks:
@zone117x What's the best way to detect period-2 via API endpoints?
I've tried using the docker regtest env with tag `stacks2.1-transition`, but period 1 and 2 seem indistinguishable from each other.

e.g. Running `docker run -p "20443:20443" -e "MINE_INTERVAL=30s" -e "STACKS_21_HEIGHT=110" -e "STACKS_POX2_HEIGHT=120" zone117x/stacks-api-e2e:stacks2.1-transition` :
These are the /v2/pox outputs for block `108` (before 2.1) and `111` (after 2.1, but before pox-2)
_differing lines are highlighted in blue/gray_

zone117x:
> @zone117x What's the best way to detect period-2 via API endpoints?

IIRC, during period 2, `contract_id` will be `.pox-2`, yet `contract_versions[1].activation_burnchain_block_height` will be less than `current_burnchain_block_height`.

janniks:
my bad, was missing a mock — now all tests should work 👍

zone117x (commented):
This is looking great @janniks!
Should there be an explicit check to ensure a segwit address is not allowed when pox-v1 is still in use?

janniks:
Good call! ~I think the ABI would throw for a buff too large, but an explicit error message would be better~ — will add!

EDIT: Also just realized one of the native segwits has 20 bytes, so that one would fall through...

packages/stacking/src/utils.ts
```diff
@@ -23,68 +37,99 @@ export class InvalidAddressError extends Error {
...
+  if (witnessVersion == 0)
+    throw new Error('Addresses with witness version 1 should be encoded in bech32');
+
+  return {
+    witnessVersion,
+    data: bech32m.fromWords(bech32MWords.slice(1)),
+  };
+}
+
+function decodeNativeSegwitBtcAddress(btcAddress: string): {
```
zone117x, line 98: Nit: the mildly inefficient (and mildly annoying in debuggers) try/catch here can be replaced with something like the following:
https://github.com/hirosystems/stacks-blockchain-api/blob/ae0eb65c4d6901db172f2b4ea751817d8ef8c05c/src/pox-helpers.ts#L180-L189

packages/stacking/tests/utils.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    expect(decoded.data).toEqual(hexToBytes(expectedHash));
+    expect(decoded.data).toHaveLength(expectedLength);
+
+    const tuple = poxAddressToTuple(address);
+    expect(tuple.data['version'].buffer).toHaveLength(1);
+    expect(tuple.data['version'].buffer[0]).toBe(expectedVersion);
+    expect(tuple.data['hashbytes'].buffer).toEqual(hexToBytes(expectedHash));
+    expect(tuple.data['hashbytes'].buffer).toHaveLength(expectedLength);
+  }
+);
```
zone117x, line 80: Could you please add additional test vectors here for more address type permutations? For example, could you take tests/vectors from here: https://github.com/hirosystems/stacks-blockchain-api/blob/ae0eb65c4d6901db172f2b4ea751817d8ef8c05c/src/tests/helpers-tests.ts#L193-L530

janniks: I've [added](https://github.com/hirosystems/stacks.js/pull/1354/commits/2029a3a17e38a3d2a7739f7b2b71ef50640c04aa) a decode/encode test suite with those test vectors. Do you think it should go further and also duplicate some of the tests you linked here?

Also, do you know how I could setup tests with vectors closer to the `hashbytes` on the blockchain — i.e. compare to hashbytes that I know will be received when paying out through blockchain/miner code. I'm trying to sift through the codebase (`reward_address.bytes()`, `PoxAddress`, etc.), but haven't found any useful tests...

zone117x: Thanks! I think there still might be missing vectors from the API repo, for example the P2TR ones. Or am I just missing them?

zone117x: > Also, do you know how I could setup tests with vectors closer to the hashbytes on the blockchain — i.e. compare to hashbytes that I know will be received when paying out through blockchain/miner code. I'm trying to sift through the codebase (reward_address.bytes(), PoxAddress, etc.), but haven't found any useful tests...

Not really. I think that would require more of an end-to-end test. Perhaps you could grab those outputs manually to use as test vectors?

janniks: > Thanks! I think there still might be missing vectors from the API repo, for example the P2TR ones. Or am I just missing them?

You're completely right! 😬 Added now

janniks: Added a bunch more test vectors.
This was the only one I couldn't reproduce: https://gist.github.com/janniks/b7ba98b1770228880ef05fc96ccd14cb#file-btc-address-test-vectors-js-L346-L354

janniks:
@zone117x ready for another review
- Requested btc address test vectors were added
- Missing `aggregation..` methods were added

janniks:
Noticed there was still incorrect Period definitions in there. 3 was basically 2b, now I added all periods and more explicit tests. ✅

zone117x (approved):
LGTM, thanks @janniks!

PR #1355, opened 2022-09-20
Updated Faucet Description and Example in CLI Readme

BowTiedDeployer:
## Description
As a Stacks developer, I want to get STX in my wallet easily directly from the place where I am doing most of the blockchain calls. I don't want to go to the stacks explorer and press some clicks for this, I want it done with one line directly in the cli. 

I was looking at the documentation for it when I saw it was a `duplicate text from the encrypted keychain` explanation. I changed it to be the description and example specific for the faucet.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -280,15 +280,15 @@ Example:
 ```
 ### faucet
 ```shell script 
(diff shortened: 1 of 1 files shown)


janniks (approved):
Thanks 🙏

PR #1381, opened 2022-10-19
Improve types for profiles

friedger:
## Description

As a developer, I want to use types when creating or updating a public profile.

For details refer to issue [https://github.com/hirosystems/stacks-wallet-web/issues/2544](https://github.com/hirosystems/stacks-wallet-web/issues/2544)

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

This should be part of a SIP

- [x] Tag 1 of @janniks  @kyranjamie


packages/profile/src/index.ts
```diff
@@ -14,3 +14,5 @@ export {
   verifyProfileToken,
   extractProfile,
 } from './profileTokens';
+
+export { PublicProfile, PublicPersonProfile } from './types';
```
janniks, line 18: Curious, do we need "Public" in the name? Are there Private alternatives to need a distinction?

friedger: We should use a name that references objects like the payload in https://gaia.blockstack.org/hub/1Maw8BjWgj6MWrBCfupqQuWANthMhefb2v/0/profile.json
I wanted to make a distinction to Person schema, as we have more attributes.

It is public data, users should be aware of that. Furthermore, there is a `Profile` class as a helper class for the `PublicProfile` type.

There is also #1382 that is related to this topic.

Open to any suggestions.

janniks:
Merged via #1383

PR #1402, opened 2022-11-30
fix: remove stacking into period3 check

janniks:
> This PR was published to npm with the version `6.0.2-pr.da30fce.0`
> e.g. `npm install @stacks/common@6.0.2-pr.da30fce.0 --save-exact`

> This bug was a remnant of not understanding the periods initially.

I believe, the check to see if something overlaps into Period 3 is no longer needed (and was implemented incorrectly before).
It's not needed since Stacks.js will prefer PoX-2 if available, so @stacks/stacking shouldn't be able to generate txs that target PoX-1 after the 2.1 fork.

### fix:
- removes `ensurePox1DoesNotCreateStateIntoPeriod3` check

### test:
- adds more extensive tests for correct contracts usage

> I will deprecate the faulty version on npm after releasing this fix.

### DIFF
--- a/packages/stacking/src/index.ts
+++ b/packages/stacking/src/index.ts
@@ -35,7 +35,6 @@ import {
 import { PoxOperationPeriod, StackingErrors } from './constants';
 import {
   ensureLegacyBtcAddressForPox1,
-  ensurePox1DoesNotCreateStateIntoPeriod3,
   ensurePox2IsLive,
   poxAddressToTuple,
   unwrap,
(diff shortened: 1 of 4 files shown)


zone117x (commented):
To clarify my understanding, this change ensures "if stacking and if PoX-2 contract is deployed, then submit tx for the pox-2 contract", right?

janniks:
> To clarify my understanding, this change ensures "if stacking and if PoX-2 contract is deployed, then submit tx for the pox-2 contract", right?

Essentially, the behavior already previously was: `IF  now == Period1 THEN use pox-1 ELSE use pox-2`, where `Period1 = beforeFork || isPox2NotYetConfigured`. This behavior is unchanged by this PR.

---

This PR just removes the `ensurePox1DoesNotCreateStateIntoPeriod3` check.

This check was buggy, since it didn't check the current contract at all. And the check doesn't make sense now anyway — since non-Period1 will always use pox-2. So, either overlapping-into-Period3 would be allowed (in Period1 via pox-1) or we use pox-2 anyway.

zone117x (approved):

PR #1412, opened 2022-12-11
docs: update sponsored transactions docs

obycode:
Set the fee in the initial transaction to 0, and set the sponsor nonce in the sponsor's transaction.


janniks (approved):
Thanks 🙏

PR #1415, opened 2022-12-13
docs: fix import

edu-stx:
## Description

I believe the examples intended to use the `broadcastTransaction` function, but instead are using the the `broadcastResponse` function, which doesn't seem to exist.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

no

no

## Testing information
n/a

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

Tagging @yknl for review. Not sure how to update the changelog, let me know if this PR requires further work.


janniks:
Thanks, good catch! 🙏

PR #1422, opened 2023-01-04
feat: add estimation fallback

janniks:
> This PR was published to npm with the version `6.0.3-pr.5dbcd63.0`
> e.g. `npm install @stacks/common@6.0.3-pr.5dbcd63.0 --save-exact`

- adds a fallback estimation if `NoEstimateAvailable` is received from the stacks node
- this is closer to legacy behavior and once existed in a way (before transaction estimates were a thing and everything was a transfer estimate)


obycode (approved):
LGTM! And I worked with a user today that tested it out this branch successfully!

PR #1432, opened 2023-01-26
feat: allow string name for AnchorMode

zone117x:
> This PR was published to npm with the version `6.1.1-pr.35dd959.0`
> e.g. `npm install @stacks/common@6.1.1-pr.35dd959.0 --save-exact`

Small DX improvement by allowing `anchorMode` to be specified as a string value. `AnchorMode` is commonly the only enum import required when building txs. This PR makes it possible to specify the value as a string, similar to the `network: 'mainnet' | 'testnet'` option.

Example
```ts
import { AnchorMode, makeSTXTokenTransfer } from '@stacks/transactions';

const transaction = await makeSTXTokenTransfer({
  recipient: 'SP3FGQ8Z7JY9BWYZ5WM53E0M9NK7WHJF0691NZ159',
  amount: 12345n,
  senderKey: 'b244296d5907de9864c0b0d51f98a13c52890be0404e83f273144cd5b9960eed01',
  network: 'mainnet',
  memo: 'test',
  nonce: 0n,
  fee: 200n,
  anchorMode: AnchorMode.Any, // can now be 'any'
});
```

### DIFF
--- a/packages/transactions/src/builders.ts
+++ b/packages/transactions/src/builders.ts
@@ -32,6 +32,7 @@ import {
   RECOVERABLE_ECDSA_SIG_LENGTH_BYTES,
   StacksMessageType,
   ClarityVersion,
+  AnchorModeName,
 } from './constants';
 import { ClarityAbi, validateContractCall } from './contract-abi';
 import { NoEstimateAvailableError } from './errors';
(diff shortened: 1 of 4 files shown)


janniks (approved):
Very nice change! 🤓 Bye bye imports

PR #1459, opened 2023-03-14
feat: strict ECDSA Signatures

fpbgg:
> This PR was published to npm with the version `6.2.2-pr.4b3d469.0`
> e.g. `npm install @stacks/common@6.2.2-pr.4b3d469.0 --save-exact`

Strict signatures contain the following properties:

1. Signatures encoded as (R,S), S is always smaller than the group order.
2. The signature is calculated over the hash of the public key and the message

This second property satisfies the requirements for Universal Exclusive Ownership, as defined in [this paper](https://www.bolet.org/~pornin/2005-acns-pornin+stern.pdf) (section 3.3).

Strict signatures are more suitable in applications where it's undesirable for `s1 = s2` given `s1 = sign(k1, msg1)` and `s2 = sign(k2, msg2)`. Direct ECDSA permits this condition (even if `msg1 = msg2`).

#### Breaking change?

No, it's a new feature.

### Example

---

### Checklist

### DIFF
--- a/packages/encryption/src/cryptoRandom.ts
+++ b/packages/encryption/src/cryptoRandom.ts
@@ -1,17 +1,17 @@
 import { utils } from '@noble/secp256k1';
 
 /**
- * Reexports @noble/secp256k1's randombytes
+ * Reexports @noble/secp256k1's `randomBytes`.
  * Generates bytes with random bytes of given length
- * @param bytesLength an optional bytes length, default `32` bytes
+ * @param bytesLength an optional bytes length, default `32`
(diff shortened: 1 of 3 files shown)


janniks:
Thanks, great change. Maybe we should mention the currently implemented SIP-018 somewhere, this will stay unstrict 🤔

janniks (approved):

janniks:
Added some test cases

scottsolowild:
@fpbgg I notice this PR is in draft. Is it intended for merge?

PR #1460, opened 2023-03-16
v18.1.6 - delegate skeleton for multisig

janniks:
- adds `makeDelegate` function for stx over btc wire formats

- ~the wire format in SIP-015 might be incorrect, seems the last bit-length is off (see comment below)~ using same wire-format as rust code now (u64 as last part) 

CC @donpdonp

### DIFF
--- a/src/operations/skeletons.js
+++ b/src/operations/skeletons.js
@@ -133,6 +133,9 @@ export class BlockstackNamespace {
   }
 }
 
+function isNullish(value: any): boolean {
+  return value === null || value === undefined
+}
 
 function asAmountV2(amount: AmountType): AmountTypeV2 {
   // convert an AmountType v1 or v2 to an AmountTypeV2.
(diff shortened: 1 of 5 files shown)


src/operations/skeletons.js
```diff
@@ -824,3 +829,65 @@ export function makeV2PreStxOpSkeleton(
...
+  untilBurnHeight: ?BigInteger
+) {
+  /*
+  Format:
+  https://github.com/stacksgov/sips/blob/feat/sip-015/sips/sip-015/sip-015-network-upgrade.md#new-burnchain-transaction-delegate-stx
+
+  0      2   3                          19       20                                24       25                       33
+  |------|---|--------------------------|--------|---------------------------------|--------|------------------------|
+     magic  op    uSTX to delegate (u128)    00|01   output of pox-addr tuple (?u32)    00|01    until-burn-ht (?u128)
+
```
janniks, line 844: Not sure what the correct implementation is -- @donpdonp can you point me to the rust file maybe?

kantai: Here's the rust parse implementation:

https://github.com/stacks-network/stacks-blockchain/blob/master/src/chainstate/burn/operations/delegate_stx.rs#L34

donpdonp: src/chainstate/burn/operations/delegate_stx.rs:34 is the rust implementation. while the diagram in skeletons.js has a bit more detail, the implementation in delegate_stx.rs conforms to the diagram.

donpdonp: correction, @janniks pointed out to me that the rust is reading u64 (8 bytes) when the spec says u128.

Which bit-width is best to hold btc-burn height? The bitcoin blockchain doesn't encode a height (https://en.bitcoin.it/wiki/Block), so the count is unbounded with the length of the chain.

 `2**32 /6/24/365 => 81,715 years` of time is provided for by a u32 worth of btc blocks.

kantai: `u64` is the way to go -- the parser will only read 8 big endian bytes.

kantai:
I'm getting build errors (during browserify) due to the `stacks.js` include:

```sh
$ npm run build

> blockstack@18.1.5 build
> npm run compile && npm run browserify


> blockstack@18.1.5 compile
> rm -rf lib; babel src -d lib

Successfully compiled 48 files with Babel (1864ms).

> blockstack@18.1.5 browserify
> ./node_modules/.bin/browserify lib/index.js --standalone blockstack -o ./dist/blockstack.js


/Users/aaron/devel/stacks.js/node_modules/@stacks/transactions/dist/transaction.js:20
                ...payload,
                ^
ParseError: Unexpected token
```

I tried updating the babel dependencies, which improved the error message slightly:

```diff
diff --git a/.babelrc b/.babelrc
index 6bd4d284..900e2920 100644
--- a/.babelrc
+++ b/.babelrc
@@ -1,11 +1,9 @@
 {
-  "presets": ["env", "flow"],
+  "presets": ["@babel/preset-env", "@babel/preset-flow"],
   "plugins": [
-    ["transform-runtime", {
+    ["@babel/transform-runtime", {
       "helpers": false,
-      "polyfill": false,
       "regenerator": true,
-      "moduleName": "babel-runtime"
     }]
   ]
 }
diff --git a/package.json b/package.json
index 589bf3ca..df19fe23 100644
--- a/package.json
+++ b/package.json
@@ -78,14 +78,14 @@
     }
   ],
   "devDependencies": {
-    "babel-cli": "^6.26.0",
+    "@babel/cli": "^7.21.0",
     "babel-eslint": "^8.2.5",
-    "babel-plugin-transform-runtime": "^6.23.0",
-    "babel-polyfill": "^6.26.0",
-    "babel-preset-env": "^1.7.0",
-    "babel-preset-flow": "^6.23.0",
+    "@babel/plugin-transform-runtime": "^7.0.0",
+    "@babel/polyfill": "^7.12.0",
+    "@babel/preset-env": "^7.0.0",
+    "@babel/preset-flow": "^7.0.0",
     "blue-tape": "^1.0.0",
-    "browserify": "^13.1.1",
+    "browserify": "^17.0.0",
     "documentation": "^8.0.0",
     "eslint": "^4.19.1",
     "eslint-config-airbnb": "^17.0.0",
```

janniks:
Ah sorry, pushed the latest now. I removed the stacks.js deps -- after reading the rust part I realized this wasn't needed.

I thinks it works for my tests, at least the tx are able to broadcast to testnet.

janniks:
Ready for a final review -- First tx worked and I'm testing more variations tomorrow.

kantai (approved):
LGTM!

janniks:
I tested a bunch of variations and the final version looks good 😎 e.g. [tx](https://explorer.stacks.co/txid/0x2d5bc598ea6106e6818f4da9337984608585f9f691c34befc4d5a5d59defe03b?chain=testnet)
~With @kantai green light,~ I will publish to `18.1.6` under the npm `v18-support` tag

- The only [unreviewed change](https://github.com/hirosystems/stacks.js/compare/96ceefb46f8a6dd614efaa118f4091cdd1c1b6d5..21458d62706475c7ebf0edce7eec30035daa4c31) was a bug leftover from the u128→u64 change.

---

~_(I know I mentioned the version `18.1.6` but a 'feat' bump to `18.2.0` is probably more correct -- I don't have strong preferences though)_~ `18.2.0` already exists, `18.1.6` it is 🎬

PR #1466, opened 2023-03-24
feat: add pc post condition builder

janniks:
> This PR was published to npm with the version `6.4.1-pr.eccf9dc.0`
> e.g. `npm install @stacks/common@6.4.1-pr.eccf9dc.0 --save-exact`

### `Pc` Behavioral Builder Pattern ✨

- Inspired by Ruby style behavioral/readable chained method calls.
- Tries to fix very cumbersome current post-condition construction.

This PR introduces an easier way of constructing post-conditions. These can be very confusing for developers. This solution doesn't solve all problems, but is a good start imho.

> **Note**
> Marking as `@beta` for now to collect feedback from builders.
> I believe the syntax can be generally kept, but maybe folks have different opinions about capitalization, asset/nft/name/id naming, etc.

#### Before
```ts
import { FungibleConditionCode, makeStandardSTXPostCondition } from '@stacks/transactions';
const pc = makeStandardSTXPostCondition(
  'ST1WT20920NVRQ892MS535R7XEMV6KD6M6X2HQPK3',
  FungibleConditionCode.Equal,
  100
);
```
```ts
import { createAssetInfo, NonFungibleConditionCode.Sends, makeContractNonFungiblePostCondition, uintCV } from '@stacks/transactions';
const pc = makeContractNonFungiblePostCondition(
  'ST13F481SBR0R7Z6NMMH8YV2FJJYXA5JPA0AD3HP9',
  'subnet-v1',
  NonFungibleConditionCode.Sends,
  createAssetInfo('ST13F481SBR0R7Z6NMMH8YV2FJJYXA5JPA0AD3HP9', 'super-nft', 'super'),
  uintCV(21)
);
```
#### After
```ts
import { Pc } from '@stacks/transactions';
const pc = Pc.principal('ST1WT20920NVRQ892MS535R7XEMV6KD6M6X2HQPK3').willSendEq(100).ustx();
```

```ts
import { Pc, uintCV } from '@stacks/transactions';
const pc = Pc.principal(`ST13F481SBR0R7Z6NMMH8YV2FJJYXA5JPA0AD3HP9.subnet-v1`)
  .willSendAsset()
  .nft(`ST13F481SBR0R7Z6NMMH8YV2FJJYXA5JPA0AD3HP9.super-nft`, 'super', uintCV(21));
```
---

### Checklist

### DIFF
--- a/packages/transactions/src/pc.ts
+++ b/packages/transactions/src/pc.ts
@@ -1,0 +1,235 @@
+import { IntegerType } from '@stacks/common';
+import {
+  makeContractFungiblePostCondition,
+  makeContractNonFungiblePostCondition,
+  makeContractSTXPostCondition,
+  makeStandardFungiblePostCondition,
+  makeStandardNonFungiblePostCondition,
+  makeStandardSTXPostCondition,
+} from './builders';
+import { ClarityValue } from './clarity';
+import { FungibleConditionCode, NonFungibleConditionCode } from './constants';
+import { createAssetInfo } from './postcondition-types';
+
+/// `Pc.` Post Condition Builder
+//
+// This is a behavioral helper interface for constructing post conditions.
+//
+// The general pattern is:
+//   PRINCIPAL -> [AMOUNT] -> CODE -> ASSET
+//
+
+/**
+ * An address string encoded as c32check
+ * @internal
+ */
+type AddressString = string;
+
+/**
+ * A contract identifier string given as `<address>.<contract-name>`
+ * @internal
+ */
+type ContractIdString = `${string}.${string}`;
+
+/**
+ * ### `Pc.` Post Condition Builder
+ * @param {AddressString | ContractIdString} principal The principal to check, which should/should-not be sending assets. A string in the format "address" or "address.contractId".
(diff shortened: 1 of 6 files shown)


packages/transactions/src/pc.ts
```diff
@@ -0,0 +1,229 @@
...
+      return makeContractSTXPostCondition(this.address, this.contractName, this.code, this.amount);
+    }
+    return makeStandardSTXPostCondition(this.address, this.code, this.amount);
+  }
+
+  /**
+   * ### Fungible Token Post Condition
+   * ⚠ Amount of fungible tokens is denoted in the smallest unit of the token
+   */
+  ft(contractId: ContractIdString, tokenName: string) {
```
zone117x, line 222: For this and also nfts, is it common to use a fully-qualified identifier in the format `<principal>::<token_name>`? If so, might be useful to also support that format as an arg?

zone117x (approved):
Love this builder pattern!

packages/transactions/src/pc.ts
janniks: Oh nice, good to know -- will add that

PR #1474, opened 2023-04-06
fix: add StacksDevnet constructor, closes #1470

### LINKED ISSUE #1470: Add support for devnet
### Problem

I would like to easily switch between devnet (clarinet integrate) and testnet or mainnet. However, `StacksNetworks` only allows the latter two.

### Solution

Add devnet as poosible value of `StacksNetworks`.

Furthermore, define StacksDevnet with url 'http://localhost:3999'

janniks:
> This PR was published to npm with the version `6.5.1-pr.9f6c5d2.0`
> e.g. `npm install @stacks/common@6.5.1-pr.9f6c5d2.0 --save-exact`

### DIFF
--- a/packages/network/src/network.ts
+++ b/packages/network/src/network.ts
@@ -10,11 +10,11 @@ export interface NetworkConfig {
   fetchFn?: FetchFn;
 }
 
-export const StacksNetworks = ['mainnet', 'testnet'] as const;
+export const StacksNetworks = ['mainnet', 'testnet', 'devnet', 'mocknet'] as const;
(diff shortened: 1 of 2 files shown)


friedger:
It might make sense to add network mode that determines how addresses are encoded.

This is used in the explorer

janniks:
Like a `getNetworkFromAddress`? It would result in Testnet for Devnet/Mocknet addresses, since there's no difference. Is that acceptable?

friedger:
Yes, something like that

PR #1479, opened 2023-04-10
chore: add cl helper

### LINKED ISSUE #1444: [RFC] `Cl` bundled export
## Problem

- Bad Discoverability for `CV` "clarity value" types and helpers
    - Many functions are in separate packages
    - Often grouped functions use the same suffixes, no way to easily auto-complete-list `..CV` functions in editor
- Unclear concepts data-types (for newcomers); We imply a lot of knowledge with our namings of things

## Solution

Add a `Cl` export that groups all related "Clarity" methods.

Allows for better auto-complete in editors:
Screenshot 2023-02-01 at 14 39 48

Solution can be backwards compatible (essentially just wrapping the existing methods as an alias).

Packages like `@stacks/connect` can now re-export things like `Cl` to allow users to working within a single package.

janniks:
> This PR was published to npm with the version `6.5.1-pr.f11b880.0`
> e.g. `npm install @stacks/common@6.5.1-pr.f11b880.0 --save-exact`

- closes https://github.com/hirosystems/stacks.js/issues/1444

## summary

Adds a `Cl` bundled export that allows better discovery of clarity methods and less importing.

## examples

NEW
```ts
import { Cl } from '@stacks/transactions'
const tup = Cl.tuple({ a: Cl.unit(2) })
```
VS previously
```ts
import { uintCV, tupleCV } from '@stacks/transactions'
const tup = tupleCV({ a: unitCV(2) })
```

---

## discussion

I'm still in love with the idea to add the type-constructors in uppercase, e.g. `Cl.Int(2)` _(lowercase `Cl.false` won't works since `false` is a keyword)_ and keep methods lowercase, e.g. `Cl.serialize(...)` to add more distinction (and potentially group more methods under a new namespace, similar to `Buffer.from`) -- but I realize it could be confusing.. 🤷‍♂ cc @hugocaillard  @lgalabru

### DIFF
--- a/packages/transactions/src/clarity/types/listCV.ts
+++ b/packages/transactions/src/clarity/types/listCV.ts
@@ -9,7 +9,7 @@ interface ListCV<T extends ClarityValue = ClarityValue> {
 /**
  * Create list of clarity types
  *
- * @param {ClarityValue[]} values list of  ClarityValues to be converted to ListCV clarity type
+ * @param {ClarityValue[]} list of ClarityValues to be converted to ListCV clarity type
  *
  * @returns {ListCV<T>} instance of type ListCV<T> of the provided values
  *
--- a/packages/transactions/src/clarity/types/listCV.ts
+++ b/packages/transactions/src/clarity/types/listCV.ts
@@ -22,7 +22,7 @@ interface ListCV<T extends ClarityValue = ClarityValue> {
  * ```
  *
  * @visit
- * {@link https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/tests/clarity.test.ts clarity test cases for more examples}
+ * {@link https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/tests/clarity.test.ts | clarity test cases for more examples}
  */
 function listCV<T extends ClarityValue = ClarityValue>(values: T[]): ListCV<T> {
   return { type: ClarityType.List, list: values };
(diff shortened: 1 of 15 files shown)


packages/transactions/src/clarity/types/listCV.ts
```diff
@@ -9,9 +9,9 @@ interface ListCV<T extends ClarityValue = ClarityValue> {
 /**
  * Create list of clarity types
  *
- * @param {ClarityValue>values: T[]} list of  ClarityValues to be converted to ListCV clarity type
+ * @param {ClarityValue[]} values list of  ClarityValues to be converted to ListCV clarity type
```
hugoclrd:
```suggestion
 * @param {ClarityValue[]} list of ClarityValues to be converted to ListCV clarity type
```

packages/transactions/src/clarity/types/principalCV.ts
```diff
@@ -51,8 +49,8 @@ function principalCV(principal: string): PrincipalCV {
  *  // { type: 5, address: { type: 0, version: 22, hash160: 'a5d9d331000f5b79578ce56bd157f29a9056f0d6' } }
  * ```
  *
- * @visit
- * {@link https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/tests/clarity.test.ts clarity test cases for more examples}
+ * @see
```
hugoclrd, line 52: Is `see` an alias for `visit`? I see that other modified files (optionalCV, responseCV, ...) still uses `visit`

hugoclrd (approved):
LGTM, just small remarks regarding JSDocs.

I feel like `Cl.trueBool` and `Cl.falseBool` could almost be omitted (in favor of `bool(true|false)`). But I guess it makes sense to have them for completeness.

packages/transactions/src/clarity/types/principalCV.ts
janniks: Thanks I'll change all of them -- yeah, I think `@visit` technically doesn't exist in tsdoc

janniks:
I sort of agree, I don't like the `trueBool` method -- leaving out for now, we can always add it again later

PR #1491, opened 2023-05-03
fix(pox-3): update stacking lib to always use the latest activated / …

### LINKED ISSUE #1495: Update stacks.js for PoX-3 / 2.4 activation

janniks:
> This PR was published to npm with the version `6.5.3-pr.693e11b.0`
> e.g. `npm install @stacks/common@6.5.3-pr.693e11b.0 --save-exact`

### pox-3 update

- Adds a `.current` pointer to the pox operation object, which points to the latest activated pox contract version. This should future-proof the code for any further fork a bit.
- Keeps backwards compatibility with the previous regtest tests

> No additional checks are made that the latest activated pox contract version is also the "active" one according to the v2/pox endpoint. maybe this should be used instead, we're assuming the response to be correct currently

### todo
- [x] Add tests based on responses from latest (2.4) ~mocknet~regtest


zone117x (approved):
💯

PR #1520, opened 2023-07-03
next

janniks:
> This PR was published to npm with the version `next`
> e.g. `npm install @stacks/common@next --save-exact`

## The `next` major version of Stacks.js


ryanwaits (approved):
here we go 🚀

PR #1551, opened 2023-09-20
feat: add Cl.prettyPrint

hugoclrd:
> This PR was published to npm with the version `6.8.2-pr.12aa132.0`
> e.g. `npm install @stacks/common@6.8.2-pr.12aa132.0 --save-exact`

### Description

`Cl.prettyPrint` allows to format a clarity value into 

### Example

```ts
const value = Cl.tuple({
  id: Cl.uint(1),
  messageAscii: Cl.stringAscii('hello world'),
  someMessageUtf8: Cl.some(Cl.stringUtf8('hello world')),
  items: Cl.some(
    Cl.list([
      Cl.ok(
        Cl.tuple({
          id: Cl.uint(1),
          owner: Cl.some(Cl.standardPrincipal(addr)),
        })
      ),
      Cl.ok(
        Cl.tuple({
          id: Cl.uint(2),
          owner: Cl.none(),
        })
      ),
    ])
  ),
});



const result = Cl.prettyPrint(value, 2);

/*
{
  id: u1,
  messageAscii: "hello world",
  someMessageUtf8: (some u"hello world"),
  items: (some (list
    (ok {
      id: u1,
      owner: (some 'ST2CY5V39NHDPWSXMW9QDT3HC3GD6Q6XX4CFRK9AG)
    })
    (ok {
      id: u2,
      owner: none
    })
  ))
}
*/
```

---

### Checklist

- [x] Tagged @janniks for review

### DIFF
--- a/packages/transactions/src/cl.ts
+++ b/packages/transactions/src/cl.ts
@@ -16,8 +16,9 @@ import {
   stringUtf8CV,
   tupleCV,
   uintCV,
-  prettyPrint,
 } from './clarity';
+
+import { prettyPrint } from './clarity/prettyPrint';
 
 // todo: https://github.com/hirosystems/clarinet/issues/786
 
--- a/packages/transactions/src/cl.ts
+++ b/packages/transactions/src/cl.ts
@@ -16,6 +16,7 @@ import {
   stringUtf8CV,
   tupleCV,
   uintCV,
+  prettyPrint,
 } from './clarity';
 
 // todo: https://github.com/hirosystems/clarinet/issues/786
@@ -158,5 +159,7 @@ export const serialize = serializeCV;
  * @see {@link serialize}
  */
 export const deserialize = deserializeCV;
+
+export { prettyPrint };
 
 // todo: add `deserializeReadable` methods that translates enums into name strings
--- a/packages/transactions/src/cl.ts
+++ b/packages/transactions/src/cl.ts
@@ -18,7 +18,7 @@ import {
   uintCV,
 } from './clarity';
 
-import { prettyPrint } from './clarity/prettyPrint';
+export { prettyPrint } from './clarity/prettyPrint';
 
 // todo: https://github.com/hirosystems/clarinet/issues/786
 
@@ -160,7 +160,5 @@ export const serialize = serializeCV;
  * @see {@link serialize}
  */
 export const deserialize = deserializeCV;
-
-export { prettyPrint };
 
 // todo: add `deserializeReadable` methods that translates enums into name strings
--- a/packages/transactions/src/clarity/prettyPrint.ts
+++ b/packages/transactions/src/clarity/prettyPrint.ts
@@ -1,0 +1,136 @@
+/*
+  Format Clarity Values into Clarity style readable strings
+  eg:
+  `Cl.uint(1)` => u1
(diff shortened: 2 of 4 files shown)


janniks (approved):
So good ✨

packages/transactions/src/cl.ts
```diff
@@ -18,6 +18,8 @@ import {
   uintCV,
 } from './clarity';
 
+import { prettyPrint } from './clarity/prettyPrint';
+
```
janniks, line 22: I think this could even be an `export from`

packages/transactions/src/clarity/prettyPrint.ts
```diff
@@ -0,0 +1,138 @@
...
+  Format Clarity Values into Clarity style readable strings
+  eg:
+  `Cl.uint(1)` => u1
+  `Cl.list(Cl.uint(1))` => (list u1)
+  `Cl.tuple({ id: u1 })` => { id: u1 }
+*/
+
+import { BufferCV, ClarityType, ClarityValue, ListCV, TupleCV, principalToString } from '.';
+
+const byteToHex: string[] = [];
```
janniks: In Stacks.js we already have bytesToHex somewhere, so we can re-use that

janniks (changes requested):

packages/transactions/src/clarity/prettyPrint.ts
hugoclrd: Ah yes forgot about that!

packages/transactions/src/cl.ts
hugoclrd: Will try, although it's rendered a bit differently in the docs

packages/transactions/src/clarity/prettyPrint.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  if (cv.type === ClarityType.List) {
+    return formatList(cv, space, depth + 1);
+  }
+  if (cv.type === ClarityType.Tuple) {
+    return formatTuple(cv, space, depth + 1);
+  }
+
+  // make sure that we exhausted all ClarityTypes
+  exhaustiveCheck(cv);
```
janniks, line 107: Interesting pattern 👍 I like the exhaustive check. Is that why you didn't go for a switch?

hugoclrd: I generally don't use `switch` often because I find it has an odd syntax within the JS ecosystem.
But in TS it has the advantage of checking exhaustiveness. But the exhaustiveCheck can be useful to make TS happy sometimes.

If it fits within Stacks.js coding style, I can change for `switch`, let me know what you prefer. In this particular code sample it would make sense I guess

packages/transactions/src/cl.ts
hugoclrd: addressed

janniks (approved):
Let's goo!

packages/transactions/src/clarity/prettyPrint.ts
janniks: Either is fine with me, 🚀

hugoclrd:
Thanks @janniks!

PR #1564, opened 2023-10-03
Update overview.md: Teeny typo [ als -> also ]

saralab:
Update overview.md: Teeny typo [ als -> also ]

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -74,7 +74,7 @@ Below is a list of all Stacks.js libraries and a few JS libraries and helpers ma
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


max-crawford:
@saralab I actually don't know what this clause is trying to say: There is a main Stacks.js monorepo containing most of the packages, _but there are a few others often als considered as a part of "Stacks.js"._ (italicized emphasis is mine)

Is this sentence saying that there are packages that are considered part of Stacks.js, but are not part of the main monorepo? If so, which packages? Why are they not included in the monorepo? 

Should we just remove this clause altogether and effectively say 'here is the monorepo' ?

PR #1567, opened 2023-10-03
Improves readme and contribution guidelines and adds code of conduct

andresgalante:
This PR improves contribution guidelines and readme, and adds code of conduct.

Reviewers:

- @janniks for contribution guidelines
- @max-crawford for voice consistency


max-crawford (approved):
This looks good to me!

janniks:
Should we maybe think about having the contributing document in one place (rather than using templates)? And then just link to the docs page about contributing (one for all repos, assuming the process is the same), instead of each repo having a long document

andresgalante:
@janniks that is a good idea.

Some things can be extracted, but not all repos share the same guidelines. most of the instructions around running, building and testing projects are different.

PR #1596, opened 2023-11-13
🚧 WIP [next] feat!: refactor stacks.js

janniks:
> 💡 To reduce reviewing difficulty some PRs will merge into this branch. Also no real "logic" is changed here, to keep it simple.

### High-level changes:
- Move to new `@stacks/api` package (main way to do networking for users)
- Move to static "network" objects (simpler way to specify opts for methods)

---

Screenshot 2023-12-18 at 16 20 39

## Refactor 🚂

- Removes `@stacks/network` package
  - Moves most functionality to transactions package (this will become the main package allowing most folks to use it and not need to worry about other imports)
  - Move to a network OBJECT approach (similar to bitcoin-js, micro-btc-signer, and others). The network will not be an instantiated class, but a static exported object with all information that might be relevant (txVersion, magicBytes, etc.)
  - Remove `fetch` functions from network and introduce `StacksNodeApi`, which can derive it's URL from a network, but should be treated separately -- more closely mirroring to what's actually happening in the background.
  - The majority of users doesn't realize network is also setting changes for "networking"
    - Most users don't customize fetchFn or URL. For users hosting their own node the only change becomes `{ api: { url: "my-node.com" }, ...}` in the params of most relevant tx functions.
- Removes duplicate/unnecessary definitions from `@stacks/common`

> Some changes might irritate users at first, but I think it's an approach that follows the structure of the blockchain and in a way educated while using the library. _More to follow_

Curious to get feedback 🙏 @hugocaillard @zone117x @rafaelcr 

### Open Questions:
 
- Thoughts on param naming?
  - Should `url` be `baseUrl`, ~or `nodeUrl`~?
  - ~Should `StacksApi` be `StacksNode` or~ `StacksNodeApi` ✅
    - ~I like it short, but "node" could be helpful (even though most users wrap their node with the API). We could also add a "node-only mode" which ensures the requests it's trying to make exist on the v2 rather than extended.~
- What should live where?
  - ~Should we move most things into `@stacks/transactions` (is there a case for having users import from `@stacks/common` for basic interactions or should it stay for helpers used by all `@stacks/` packages?)~ *common* will be non-Stacks things, *transactions* will be the main package ✅

## Examples

Diff for most methods
```diff
const transaction = await makeSTXTokenTransfer({
  recipient,
  amount,
  senderKey,
  memo,
- network: new StacksTestnet({ url: "mynode-optional.com" }), // optional options
+ network: STACKS_TESTNET,
+ api: { url: "mynode-optional.com" } // optional param
});
```
> String enum params still work (e.g. `'testnet'`) ☝️

New API class
```ts
const api = new StacksNodeApi({ network: STACKS_TESTNET });
await api.broadcastTransaction(tx); // bundled client, discoverable API maintained endpoints/methods (with helper processing/decoding)
```


packages/transactions/src/fetch.ts
```diff
@@ -0,0 +1,569 @@
...
+export const HIRO_TESTNET_URL = 'https://stacks-node-api.testnet.stacks.co';
+export const DEVNET_URL = 'http://localhost:3999';
+
+export const BROADCAST_PATH = '/v2/transactions';
+export const TRANSFER_FEE_ESTIMATE_PATH = '/v2/fees/transfer';
+export const TRANSACTION_FEE_ESTIMATE_PATH = '/v2/fees/transaction';
+export const ACCOUNT_PATH = '/v2/accounts';
+export const CONTRACT_ABI_PATH = '/v2/contracts/interface';
+export const READONLY_FUNCTION_CALL_PATH = '/v2/contracts/call-read';
+export const MAP_ENTRY_PATH = '/v2/map_entry';
```
zone117x, line 30: > Should StacksApi be StacksNode or StacksNodeApi

With almost all the endpoints being for the stacks node RPC, I'd be in favor of calling it something other than "API" because that ends up confusing a lot of people

janniks: I'll rename to `StacksNodeApi`. I'll also add methods for calling mempool-nonce estimates (via API) in a PR soon. I'll keep the behavior to fallback to node fetches if API fetches don't work.

janniks:
Any other thoughts on this change? @zone117x 
I still don't like the separation of common+transactions. It's super unclear when something should be in which.
Currently `common` houses things like `TransactionVersion`/`ChainID` which is very transactiony. 😅

Screenshot 2023-12-04 at 11 32 25
For dependents it doesn't really matter, since pretty much everything already imports from `transaction`

A) Do you think we could remove the package and move its contents to `transactions`?
B) Or maybe move most things to `transactions`, and leave `common` for very pure "helpers" (e.g. buffer, bytes, hex).

I'm thinking if most people also need to do text<>hex conversion or similar, we might as well do everything in `transactions`, so folks only need one dependency for 95% of use-cases.

zone117x:
Hmm good arguments for both options. I don't really have a strong opinion. Kinda lean towards keeping the `common` package so that other packages aren't transitively dependent on the `transaction` package for helpers that are unrelated to transactions

packages/common/src/constants.ts
```diff
@@ -1,41 +1,9 @@
...
- * For mainnet/testnet the v2/info response `.network_id` refers to the chain ID.
- * For subnets the v2/info response `.network_id` refers to the peer network ID and the chain ID (they are the same for subnets).
- * The `.parent_network_id` refers to the actual peer network ID (of the parent) in both cases.
- */
-export enum PeerNetworkID {
-  Mainnet = 0x17000000,
-  Testnet = 0xff000000,
-}
+export const HIRO_MAINNET_URL = 'https://stacks-node-api.mainnet.stacks.co';
+export const HIRO_TESTNET_URL = 'https://stacks-node-api.testnet.stacks.co';
```
rafaelcr, line 2: We should change these to `hiro.so` given that the variable name is already HIRO

janniks:
Splitting this out into separate PRs

janniks:
closed in favor of #1622

PR #1598, opened 2023-11-27
[next] feat: switch to simple string private keys

janniks:
- removes "wrapped" string private keys in favor of "flat" string/Uint8Array type aliases.
  - tldr: now we can just use string hex encoded private keys for all methods. before we often had to wrap with `createStacksPrivateKey(...)`

### DIFF
--- a/packages/transactions/src/keys.ts
+++ b/packages/transactions/src/keys.ts
@@ -14,8 +14,8 @@ import {
   hexToBytes,
   intToHex,
   parseRecoverableSignatureVrs,
-  privateKeyToBytes,
   PRIVATE_KEY_COMPRESSED_LENGTH,
+  privateKeyToBytes,
   signatureRsvToVrs,
   signatureVrsToRsv,
 } from '@stacks/common';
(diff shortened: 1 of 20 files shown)


packages/transactions/src/keys.ts
```diff
@@ -111,64 +113,62 @@ export function publicKeyFromSignatureRsv(
...
-  const compressed = data.length == PRIVATE_KEY_COMPRESSED_LENGTH;
-  return { data, compressed };
-}
-
-export function makeRandomPrivKey(): StacksPrivateKey {
-  return createStacksPrivateKey(utils.randomPrivateKey());
-}
+// todo: complete refactor
+export type PrivateKey = string | Uint8Array;
+export type PublicKey = string | Uint8Array;
```
zone117x, line 163: Do you think it's worth keeping around some backward compatibility? For example, functions that accept a key as an arg could use something like
```ts
// Allow the `StacksPrivateKey` type for backwards compat
export type PrivateKeyArg = string | Uint8Array | StacksPrivateKey;
```

Though maybe that's too much baggage to carry around

zone117x (approved):
Had a question around backwards compatibility. Seems like a good change to me 👍

packages/transactions/src/keys.ts
janniks: Interesting idea. I was going for a more breaking change (combined with some other PRs).

This backward-comp would be okay to keep, but I'm thinking if people likely need to make code changes for migrating, this one would be very simple (and likely not one a lot of folks would even run into).

_not sure_

zone117x: I don't feel strongly either way. Breaking change seems fine to me

PR #1599, opened 2023-11-27
feat: cli option to decode serialized Clarity values

zone117x:
> This PR was published to npm with the version `6.9.1-pr.b33d848.0`
> e.g. `npm install @stacks/common@6.9.1-pr.b33d848.0 --save-exact`

### Description

New CLI feature to decode Clarity values into json/repr/pretty strings.

### Example

The `repr` format string is returned by default:
```shell
$ stx decode_cv 0b00000003000000000000000000000000000000000100000000000000000000000000000000020000000000000000000000000000000003
(list 1 2 3)
```

Use `--format json` to output JSON:
```shell
$ stx decode_cv --format json 0b00000003000000000000000000000000000000000100000000000000000000000000000000020000000000000000000000000000000003
{"type":"(list 3 int)","value":[{"type":"int","value":"1"},{"type":"int","value":"2"},{"type":"int","value":"3"}]}
```

Use `--format pretty` to output a pretty-print repr string:
```shell
$ stx decode_cv --format pretty 0b00000003000000000000000000000000000000000100000000000000000000000000000000020000000000000000000000000000000003
(list
  1
  2
  3
)
```

Values can also be piped into the command by using `-` in place of the value:
```shell
$ echo 0x050011deadbeef11ababffff11deadbeef11ababffff | stx decode_cv --format json -
{"type":"principal","value":"S08XXBDYXW8TQAZZZW8XXBDYXW8TQAZZZZ88551S"}
```

Real world example of decoding output from a Stacks node RPC call:
```shell
$ curl -sX POST 'https://stacks-node-api.testnet.stacks.co/v2/map_entry/ST000000000000000000002AMW42H/pox-3/delegation-state?proof=0' -H 'Content-Type: application/json' -d '"0x0c0000000107737461636b65720515f1a77e086d9de40dfa879175c4fb22f426950c5d"' | jq -r .data | stx decode_cv --format pretty -
(some {
  amount-ustx: u10000000,
  delegated-to: 'SND3WM4FYSYFZM2C8K62K1XJD97W6YTXET0N4C6A,
  pox-addr: none,
  until-burn-ht: (some u2540805)
})
```
---

### Checklist

### DIFF
--- a/packages/cli/src/argparse.ts
+++ b/packages/cli/src/argparse.ts
@@ -424,7 +424,7 @@ export const CLI_ARGS = {
         '    {"type":"principal","value":"S08XXBDYXW8TQAZZZW8XXBDYXW8TQAZZZZ88551S"}\n' +
         '    $ echo 0x050011deadbeef11ababffff11deadbeef11ababffff | stx decode_cv -\n' +
         '    S08XXBDYXW8TQAZZZW8XXBDYXW8TQAZZZZ88551S\n',
-      group: 'Account Management',
+      group: 'Utilities',
     },
     convert_address: {
       type: 'array',
--- a/packages/cli/src/argparse.ts
+++ b/packages/cli/src/argparse.ts
@@ -422,7 +422,7 @@ export const CLI_ARGS = {
         '    S08XXBDYXW8TQAZZZW8XXBDYXW8TQAZZZZ88551S\n' +
         '    $ stx decode_cv --format json SPA2MZWV9N67TBYVWTE0PSSKMJ2F6YXW7CBE6YPW\n' +
         '    {"type":"principal","value":"S08XXBDYXW8TQAZZZW8XXBDYXW8TQAZZZZ88551S"}\n' +
-        '    $ echo SPA2MZWV9N67TBYVWTE0PSSKMJ2F6YXW7CBE6YPW | stx decode_cv -\n' +
+        '    $ echo 0x050011deadbeef11ababffff11deadbeef11ababffff | stx decode_cv -\n' +
         '    S08XXBDYXW8TQAZZZW8XXBDYXW8TQAZZZZ88551S\n',
       group: 'Account Management',
     },
--- a/packages/cli/src/argparse.ts
+++ b/packages/cli/src/argparse.ts
@@ -395,6 +395,37 @@ export const CLI_ARGS = {
         '\n',
       group: 'Account Management',
     },
+    decode_cv: {
+      type: 'array',
+      items: [
+        {
+          name: 'clarity_value',
+          type: 'string',
+          realtype: 'string',
+          pattern: '-|^(0x|0X)?[a-fA-F0-9]+$',
+        },
+        {
+          name: 'format',
+          type: 'string',
+          realtype: 'format',
+          pattern: '^(repr|pretty|json)$',
+        },
+      ],
+      minItems: 1,
+      maxItems: 4,
+      help:
+        'Decode a serialized Clarity value.\n' +
+        '\n' +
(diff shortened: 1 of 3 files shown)


packages/cli/src/cli.ts
```diff
@@ -952,6 +954,37 @@ async function readOnlyContractFunctionCall(
...
+ * args:
+ * @value (string) the hex string of the serialized value, or '-' to read from stdin
+ * @format (string) the format to output the value in; one of 'pretty', 'json', or 'repr'
+ */
+function decodeCV(_network: CLINetworkAdapter, args: string[]): Promise<string> {
+  const inputArg = args[0];
+  const format = args[1];
+
+  let inputValue: string;
+  if (inputArg === '-') {
```
janniks, line 968: Could stdin be detected from the absence of this parameter?

packages/cli/src/argparse.ts
```diff
@@ -395,6 +395,37 @@ export const CLI_ARGS = {
...
+        '\n' +
+        'Example:\n' +
+        '\n' +
+        '    $ stx decode_cv 0x050011deadbeef11ababffff11deadbeef11ababffff\n' +
+        '    S08XXBDYXW8TQAZZZW8XXBDYXW8TQAZZZZ88551S\n' +
+        '    $ stx decode_cv --format json SPA2MZWV9N67TBYVWTE0PSSKMJ2F6YXW7CBE6YPW\n' +
+        '    {"type":"principal","value":"S08XXBDYXW8TQAZZZW8XXBDYXW8TQAZZZZ88551S"}\n' +
+        '    $ echo SPA2MZWV9N67TBYVWTE0PSSKMJ2F6YXW7CBE6YPW | stx decode_cv -\n' +
+        '    S08XXBDYXW8TQAZZZW8XXBDYXW8TQAZZZZ88551S\n',
+      group: 'Account Management',
```
janniks: group, idk if there's something like Utilities

packages/cli/tests/cli.test.ts
```diff
@@ -53,6 +56,39 @@ const testnetNetwork = new CLINetworkAdapter(
   {} as CLI_NETWORK_OPTS
 );
 
+describe('decode_cv', () => {
```
zone117x, line 59: A jest test for reading from stdin (e.g. `echo 0x1234 | stx decode_cv -`)  requires some trickery with mocking stdin, which doesn't seem worth it imo. It works when testing locally.

packages/cli/src/cli.ts
zone117x: I think something like `const shouldReadFromStdin = !inputArg && !process.stdin.isTTY;` would work for that. But the command-line parsing in this lib is pretty weird and limiting with optional and positional args. Like if you just run `stx decode_cv` (no args) then it prints the help description. I think the explicit approach works better until this lib is refactored to use a reasonable command-line parser library.

packages/cli/src/argparse.ts
zone117x: Done

packages/cli/src/cli.ts
janniks: Got it, makes sense 🎉

janniks (approved):
Very useful!! 🎉

packages/cli/tests/cli.test.ts
SamiQu: > A jest test for reading from stdin (e.g. `echo 0x1234 | stx decode_cv -`)  requires some trickery with mocking stdin, which doesn't seem worth it imo. It works when testing locally.

PR #1603, opened 2023-12-10
fix: handle `NoEstimateAvailable` text error body

zone117x:
> This PR was published to npm with the version `6.10.1-pr.ccf1faa.0`
> e.g. `npm install @stacks/common@6.10.1-pr.ccf1faa.0 --save-exact`

The latest stacks-node in the next branch can have an error response as a text body (as opposed to json). Not sure if this is a regression in the stacks-node. Either way, this PR will handle the case of a `NoEstimateAvailable` text body response.


janniks (approved):
Thanks 🎈

PR #1607, opened 2023-12-16
feat: CLI add rate limit logic to gaia dump and restore

pradel:
### Description

Added some sleep logic in the CLI to wait between batches for `gaia_dump_bucket` and `gaia_restore_bucket` as some users have been hitting the rate limit when dumping their Gaia bucket as they have many files (> 200).

I did manually test the logic and with the current config not hitting the rate limit, uploading 99 files every mins does hit the rate limit so switched to 99 every 2 mins to get it to work.

#### Breaking change?

No

### Example

Create a bucket with many files and see that you can download/upload all of them without hitting the rate limit.

---

### Checklist

- [ ] Unit tested updated code paths
- [ ] Tagged 1 of @janniks or @zone117x for review


zone117x (approved):

janniks (approved):

janniks:
Thanks, 👏 Looks good

PR #1614, opened 2024-01-03
feat: add pox-4 signer-key to StackingClient methods

janniks:
> This PR was published to npm with the version `6.12.2-pr.2a2f6aa.0`
> e.g. `npm install @stacks/common@6.12.2-pr.2a2f6aa.0 --save-exact`

- pox-4 changes
  - adds signer-key to methods _(as optional arguments, so this change can be merged before pox-4 is live)_
  - removes deprecated data-vars

Waiting on
- [x] https://github.com/stacks-network/stacks-core/issues/4274
- [x] https://github.com/stacks-network/stacks-core/pull/4277


zone117x (commented):
What happens if a signer key arg is provided before pox-4 is activated?

janniks:
Good point 😉 For now I assumed folks adding a signer-key know what they're doing, currently the ABI would fail.
But it's probably better to add another throw to the 'ensure' helper for both missing and given signer-key and the respective pox.

janniks:
Added a better ensure check. 
Also waiting to see if https://github.com/stacks-network/stacks-core/issues/4230 is planned.
The API should be able to use the pre-release of this PR

friedger:
Delegate stack stx won't have signer key. There is an issue https://github.com/stacks-network/stacks-core/issues/4274

hstove:
Also FYI new function arg `signer-sig` for a few functions: https://github.com/stacks-network/stacks-core/pull/4277

janniks:
Thanks 🙏 Will wait for the PRs to merge to update the remainder

hstove:
Added a helper for the signer signature: #1628 

Now that this is in the next branch, I don't think there will be any more changes to the pox-4 function signatures.

janniks:
Added latest args (thanks to @hstove) 🙏

Will update README examples, but should be stable now -- Do we think the latest interface is final and can be merged?

hstove:
@janniks from my perspective there won't be any more changes to the function signatures that are user-facing (ie the ones exposed here). Obviously there could be changes if something comes up in the future, but nothing is planned. I think there might be a change to pox-4 that moves a function into pox-4 from elsewhere (`vote-for-aggregate-public-key`) but that won't effect the functions exposed in this library

janniks:
added a signer-sig test and refactored the helper method you added @hstove -- lmk if you dislike it or otherwise want changes.

hstove:
Perfect, I was actually going to ask if you could provide a helper to return the message and domain, but you did!

packages/stacking/README.md
```diff
@@ -94,13 +94,28 @@ const privateKey = 'd48f215481c16cbe6426f8e557df9b78895661971d71735126545abddcd5
 // block height at which to stack
 const burnBlockHeight = 2000;
 
+// signer key
+const signerPrivateKey = randomPrivKey();
+const signerKey = getPublicKey(signerPrivateKey);
+
+const signerSignature = signPox4SignatureHash({
```
zone117x: What do you think about making it so that the signature is generated by the `client.stack(...)` family of functions rather than requiring this separate step?

hstove: I think it's better to have both options. Some signers will generate a hex signature from an out-of-band process, and in many/some cases that's better for security

zone117x: Good point, having both options makes sense to me

janniks: Yeah, been discussing this w/ @friedger as well.

- 1: One thought was adding a `stackingPrivateKey` to the client constructor, which then allows handling everything key/sig wise for the developer.
- 2: Or going even further with out-of-band signing, adding some sort of hook/plugin system (callback style) which can be provided by the user to even allow hardware device signing without exposure of a private key to JS code. 

Both get complicated fast when considering allowing overloads for different needs. I kept this base case for now, but would like to add option 1) to the next refactor. Option 2) is probably a lot of additional work and would touch other packages for adding these hooks everywhere, but should also be explored further. ✨

packages/stacking/README.md
```diff
@@ -94,13 +94,27 @@ const privateKey = 'd48f215481c16cbe6426f8e557df9b78895661971d71735126545abddcd5
...
 const burnBlockHeight = 2000;
 
+// signer key
+const signerPrivateKey = randomPrivKey();
+const signerKey = getPublicKeyFromPrivate(signerPrivateKey.data);
+
 // Refer to initialization section to create client instance
+const signerSignature = client.signPoxSignature({
+  topic: 'stack-stx',
+  rewardCycle: 132, // current reward cycle
```
zone117x: Should we include a snippet of code around here for the best way to grab the reward cycle?

packages/stacking/README.md
```diff
@@ -94,13 +94,27 @@ const privateKey = 'd48f215481c16cbe6426f8e557df9b78895661971d71735126545abddcd5
 // block height at which to stack
 const burnBlockHeight = 2000;
 
+// signer key
+const signerPrivateKey = randomPrivKey();
```
zone117x: Should be
```ts
import { makeRandomPrivKey } = from '@stacks/transactions';

...

const signerPrivateKey = makeRandomPrivKey();
```

packages/stacking/README.md
```diff
@@ -94,13 +94,27 @@ const privateKey = 'd48f215481c16cbe6426f8e557df9b78895661971d71735126545abddcd5
...
+const signerKey = getPublicKeyFromPrivate(signerPrivateKey.data);
+
 // Refer to initialization section to create client instance
+const signerSignature = client.signPoxSignature({
+  topic: 'stack-stx',
```
zone117x, line 103: Can this be an imported const or enum rather than a magic string?

packages/stacking/README.md
janniks: Yeah, this one is weird. It should be the current reward-cycle, if I read the pox-4 correctly. We could also default to current and do a fetch call to get the current (requires async, that's why I left it out for now).

packages/stacking/README.md
janniks: Already an enum as `Pox4SignatureTopic`, I made ts stringify the enum for better IDE support. I'll change it to a union of both maybe for discoverability. 
In the docs you would prefer the enum import?

packages/stacking/README.md
hstove: Yes, for `stack-stx` and `stack-extend` it needs to be the current reward cycle. For `stack-aggregation-commit`, because this can be called in future cycles, the `rewardCycle` needs to be equal to the `reward-cycle` param used in `stack-aggregation-commit`.

janniks:
Will wait for https://github.com/stacks-network/stacks-core/pull/4377 to merge until updating the interface, but probably safe to say `signer-sig` will be an optional.

hstove:
@janniks that was just merged. Sorry for saying "i don't think it will change" and then changing it 😅

janniks:
Oh nice 🙏🏻 no no, all good ^^

hstove:
Sharing for your visibility: potential new changes to pox-4 that would require updates here: https://github.com/stacks-network/stacks-core/discussions/4424

packages/stacking/src/index.ts
```diff
@@ -1007,7 +1007,7 @@ export class StackingClient {
       uintCV(cycles),
     ] as ClarityValue[];
 
-    if (signerSignature) functionArgs.push(someCV(bufferCV(hexToBytes(signerSignature))));
+    if (signerSignature) functionArgs.push(someCV(someCV(bufferCV(hexToBytes(signerSignature)))));
```
zone117x: some(some(..))

janniks: fixed, sorry

hstove:
Made some changes to pox-4 stuff in https://github.com/hirosystems/stacks.js/pull/1647, based on https://github.com/stacks-network/stacks-core/pull/4461 (which is not merged)

janniks:
Changed some minor missing checks/params and addressed review comments. Could I get a re-review on this? @zone117x or @hstove

zone117x (approved):
LGTM, nice job @janniks!

hstove (approved):

hstove:
This is great, and I've been testing with a local version of Lockstacks and it works as expected.

One change request, though: can you change `authId` to be `IntegerType`? We could run into issues with the `number` type because it can technically be any u128.

janniks:
Done ✅ Thanks

PR #1615, opened 2024-01-03
chore: update mempool rejection reasons

bestmike007:
### Description

Update mempool rejection reasons according to https://github.com/stacks-network/stacks-core/blob/2.4.0.0.4/src/chainstate/stacks/db/blocks.rs#L229-L330

#### Breaking change?

No.

### Example

---

### Checklist

- [ ] Unit tested updated code paths


bestmike007:
@zone117x

janniks (approved):
Cool, thanks for adding

PR #1622, opened 2024-02-05
[1] feat!: stacksjs refactor

janniks:
### EPIC 🏰 _Breaking Refactoring Stacks.js_

**Goal:** Make network/networking more obvious and less confusing for developers. While the default code will stay largely the same for most users, it should become more clear what node URL, and fetch implementation is used when.

291331642-81426bac-a6b0-4273-8c37-f92f4c8d8db6

  - Move to a network OBJECT approach (similar to bitcoin-js, micro-btc-signer, and others). The network will not be an instantiated class, but a static exported object with all information that might be relevant (txVersion, magicBytes, etc.)
  - Remove `fetch` functions from network and introduce `StacksNodeApi`, which can derive it's URL from a network, but should be treated separately -- more closely mirroring to what's actually happening in the background. Can be seen as a maintained API client, with post-processing of responses.
  - The majority of users doesn't realize network is also setting changes for "networking"
    - Most users don't customize fetchFn or URL. For users hosting their own node the only change becomes `{ api: { url: "my-node.com" }, ...}` in the params of most relevant tx functions.

Supersedes https://github.com/hirosystems/stacks.js/pull/1596

**@ Reviewers:**  ✨
Please feel free to leave any feedback on these PRs. Going over the code will show what it's like to code in the new style. So any ideas and feedback (especially negative) is welcome 🙏

---

This PR:
- removes duplicate definitions for next steps
- refactors common package

--- 

> Split into multiple PRs for easier reviewing

`next`
    ↑
`feat/next-cleanup-common-files` this PR 🟢
    ↑
`feat/next-add-new-network` #1623 
    ↑
`feat/next-add-api-package` #1624 
    ↑
`feat/next-update-api-stacking` #1625
    ↑
`feat/next-update-cli` #1626

### DIFF
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -1,0 +1,2 @@
+/** Hex-encoded string (without a 0x prefix) */
+export type Hex = string; // todo: should prefix always be allowed?
--- a/packages/common/src/config.ts
+++ b/packages/common/src/config.ts
@@ -1,10 +1,8 @@
-// import { network } from './network'
-
 /**
  * @ignore
+ * todo: remove this file (needed for CLI?)
  */
 const config = {
-  // network: network.defaults.MAINNET_DEFAULT,
   network: {
     layer1: 'placeholder',
   },
--- a/packages/common/src/constants.ts
+++ b/packages/common/src/constants.ts
@@ -1,41 +1,9 @@
-/**
- * The **chain** ID.
- * Is used for signing, so transactions can't be replayed on other chains.
- */
-export enum ChainID {
-  Testnet = 0x80000000,
-  Mainnet = 0x00000001,
-}
-
-/**
- * The **transaction** version.
- * Is used for signing, so transactions can't be replayed on other networks.
- */
-export enum TransactionVersion {
-  Mainnet = 0x00,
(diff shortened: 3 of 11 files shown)


hugoclrd:
@janniks 
So this PR is a small set of changes and is mostly just refactoring with almost no impact for developers right?

janniks:
Yes mostly small refactoring, however there are breaking changes with many consequences in the chain of PRs. I tried to split them up for easier review and following the refactor. Feel free to take a look at the other PRs mentioned in the description, some will be more breaking than others, but follow the same narrative.

packages/common/src/types.ts
```diff
@@ -0,0 +1,2 @@
+/** Hex-encoded string (without a 0x prefix) */
+export type Hex = string; // todo: should prefix always be allowed?
```
hugoclrd, line 2: The everlasting question

hugoclrd (approved):
LGTM 👍

packages/common/src/types.ts
janniks: lol -- I'm starting to consider it more and more, one more if shouldn't hurt perf 😅

PR #1623, opened 2024-02-05
feat!: update to new network object

janniks:
Part of #1622 

This PR:
- adds the new "static" network (keeping the network package, most users won't need to import, because they can use string literals)
- adds new `.api` options to allow separately from network defining a `.url` and `.fetch` to use
- moves "fetch" functions from `tx/builders.ts` to `tx/fetch.ts`

--- 

> Split into multiple PRs for easier reviewing

`next`
    ↑
`feat/next-cleanup-common-files` #1622
    ↑
`feat/next-add-new-network` this PR 🟢
    ↑
`feat/next-add-api-package` #1624 
    ↑
`feat/next-update-api-stacking` #1625
    ↑
`feat/next-update-cli` #1626

### DIFF
--- a/packages/transactions/src/fetch.ts
+++ b/packages/transactions/src/fetch.ts
@@ -1,0 +1,383 @@
+import {
+  ApiParam,
+  bytesToHex,
+  createFetchFn,
+  defaultApiLike,
+  validateHash256,
+  with0x,
+} from '@stacks/common';
+import { estimateTransactionByteLength } from './builders';
+import { ClarityValue, NoneCV, deserializeCV, serializeCV } from './clarity';
+import { NoEstimateAvailableError } from './errors';
+import { serializePayload } from './payload';
+import { StacksTransaction, deriveNetworkFromTx } from './transaction';
+import { cvToHex, defaultApiFromNetwork, parseReadOnlyResponse } from './utils';
+
+export const BROADCAST_PATH = '/v2/transactions';
+export const TRANSFER_FEE_ESTIMATE_PATH = '/v2/fees/transfer';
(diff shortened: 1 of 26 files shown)


packages/transactions/src/fetch.ts
```diff
@@ -0,0 +1,382 @@
...
+  // todo: could derive the network from the address and use as default if no apiOd
+
+  const api = defaultApiLike(apiOpt);
+  const url = `${api.url}${ACCOUNT_PATH}/${address}?proof=0`;
+  const response = await api.fetch(url);
+
+  if (!response.ok) {
+    const msg = await response.text().catch(() => '');
+    throw new Error(
+      `Error fetching nonce. Response ${response.status}: ${response.statusText}. Attempted to fetch ${url} and failed with the message: "${msg}"`
```
MicaiahReid, line 104: It would feel a bit clunky as a user to get `failed with the message: ""` in the case that fetching the msg throws. Could we throw a different error if the msg is not available?

MicaiahReid (changes requested):
I just had one comment, but there were a few occurrences of it. Great work!

packages/transactions/src/fetch.ts
janniks: Nice thanks -- yeah not super pretty. This is the behavior live right now, this method got moved. But I'll open a new issue for it and address in a new PR if that's ok?

MicaiahReid: Yep, works for me!

MicaiahReid (approved):

rafaelcr (approved):
Nice!

PR #1628, opened 2024-02-08
feat: helper function to make pox-4 signer signature

hstove:
A PR was just merged into the next branch of Stacks Core that includes a new parameter `signer-sig`. This is a signature that proves control/approval of the provided signer key. This PR adds a little helper function to generate the proper signed structured data that is needed for this parameter.


janniks (approved):
Awesome, thanks for this!! Looks good

PR #1630, opened 2024-02-19
fix: add stacks transaction memo equality by removing trailing empty …

### LINKED ISSUE #1575: Transaction not identical after serialize/deserialize

4.3.8

### Describe the bug

Serializing and the deserializing a transaction does not quite result in the same transaction. For example, this following `vitest` unit test will fail:

```typescript
test('StacksTransaction serialize/deserialize', async () => {
  const options = { /* ... */ };
  const tx = await makeUnsignedSTXTokenTransfer(options);
  const tx_encoded = tx.serialize();
  const tx_decoded = deserializeTransaction(tx_encoded);

  expect(tx_decoded).toEqual(tx);
})
```

The reason it fails is that the `memo.content` field will have a large array of null bytes:

```
content: '\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

When it should be an empty string:

```
content: ''
```

#### How to reproduce

See example unit test above

#### Expected behavior

`memo.content` should deserialize to en empty string, and the deserialized `StacksTransaction` should deep equal the initial `StacksTransaction`

### Additional context

This is using NodeJS 20.7.0

janniks:
> This PR was published to npm with the version `6.11.4-pr.0cd4885.0`
> e.g. `npm install @stacks/common@6.11.4-pr.0cd4885.0 --save-exact`

- fixes https://github.com/hirosystems/stacks.js/issues/1575

thx for the find @jbencin 🙏

### DIFF
--- a/packages/transactions/src/types.ts
+++ b/packages/transactions/src/types.ts
@@ -274,7 +274,7 @@ export function serializeMemoString(memoString: MemoString): Uint8Array {
 
 export function deserializeMemoString(bytesReader: BytesReader): MemoString {
   let content = bytesToUtf8(bytesReader.readBytes(MEMO_MAX_LENGTH_BYTES));
-  content = content.replace(/\u0000*$/, ''); // remove all trailing unicode empty characters
+  content = content.replace(/\u0000*$/, ''); // remove all trailing unicode null characters
   return { type: StacksMessageType.MemoString, content };
 }
 
--- a/packages/transactions/src/types.ts
+++ b/packages/transactions/src/types.ts
@@ -273,7 +273,8 @@ export function serializeMemoString(memoString: MemoString): Uint8Array {
 }
 
 export function deserializeMemoString(bytesReader: BytesReader): MemoString {
-  const content = bytesToUtf8(bytesReader.readBytes(MEMO_MAX_LENGTH_BYTES));
+  let content = bytesToUtf8(bytesReader.readBytes(MEMO_MAX_LENGTH_BYTES));
+  content = content.replace(/\u0000*$/, ''); // remove all trailing unicode empty characters
   return { type: StacksMessageType.MemoString, content };
 }
 
--- a/packages/transactions/tests/builder.test.ts
+++ b/packages/transactions/tests/builder.test.ts
@@ -1,24 +1,33 @@
-import { bytesToHex, utf8ToBytes } from '@stacks/common';
+import { bytesToHex, bytesToUtf8, hexToBytes, utf8ToBytes } from '@stacks/common';
 import {
-  createApiKeyMiddleware,
-  createFetchFn,
   StacksMainnet,
   StacksTestnet,
+  createApiKeyMiddleware,
+  createFetchFn,
 } from '@stacks/network';
 import * as fs from 'fs';
 import fetchMock from 'jest-fetch-mock';
 import {
+  createFungiblePostCondition,
+  createSTXPostCondition,
+  serializePostCondition,
+} from '../src';
+import {
+  MultiSigSpendingCondition,
+  SingleSigSpendingCondition,
+  SponsoredAuthorization,
+  StandardAuthorization,
   createSingleSigSpendingCondition,
   createSponsoredAuth,
   emptyMessageSignature,
   isSingleSig,
-  MultiSigSpendingCondition,
   nextSignature,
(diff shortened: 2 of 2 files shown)


packages/transactions/src/types.ts
```diff
@@ -273,7 +273,8 @@ export function serializeMemoString(memoString: MemoString): Uint8Array {
 }
 
 export function deserializeMemoString(bytesReader: BytesReader): MemoString {
-  const content = bytesToUtf8(bytesReader.readBytes(MEMO_MAX_LENGTH_BYTES));
+  let content = bytesToUtf8(bytesReader.readBytes(MEMO_MAX_LENGTH_BYTES));
+  content = content.replace(/\u0000*$/, ''); // remove all trailing unicode empty characters
```
zone117x: I'm curious if this always trims as expected. I think it depends on how `bytesToUtf8` handles invalid utf8 bytes. For example, I know nodejs encoders typically output the "unicode replacement character" `U+FFFD`, while some web browser implementations ignore the invalid bytes in their output.

In other words, it seems possible to construct a tx with a memo that has equality with another memo even with different bytes.

It might be helpful to add test vectors with 
A) Last character contains a trailing zero byte (`0x00`)
B) Trailing bytes are invalid utf8

janniks: Yeah, hard to say what the best behavior is here. I'm thinking it would be fine to leave as is, but some users might not expect the right-padded zero bytes from a serialized transaction. So all trailing zero-bytes are trimmed (comment might have been confusing). `\u0000` is ASCII/UTF-8 compatible `0` byte.
B) is interesting.
Added more test-cases.

But essentially, it's always possible to break this -- if this behavior is too unclear, we can also leave as is and let the developer deal with `.trim()`

zone117x: I think this is a useful PR -- establish the expected behavior. I think it does a decent job at that as-is. But there are possibly edge cases where the same or very similar memos can decode to things like
`1234�`
`1234`
`1234����������`
`1234\0\0\0\0\0\0\0\0\0\0\0\0\0\0`

I think the above test vectors I mentioned would clarify the expected output.

janniks: Yeah, these are tricky to represent / test in JS, since TextEncoder already replaces invalid byte-sequences. But I added some similar test-cases. I think it's ok to work under the assumption that null characters shouldn't be used, and if someone needs them, they can use their own deserialization. Current utf8 behavior is okay for me as well, and right-stripping null characters. 
I'm fine with this being a "bug fix", but we can also treat it as a breaking change. Or keep the status quo (and not have 1 to 1 de-/serializable tx in JS).
cc @jbencin

zone117x: Test cases LGTM

zone117x (approved):

packages/transactions/src/types.ts
jbencin: Where are these null characters coming from? Rather than removing them, is it possible not to introduce them in the first place? I think if the user intentionally puts null characters in this field, they should be preserved. There are legitimate use-cases for this. For example, maybe the user wants to put binary data in the field, or maybe they have a list of strings and are using `\0` as a separator

janniks: That won't be possible, while adding this feature.

The null characters come from the serialization on the blockchain (wire format). The memo has a fixed length of bytes, and is right-padded if the actual content is shorter.

I'll double check if this is true in the impl, or if there's a way to strip the trailing bytes, hard to follow the macro magic 😅

janniks: I can't find anything on `memo` in the SIPs, but core uses `read_exact` via `impl_byte_array_message_codec` for it, on a fixed byte length of 34. So I'm assuming this was always how it worked.

jbencin: So if it's a fixed-length field, we'd still submitting the tx with the null bytes then? Maybe it is better to just make that explicit to the user, I don't know

zone117x: Exchanges pretty much all use null terminated strings for memo, so this PR makes sense

jbencin (approved):

PR #1631, opened 2024-02-19
feat: sort tuple properties in cl pretty print

hugoclrd:
> This PR was published to npm with the version `6.11.4-pr.3d18e85.0`
> e.g. `npm install @stacks/common@6.11.4-pr.3d18e85.0 --save-exact`

### Description

Change `Cl.prettyPrint` so that it sorts properties in a tuple.
This an opinionated change and I could do it in Clarinet, but it kind of make sense to have it natively here as well. I would understand if one disagrees though.

#### Breaking change?

Not really, in the sense that developers shouldn't rely on the property output of this method, this is mostly for debugging

### Example

`Cl.prettyPrint` is used in the clarinet sdk to show output diffs in tests.
Because the output is just a string, Vitest can fail to properly render the diff of the properties of the expected tuple vs the actual tuple are not in the same order

See below where only the value of `a` changes.
```
- Expected
+ Received

  (some {
-   d: u1010000000,
-   a: u1,
+   b: u1013,
    c: u75,
-   b: u1013
+   a: u0,
+   d: u1010000000
  })
```

With this PR, it would display

```
- Expected
+ Received

  (some {
-   a: u1,
+   a: u0,
    b: u1013,
    c: u75,
    d: u1010000000
  })
```

Clarinet already does this sorting in `expectTuple`, but this is actually needed for every composite types (some, ok, err, list, tuple) and with nested values. It could be done at the clarinet level but Cl.prettyPrint already handle the recursive nature of clarity values.

---

### Checklist

- [x] Tagged 1 of @janniks for review

### DIFF
--- a/packages/transactions/src/clarity/prettyPrint.ts
+++ b/packages/transactions/src/clarity/prettyPrint.ts
@@ -66,9 +66,7 @@ function formatTuple(cv: TupleCV, space: number, depth = 1): string {
   const spaceBefore = formatSpace(space, depth, false);
   const endSpace = formatSpace(space, depth, true);
 
-  return `{${spaceBefore}${items
-    .sort((a, b) => a[0].localeCompare(b[0]))
-    .join(`,${spaceBefore}`)}${endSpace}}`;
+  return `{${spaceBefore}${items.sort().join(`,${spaceBefore}`)}${endSpace}}`;
 }
 
 function exhaustiveCheck(param: never): never {
--- a/packages/transactions/src/clarity/prettyPrint.ts
+++ b/packages/transactions/src/clarity/prettyPrint.ts
@@ -41,14 +41,16 @@ function formatList(cv: ListCV, space: number, depth = 1): string {
 
 /**
  * @description format Tuple clarity values in clarity style strings
+ * the keys are alphabetically sorted
  * with the ability to prettify the result with line break end space indentation
  * @example
  * ```ts
- * formatTuple(Cl.tuple({ id: Cl.uint(1) }))
- * // { id: u1 }
+ * formatTuple(Cl.tuple({ id: Cl.uint(1), age: Cl.uint(20) }))
+ * // { age: 20, id: u1 }
  *
- * formatTuple(Cl.tuple({ id: Cl.uint(1) }, 2))
+ * formatTuple(Cl.tuple({ id: Cl.uint(1), age: Cl.uint(20) }, 2))
  * // {
+ * //   age: 20,
  * //   id: u1
  * // }
  * ```
@@ -64,7 +66,9 @@ function formatTuple(cv: TupleCV, space: number, depth = 1): string {
   const spaceBefore = formatSpace(space, depth, false);
   const endSpace = formatSpace(space, depth, true);
 
-  return `{${spaceBefore}${items.join(`,${spaceBefore}`)}${endSpace}}`;
+  return `{${spaceBefore}${items
+    .sort((a, b) => a[0].localeCompare(b[0]))
+    .join(`,${spaceBefore}`)}${endSpace}}`;
 }
 
 function exhaustiveCheck(param: never): never {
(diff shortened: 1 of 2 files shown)


packages/transactions/src/clarity/prettyPrint.ts
```diff
@@ -64,7 +66,9 @@ function formatTuple(cv: TupleCV, space: number, depth = 1): string {
   const spaceBefore = formatSpace(space, depth, false);
   const endSpace = formatSpace(space, depth, true);
 
-  return `{${spaceBefore}${items.join(`,${spaceBefore}`)}${endSpace}}`;
+  return `{${spaceBefore}${items
+    .sort((a, b) => a[0].localeCompare(b[0]))
```
zone117x: Are you confident that `localCompare` will always perform the correct lexicographical ordering? I think when it's called without a locale like this, then it defaults to the environment's locale (e.g. the user's browser locale settings)

hugoclrd: I used it because it enable case-insensitive string sorting. 
But indeed it might be more consistent if I enforce the locale to "en", if we consider that other different locales could result in different sorting?

Also, what really matters here in my opinion is that to identical input result in the same output
(eg: `{b: 2, a: 1}` and `{a: 1, b: 2}` should both give the same output)

hugoclrd: So I'm also happy with just `sort()`, developrs should mix upper/lower case in the tuples anyway

zone117x: Does core do case-insensitive sorting? I thought it was purely lexicographical. Either way, I think this is the typical way to do that in js:
```js
.sort((a, b) => a.toLowerCase() > b.toLowerCase() ? 1 : (a.toLowerCase() < b.toLowerCase() ? -1 : 0));
```

janniks:
I'm fine with sorting. For use in tests it's great. Maybe let's add a note to not use the function for keeping keys in a specific order. (I think in some cases key order can have consequences, but less likely users will use this for it)

packages/transactions/src/clarity/prettyPrint.ts
janniks: How about `a.localeCompare(b, 'en')`? We can set the locale to be more deterministic (with an optional ` { sensitivity: 'base' }` for case-insensitive.
But is the `[0]` correct here, or just comparing the first character?

hugoclrd: > Does core do case-insensitive sorting?

What do you mean by `core`?

If the snippet you shared and `localeCompare` return the same result, shouldn't we prefer `localeCompare`?

Replying to @janniks comment as well:
> maybe let's add a note to not use the function for keeping keys in a specific order

Or I can enforce the locale to `en` and the sorting will always be the same

zone117x: > What do you mean by core?

The clarity runtime(s)

> If the snippet you shared and localeCompare return the same result, shouldn't we prefer localeCompare?

It seems like localCompare could return different results depending on host environment. Also AFAIK there's not a guarantee that every environment will have every locale (maybe `en` is an exception?), I've seen slim nodejs environments with seemingly no locale installed.


The snippet I shared uses unicode scalar value comparisons which is essentially the same as lexicographical comparison in this case, I think. If you are confident about the behavior of `localCompare` then feel free to use it :)

hugoclrd: > Does core do case-insensitive sorting?

In clarity the order of the tuple keys are not sorted at all

>  If you are confident about the behavior of localCompare

Well I was but now I'm not (with " slim nodejs environments with seemingly no locale installed".

I'm tempted to go with just `items.sort()` now. If that was the initial implementation, would it have been fine for both of you?

janniks: Keys are always strings in Clarity, right?

hugoclrd: @janniks yes, and better than that, it must start with `[a-zA-Z]`

hugoclrd:
Comment addressed @janniks @zone117x 
Thx for the feedback!

hugoclrd:
If this can be included in the next release of stacks.js, I'd like to include it in the next release of clarinet

zone117x (approved):
LGTM!

PR #1647, opened 2024-03-05
feat: add max-amount and auth-id params

hstove:
In https://github.com/stacks-network/stacks-core/pull/4461 we made a few changes to pox-4:

- A new `max-amount` and `auth-id` param to all the functions that take a signer key/sig
- Added `signer-key` and all related params to `stack-increase`

This PR implements those changes


janniks (approved):
Awesome, thanks!!

PR #1648, opened 2024-03-06
feat: add fixture test for pox-4 message hash

hstove:
I meant to add this in my other PR, but I made a [fixture test in Rust](https://github.com/stacks-network/stacks-core/blob/c9d8f8d66bea0b8983ae2a92fd9716be9dc0c4df/stackslib/src/util_lib/signed_structured_data.rs#L380) so that we can have one here as well


janniks (approved):
👏 thx

PR #1649, opened 2024-03-07
docs: add migration docs

janniks:
[next]

Adds content to the MIGRATION.md document in the repo.

### DIFF
--- a/.github/MIGRATION.md
+++ b/.github/MIGRATION.md
@@ -1,7 +1,13 @@
 # Migration Guides
 
-- [Stacks.js (\<=4.x.x) → (5.x.x)](#stacksjs-4xx--5xx)
+- [Stacks.js (\>=5.x.x) → (7.x.x)](#stacksjs-5xx--7xx)
   - [Breaking Changes](#breaking-changes)
+  - [Stacks Network](#stacks-network)
+  - [StacksNodeApi](#stacksnodeapi)
+  - [StacksNetwork to StacksNodeApi](#stacksnetwork-to-stacksnodeapi)
+  - [Clarity Representation](#clarity-representation)
+- [Stacks.js (\<=4.x.x) → (5.x.x)](#stacksjs-4xx--5xx)
+  - [Breaking Changes](#breaking-changes-1)
     - [Buffer to Uint8Array](#buffer-to-uint8array)
     - [Message Signing Prefix](#message-signing-prefix)
 - [blockstack.js → Stacks.js (1.x.x)](#blockstackjs--stacksjs-1xx)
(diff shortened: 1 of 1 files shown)


.github/MIGRATION.md
```diff
@@ -14,6 +20,131 @@
     - [Using blockstack.js](#using-blockstackjs-2)
     - [Using @stacks/encryption or @stacks/auth](#using-stacksencryption-or-stacksauth)
 
+## Stacks.js (&gt;=5.x.x) → (7.x.x)
+
+### Breaking Changes
+
+- The `@stacks/network` `new StacksNetwork()` objects were removed. Instead `@stacks/network` now exports objects like `STACKS_MAINNET`, which are static (and shouldn't be changed for most use-cases). [Read more...](#stacks-network)
```
hugoclrd: I'd would enumerate the objects here (otherwise the plural makes it hard to understand)

.github/MIGRATION.md
```diff
@@ -14,6 +20,131 @@
...
+### Stacks Network
+
+From now on "network" objects are static (aka constants) and don't require instantiation.
+
+The `@stacks/network` package exports the following network objects:
+
+- `STACKS_MAINNET`
+- `STACKS_TESTNET`
+- `STACKS_MOCKNET`
+- `STACKS_DEVNET` (alias for `STACKS_MOCKNET`)
```
hugoclrd:
```suggestion
- `STACKS_DEVNET`
- `STACKS_MOCKNET` (alias for `STACKS_DEVNET`)
```

We try to normalize the usage of devnet (and simnet), could this be presented as if `DEVNET` was the "real one" (even though it's not really the case)

.github/MIGRATION.md
```diff
@@ -14,6 +20,131 @@
...
+
+```ts
+import { STACKS_MAINNET } from '@stacks/network';
+import { STACKS_TESTNET } from '@stacks/network';
+import { STACKS_MOCKNET } from '@stacks/network';
+import { STACKS_DEVNET } from '@stacks/network';
```
hugoclrd:
```suggestion
import { STACKS_MAINNET } from '@stacks/network';
import { STACKS_TESTNET } from '@stacks/network';
import { STACKS_DEVNET } from '@stacks/network';
```

Encore `devnet`

hugoclrd (commented):
Lot's of todos, are they for later?

.github/MIGRATION.md
```diff
@@ -14,6 +20,131 @@
...
+
+### Clarity Representation
+
+The `ClarityType` enum was replaced by a readable version.
+The previous (wire format compatible) enum is still available as `ClarityWireType`.
+These types are considered somewhat internal and shouldn't cause breaking changes for most use-cases.
+
+The property holding the value of the data type is now called `value` in all cases.
+Previously, there was a mix of `value`, `list`, `buffer` etc.
+For `bigint` values, the type of the `value` property is a now `string`, for better serialization compatibility.
```
hugoclrd, line 127: Is it to avoid this
```
// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json
// @ts-ignore
BigInt.prototype.toJSON = function () {
  return this.toString();
};
```

Are string better than bigint (apart from the bigint serialization issue?)

.github/MIGRATION.md
janniks: ~Happy to match it to what clarinet uses. I can leave our MOCKNET or rename to SIMNET -- not sure I follow completely.~ Saw the next comment ^^

.github/MIGRATION.md
janniks: 👍 will do

.github/MIGRATION.md
janniks: Yeah, it's not a really solution. Mostly for wallets, so they can have their inputs be JSON compatible (without interfering with the BigInt prototype). Not super clean, but we are trying to avoid helpers/etc. and simply stringifying seems to be fairly accepted.

janniks: Other than serialization, no there's no downside of bigints that I can think of. Although some environments may not have support for it (not many left, mostly older bundling setups)

hugoclrd: Ok ok I see. I'm marking this a resolved

hugoclrd (approved):
LGTM except for the remaining comment (and the fact that there are still many todos)

.github/MIGRATION.md
janniks: Added them (not mocknet, to discourage use) 😉

janniks:
As for todos, yes there will be follow up work here. I'll rename the Api helper to Client, and some other stuff. Happy to address sooner if you like 😉

> In general, I usually treat `todo` comments like notes. Everyone that comes across them can decide to fix them or decide not to. They can be like accepted debt. If it's critical, I'll open an issue or note elsewhere. I do like the minimal burden of noting down a comment in code when already there. If there's more overhead (adding an issue, etc. tagging issue number, idk) from experience, people are discouraged from leaving their thoughts. Many times I've come back to a codebase and learned something I had completely forgotten from my own todo comments 😅
> Maybe doesn't apply to this PR though 🤷‍♂

PR #1650, opened 2024-03-08
Get pubkey from structured data

whoabuddy:
### Description

When setting up the `sign-message` script for [aibtcdev/agent-tools-ts](https://github.com/aibtcdev/agent-tools-ts/blob/main/src/wallet/sign-message.ts), we ran into an issue where we could not pass the type `StructuredDataSignature` to either of the `getPublicKeyFromSignature` methods.

After experimenting with this it was working in our initial tests, so we added the updated code back with a test to confirm it.

#### Breaking change?

More like an *unblocking* change :wink: 

### Example

You can now pass either `MessageSignature | StructuredDataSignature` to the getPublicKeyFromSignature method, detailed example is in the test file.

---

### Checklist

- [x] Tagging @janniks based on our discussion in Discord

### DIFF
--- a/packages/transactions/src/keys.ts
+++ b/packages/transactions/src/keys.ts
@@ -38,6 +38,7 @@ import {
   UNCOMPRESSED_PUBKEY_LENGTH_BYTES,
 } from './constants';
 import { hash160, hashP2PKH } from './utils';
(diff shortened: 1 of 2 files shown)


janniks (approved):
♥️

PR #1658, opened 2024-03-13
build: fix circular dep issue

janniks:
> This PR was published to npm with the version `6.12.1-pr.5ca1638.0`
> e.g. `npm install @stacks/common@6.12.1-pr.5ca1638.0 --save-exact`

This PR is needed to fix a circular dependency issue.
In the future I would like to move all the message-type definitions to this files, but it doesn't fit into the priorities right now :(

### DIFF
--- a/packages/transactions/src/message-types.ts
+++ b/packages/transactions/src/message-types.ts
@@ -1,0 +1,8 @@
+// todo: this file should hold the type definitions for more message types later
+//       needed now to fix a circular dependency issue in structuredDataSignature
+import { StacksMessageType } from './constants';
+
+export interface StructuredDataSignature {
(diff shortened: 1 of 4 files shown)


packages/transactions/src/message-types.ts
```diff
@@ -0,0 +1,8 @@
+// todo: this file should hold the type definitions for more message types later
+//       needed now to fix a circular dependency issue in structuredDataSignature
```
MicaiahReid:
```suggestion
// needed now to fix a circular dependency issue in structuredDataSignature
```

MicaiahReid (changes requested):
***UNNACCEPTABLE*** comment formatting

MicaiahReid (approved):

PR #1665, opened 2024-04-02
fix: update buffer max size to correct 1MB limit

### LINKED ISSUE #1664: new bufferCV bug
latest

### Describe the bug
In bufferCV.ts, `buffer.length > 1_000_000` should be `buffer.length > 1_048_576`.

#### How to reproduce
Let buffer contains (1_000_000, 1_048_576) characters, bufferCV will trigger error.

#### Expected behavior
Pls fix it.

janniks:
- fixes https://github.com/hirosystems/stacks.js/issues/1664

we were incorrectly calculating 1MB before
ref: stacks-core: https://github.com/stacks-network/stacks-core/blob/c50a93088d7c0261f1dbe31ab24b95028a038447/clarity/src/vm/types/mod.rs#L47


rafaelcr (approved):

bestmike007:
Please release this asap. I need to apply this patch in order to decode the function args from https://explorer.hiro.so/txid/0xf957bf074f78376d13153d1cfde7d323d9416d91f698b5d07530484d416852d1?chain=mainnet

PR #1669, opened 2024-04-10
fix: add signer opts to agg increase

janniks:
> This PR was published to npm with the version `6.13.1`
> e.g. `npm install @stacks/common@6.13.1 --save-exact`

- missing opts since last pox-4 update


zone117x (approved):
`pox-4 support completed-final-final(2)`

PR #1676, opened 2024-04-19
feat: add verify signer sig

### LINKED ISSUE #1673: Feature request: stacking helper function for `verify-signer-key-sig`
### Problem

When a client wants to verify a Stacking payload, they can validate the signature by re-generating the PoX-4 message hash. This works for users that include a `signer-signature`, but that param is optional. When the signature is omitted, the pox-4 contract checks the `signer-key-authorization` map to ensure that an authorization was previously set.

There's a helper read-only [function in pox-4 `verify-signer-key-sig`](https://github.com/stacks-network/stacks-core/blob/develop/stackslib/src/chainstate/stacks/boot/pox-4.clar#L735) which essentially does both of these things (either verify the signature or verify that an authorization exists). It would be nice to expose this read-only function in the Stacking library, so that clients can validate a payload before submitting a transaction.

### Solution

Add a `verifySignerKey()` (or similarly named) function that calls the read-only function [`verify-signer-key-sig`](https://github.com/stacks-network/stacks-core/blob/develop/stackslib/src/chainstate/stacks/boot/pox-4.clar#L735).

janniks:

- fixes #1673


zone117x (approved):
Nice tests 💯

PR #1679, opened 2024-04-19
fix: update encodeClarityValue to parse buffers as hex

### LINKED ISSUE #1674: Fix ABI buffer encoding
I remember we've discussed this before, but I can't find or remember the answer. When calling a function, the explorer processes buff argument inputs as ascii instead of as hex. Is there any way to make it process the input as hex bytes?

Actually I think the problem is in stacks.js https://github.com/hirosystems/stacks.js/blob/f39637c924cc8b221f890a8bd9f8e0c9fb1f90e6/packages/transactions/src/contract-abi.ts#L174-L175

(https://github.com/hirosystems/stacks.js/blob/f39637c924cc8b221f890a8bd9f8e0c9fb1f90e6/packages/transactions/src/contract-abi.ts)
    case ClarityAbiTypeId.ClarityAbiTypeBuffer:
      return bufferCV(utf8ToBytes(val));

janniks:

- fixes https://github.com/hirosystems/stacks.js/issues/1674

### DIFF
--- a/packages/transactions/src/contract-abi.ts
+++ b/packages/transactions/src/contract-abi.ts
@@ -1,22 +1,22 @@
-import { cloneDeep } from './utils';
+import { hexToBytes, utf8ToBytes } from '@stacks/common';
 import {
-  ClarityValue,
-  uintCV,
-  intCV,
-  contractPrincipalCV,
-  standardPrincipalCV,
-  noneCV,
-  bufferCV,
-  falseCV,
-  trueCV,
   ClarityType,
-  getCVTypeString,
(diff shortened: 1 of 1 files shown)


obycode (approved):
Thanks @janniks. This is indeed the fix I was hoping for!

The only concern I have is that if anyone is relying on that old incorrect behavior, this would be a problem for them. Especially concerning would be a scenario like a wallet using this function for processing the memo field, since most exchanges use numbers that would parse as valid hex, but would result in a different memo than expected.

PR #1680, opened 2024-04-22
feat: add encodeAbiClarityValue method for better encoding

### LINKED ISSUE #1465: Optional input value not correctly encoded
    "@stacks/transactions": "4.1.0",

### Describe the bug

I am using encodeClarityValue with optional(uint) and value "" and receive u0

#### How to reproduce

Try explorer sandbox with pox-2 allow contract caller function.

#### Expected behavior

The result should be `none`

### Additional context
https://github.com/hirosystems/explorer/issues/956

### LINKED ISSUE #1674: Fix ABI buffer encoding

janniks:

- fixes https://github.com/hirosystems/stacks.js/issues/1674
- fixes https://github.com/hirosystems/stacks.js/issues/1465
- closes #1679 

to be safe: deprecates old method, to make clear that this might be breaking for some people (e.g. could by used for something users might not expect and change behavior of txs in practice)

### DIFF
--- a/packages/transactions/src/contract-abi.ts
+++ b/packages/transactions/src/contract-abi.ts
@@ -1,22 +1,23 @@
-import { cloneDeep } from './utils';
+import { hexToBytes, utf8ToBytes } from '@stacks/common';
 import {
-  ClarityValue,
-  uintCV,
-  intCV,
-  contractPrincipalCV,
-  standardPrincipalCV,
-  noneCV,
-  bufferCV,
-  falseCV,
-  trueCV,
   ClarityType,
-  getCVTypeString,
+  ClarityValue,
+  bufferCV,
   bufferCVFromString,
+  contractPrincipalCV,
+  falseCV,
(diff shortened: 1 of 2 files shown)


obycode (approved):
LGTM. Thanks for this!

PR #1681, opened 2024-04-24
feat: Add `Cl.parse` Clarity value parser

janniks:

- Adds `Cl.parse`, which allows us to parse clarity syntax as JS representation `ClarityValue`. 
- Renames `prettyPrint` to `stringify` (matching `JSON.` more closely in naming) _got feedback from users that they were expecting `prettyPrint` to `console.log` and they were confused that there wasn't any output._

### Examples

```ts
Cl.parse(`"my string"`);
Cl.parse('-4');
Cl.parse('u5');
Cl.parse('false');
Cl.parse('{ a: 3 }');
Cl.parse('(list 1 2 3)');
```

### DIFF
--- a/packages/transactions/src/clarity/parser.ts
+++ b/packages/transactions/src/clarity/parser.ts
@@ -280,13 +280,13 @@ function clErr(): Combinator {
 function clValue(map: (combinator: Combinator) => Combinator = v => v) {
   return either(
     [
+      clBuffer,
+      clAscii,
+      clUtf8,
       clInt,
       clUint,
       clBool,
       clPrincipal,
-      clAscii,
-      clBuffer,
-      clUtf8,
       clList,
       clTuple,
       clNone,
--- a/packages/transactions/src/clarity/parser.ts
+++ b/packages/transactions/src/clarity/parser.ts
@@ -291,6 +291,17 @@ function clValue(map: (combinator: Combinator) => Combinator = v => v) {
   );
 }
 
+/**
+ * Parse a piece of string text as Clarity value syntax.
+ * Supports all Clarity value types (primitives, sequences, composite types).
+ *
+ * @example
+ * ```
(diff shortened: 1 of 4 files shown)


packages/transactions/src/clarity/prettyPrint.ts
```diff
@@ -126,6 +126,9 @@ function prettyPrintWithDepth(cv: ClarityValue, space = 0, depth: number): strin
  * // }
  * ```
  */
-export function prettyPrint(cv: ClarityValue, space = 0): string {
+export function stringify(cv: ClarityValue, space = 0): string {
```
hugoclrd, line 129: 💯

packages/transactions/src/clarity/parser.ts
```diff
@@ -0,0 +1,318 @@
...
+  return parens(
+    sequence([
+      regex(/list/),
+      greedy(0, sequence([whitespace(), clValue()]), c => Cl.list(c as ClarityValue[])),
+    ])
+  );
+}
+
+function clTuple(): Combinator {
+  // todo: add `(tuple` syntax
```
hugoclrd: This todo can be removed

hugoclrd (changes requested):
I tried it and found this use case that throws a parse error:
```clarity
{ a: 0x68656c6c6f21 }
```

janniks:
> I tried it and found this use case that throws a parse error:
> 
> ```clar
> { a: 0x68656c6c6f21 }
> ```

@hugocaillard  Good catch, noticed that nested values will be tricky with overlaps. I thought the `entire` combinator solved this, but that's not used for the nested values.. 🙃 -- updated the order to fix this case, there might be more though..

In this case it was detecting the `0x..` as `0` matches an int, and will continue instead of trying other cl-values.

hugoclrd (approved):
LGTM 🚀

PR #1685, opened 2024-05-01
fix!: Disable legacy triplesec support for mnemonic encryption

janniks:
> This PR was published to npm with the version `6.13.0`
> e.g. `npm install @stacks/common@6.13.0 --save-exact`

- removes support for legacy triplesec encryption of mnemonics


zone117x (approved):
5+ years of depreciation 
👋

PR #1707, opened 2024-06-17
refactor!: Make StacksTransaction.serialize() return a hex-string instead of bytes

janniks:
> This PR was published to npm with the version `6.14.1-pr.42+0f50711f`
> e.g. `npm install @stacks/common@6.14.1-pr.42+0f50711f --save-exact`

- Make StacksTransaction.serialize() return a hex-string instead of bytes


MicaiahReid (approved):
Exciting!

PR #1708, opened 2024-06-17
refactor!: Rename more types to use "Wire" suffix

janniks:
> This PR was published to npm with the version `6.14.1-pr.43+5f1cb1b7`
> e.g. `npm install @stacks/common@6.14.1-pr.43+5f1cb1b7 --save-exact`

- Renames more internals to have the `Wire` suffix (maybe not the best name, but should indicate wire-format / serialization representation)
- _Just_ renames, no logic change

I would like to do this to open more naming options for things developers will actually use. Ideally the Wire suffix isn't seen by most users, so next steps would also reduce this type in function signatures of helpers (reducing more wrapper types that are unnecessarily typed for most use-cases)

### DIFF
--- a/.github/MIGRATION.md
+++ b/.github/MIGRATION.md
@@ -198,7 +198,7 @@ For easier migrating, renaming the following methods is possible to keep the pre
 The following interfaces and methods were renamed:
 
 - `AssetInfo` → `Asset`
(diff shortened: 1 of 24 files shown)


MicaiahReid (approved):
> `MessageSignature`

Tired 😴 

> `MessageSignatureWire`

Wired 🤪

PR #1710, opened 2024-06-22
feat: Add non-sequential multi-sig support

### LINKED ISSUE #1487: Multisig transaction signing order
### Describe the bug

I don't know if it's even possible, but when I try to sign a multisig transaction starting from the third address (out of 3 public keys) it gives an error "Signer hash does not equal hash of public key(s): ba3f7c941ce749ba14f2a7c24d067948858ba16d != 0f8b02bbaa17b8df1eaede6102baef275a726914"
Tried the following workarounds from this [issue](https://github.com/hirosystems/stacks.js/issues/962), but seems like it is only possible to sign a multisig transaction starting from lower index to a higher one from the initial public keys array - like (0, 1), (1, 2) (0, 2), but not (2, 1). So when the last owner, for example, wants to initiate a multisig transaction, he can't do it anyhow, as well as if the second owner initiates the transaction - the first one won't be able to confirm it after that

#### How to reproduce

```javascript
import {
  TransactionSigner,
  getPublicKey,
  createStacksPrivateKey,
  publicKeyToString,
  standardPrincipalCV,
  broadcastTransaction,
  makeUnsignedSTXTokenTransfer,
import BigNum from 'bn.js';

// Private key strings with compression byte appended
const privKeyStrings = [
];

// Compressed public keys
const pubKeys = privKeys.map(getPublicKey);

const recipient = standardPrincipalCV('SN3C9PPBTKJNWJ5HZBBWERRAF4E3W6M39AN54S92');
const amount = new BigNum(74);
const fee = new BigNum(278);
const nonce = new BigNum(0);
const memo = 'test';

(async () => {
    recipient,
    amount,
    network,
    fee,
    nonce,
    memo,
    numSignatures: 2,
    publicKeys: pubKeyStrings,
  });

  signer.appendOrigin(pubKeys[0]);
  signer.signOrigin(privKeys[1]);


  const txId = await broadcastTransaction(transaction, network);
  console.log('Transaction ID: ' + JSON.stringify(txId, null, 2).replace(/^"|"$/, ''));
})().catch((err) => {
  console.error('Error:' + err);
  process.exit(1);
});
```

#### Expected behavior

It should be possible to initiate a signature from whatever owner

janniks:

- Adds non-sequential multi-sig support ✨ closes https://github.com/hirosystems/stacks.js/issues/1487

* Fixes incorrect compression check in signing algorithm
* Fixes incorrect compression in deserialization
* Fixes incorrect field order for legacy (sequential) multisig
* Cleans up opts types/interfaces for less duplication

**Related**: 
- [📓 SIP: non-sequential-multisig-transactions](https://github.com/jbencin/sips/blob/sip-02x-non-sequential-multisig-transactions/sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md)
- Core PR https://github.com/stacks-network/stacks-core/pull/3710

---

Adds a `useNonSequentialMultiSig` flag to multi-sig transaction creation methods. (Along with the internals to support it)

So far this opt is marked `@experimental`, since the interface of the default could change. (And we likely want to see how the authorization type plays out on mainnet, before switching fully)

Example

```ts
      const tx = await makeSTXTokenTransfer({
        recipient: 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6',
        amount: 12_345n,
        fee: 1_000n,
        nonce: 2n,
        network: 'testnet',

        numSignatures: required,
        publicKeys: [k1, k2, k3],
        signerKeys: [k3, k1], // could sign in any order (even if serializing in between and not sharing state)

        useNonSequentialMultiSig: true, // <--- NEW OPTIONAL FLAG
      });
```

Comment

> The separate signing test really shows there is more work needed to make this easy to use for folks building a "real" multi-sig experience -- with serializing the transaction in between signatures. This could be added as a separate issue or addressed in a breaking change (there is an opportunity to clean up the TransactionSigner as well, which is intertwined with the StacksTransaction class).

> Follow up thoughts: This can probably be addressed with one mutating public method on the `StacksTransaction` called something like `.finalize()` or `.finalizeMultiSig()`, which ensures the transaction-auth-fields are in the intended order. (Might need the list of public-keys again, or can do an opinionated `.sort` first, but technically should work for legacy multi-sig as well).

### DIFF
--- a/packages/transactions/src/authorization.ts
+++ b/packages/transactions/src/authorization.ts
@@ -273,8 +273,13 @@ export function deserializeMultiSigSpendingCondition(
   // Partially signed multi-sig tx can be serialized and deserialized without exception (Incorrect number of signatures)
   // No need to check numSigs !== signaturesRequired to throw Incorrect number of signatures error
 
-  if (haveUncompressed && hashMode === AddressHashMode.SerializeP2SH)
+  if (
+    haveUncompressed &&
+    (hashMode === AddressHashMode.SerializeP2WSH ||
+      hashMode === AddressHashMode.SerializeP2WSHNonSequential)
+  ) {
     throw new VerificationError('Uncompressed keys are not allowed in this hash mode');
+  }
 
   return {
     hashMode,
@@ -502,7 +507,11 @@ function verifyMultiSig(
   )
     throw new VerificationError('Incorrect number of signatures');
 
-  if (haveUncompressed && condition.hashMode === AddressHashMode.SerializeP2SH)
+  if (
+    haveUncompressed &&
+    (condition.hashMode === AddressHashMode.SerializeP2WSH ||
+      condition.hashMode === AddressHashMode.SerializeP2WSHNonSequential)
+  )
     throw new VerificationError('Uncompressed keys are not allowed in this hash mode');
 
   const addrBytes = addressFromPublicKeys(
--- a/packages/transactions/src/authorization.ts
+++ b/packages/transactions/src/authorization.ts
@@ -120,7 +120,14 @@ export function createMultiSigSpendingCondition(
   nonce: IntegerType,
   fee: IntegerType
 ): MultiSigSpendingCondition {
-  const stacksPublicKeys = pubKeys.map(createStacksPublicKey);
+  const publicKeys = pubKeys.slice();
+
+  // sort public keys (only for newer non-sequential multi-sig for backwards compatibility)
+  if (isNonSequential(hashMode)) {
+    publicKeys.sort();
+  }
+
+  const stacksPublicKeys = publicKeys.map(createStacksPublicKey);
 
   // address version arg doesn't matter for signer hash generation
   const signer = addressFromPublicKeys(
@@ -140,10 +147,24 @@ export function createMultiSigSpendingCondition(
   };
 }
 
+/** @internal */
 export function isSingleSig(
   condition: SpendingConditionOpts
 ): condition is SingleSigSpendingConditionOpts {
   return 'signature' in condition;
+}
+
+/** @internal */
+export function isSequential(hashMode: AddressHashMode): boolean {
+  return hashMode === AddressHashMode.SerializeP2SH || hashMode === AddressHashMode.SerializeP2WSH;
+}
+
+/** @internal */
+export function isNonSequential(hashMode: AddressHashMode): boolean {
+  return (
+    hashMode === AddressHashMode.SerializeP2SHNonSequential ||
+    hashMode === AddressHashMode.SerializeP2WSHNonSequential
+  );
 }
 
 function clearCondition(condition: SpendingConditionOpts): SpendingCondition {
@@ -463,7 +484,7 @@ function verifyMultiSig(
       case StacksMessageType.MessageSignature:
         if (field.pubKeyEncoding === PubKeyEncoding.Uncompressed) haveUncompressed = true;
         const { pubKey, nextSigHash } = nextVerification(
-          curSigHash,
+          isNonSequential(condition.hashMode) ? initialSigHash : curSigHash,
           authType,
           condition.fee,
           condition.nonce,
@@ -481,7 +502,10 @@ function verifyMultiSig(
     publicKeys.push(foundPubKey);
   }
 
-  if (numSigs !== condition.signaturesRequired)
+  if (
+    (isSequential(condition.hashMode) && numSigs !== condition.signaturesRequired) ||
+    (isNonSequential(condition.hashMode) && numSigs < condition.signaturesRequired)
+  )
     throw new VerificationError('Incorrect number of signatures');
 
   if (haveUncompressed && condition.hashMode === AddressHashMode.SerializeP2SH)
@@ -554,7 +578,7 @@ export function verifyOrigin(auth: Authorization, initialSigHash: string): strin
     case AuthType.Standard:
       return verify(auth.spendingCondition, initialSigHash, AuthType.Standard);
     case AuthType.Sponsored:
-      return verify(auth.spendingCondition, initialSigHash, AuthType.Standard);
+      return verify(auth.spendingCondition, initialSigHash, AuthType.Standard); // todo: should this be .Sponsored?
     default:
       throw new SigningError('Invalid origin auth type');
   }
--- a/packages/transactions/src/authorization.ts
+++ b/packages/transactions/src/authorization.ts
@@ -120,14 +120,7 @@ export function createMultiSigSpendingCondition(
   nonce: IntegerType,
   fee: IntegerType
 ): MultiSigSpendingCondition {
-  const publicKeys = pubKeys.slice();
-
-  // sort public keys (only for newer non-sequential multi-sig for backwards compatibility)
-  if (isNonSequential(hashMode)) {
-    publicKeys.sort();
-  }
-
-  const stacksPublicKeys = publicKeys.map(createStacksPublicKey);
+  const stacksPublicKeys = pubKeys.map(createStacksPublicKey);
 
   // address version arg doesn't matter for signer hash generation
   const signer = addressFromPublicKeys(
@@ -155,12 +148,12 @@ export function isSingleSig(
 }
 
 /** @internal */
-export function isSequential(hashMode: AddressHashMode): boolean {
+export function isSequentialMultiSig(hashMode: AddressHashMode): boolean {
   return hashMode === AddressHashMode.SerializeP2SH || hashMode === AddressHashMode.SerializeP2WSH;
 }
 
 /** @internal */
-export function isNonSequential(hashMode: AddressHashMode): boolean {
+export function isNonSequentialMultiSig(hashMode: AddressHashMode): boolean {
   return (
     hashMode === AddressHashMode.SerializeP2SHNonSequential ||
     hashMode === AddressHashMode.SerializeP2WSHNonSequential
@@ -469,42 +462,43 @@ function verifyMultiSig(
   authType: AuthType
 ): string {
   const publicKeys: StacksPublicKey[] = [];
+
   let curSigHash = initialSigHash;
   let haveUncompressed = false;
   let numSigs = 0;
 
   for (const field of condition.fields) {
-    let foundPubKey: StacksPublicKey;
-
     switch (field.contents.type) {
       case StacksMessageType.PublicKey:
         if (!isCompressed(field.contents)) haveUncompressed = true;
-        foundPubKey = field.contents;
+        publicKeys.push(field.contents);
         break;
       case StacksMessageType.MessageSignature:
         if (field.pubKeyEncoding === PubKeyEncoding.Uncompressed) haveUncompressed = true;
         const { pubKey, nextSigHash } = nextVerification(
-          isNonSequential(condition.hashMode) ? initialSigHash : curSigHash,
+          curSigHash,
           authType,
           condition.fee,
           condition.nonce,
           field.pubKeyEncoding,
           field.contents
         );
-        curSigHash = nextSigHash;
-        foundPubKey = pubKey;
+
+        if (isSequentialMultiSig(condition.hashMode)) {
+          curSigHash = nextSigHash;
+        }
+
+        publicKeys.push(pubKey);
 
         numSigs += 1;
         if (numSigs === 65536) throw new VerificationError('Too many signatures');
-
         break;
     }
-    publicKeys.push(foundPubKey);
   }
 
   if (
-    (isSequential(condition.hashMode) && numSigs !== condition.signaturesRequired) ||
-    (isNonSequential(condition.hashMode) && numSigs < condition.signaturesRequired)
+    (isSequentialMultiSig(condition.hashMode) && numSigs !== condition.signaturesRequired) ||
+    (isNonSequentialMultiSig(condition.hashMode) && numSigs < condition.signaturesRequired)
   )
     throw new VerificationError('Incorrect number of signatures');
 
--- a/packages/transactions/src/builders.ts
+++ b/packages/transactions/src/builders.ts
@@ -1738,12 +1738,14 @@ function mutatingSignAppendMultiSig(
 
   const signer = new TransactionSigner(transaction);
 
-  const pubs = sortPublicKeysForAddress(
-    publicKeys,
-    transaction.auth.spendingCondition.signaturesRequired,
-    transaction.auth.spendingCondition.hashMode,
-    address ? createAddress(address).hash160 : undefined
-  );
+  const pubs = address
+    ? sortPublicKeysForAddress(
+        publicKeys,
+        transaction.auth.spendingCondition.signaturesRequired,
+        transaction.auth.spendingCondition.hashMode,
+        createAddress(address).hash160
+      )
+    : publicKeys;
 
   // sign in order of public keys
   for (const publicKey of pubs) {
@@ -1763,10 +1765,8 @@ function sortPublicKeysForAddress(
   publicKeys: string[],
   numSigs: number,
   hashMode: MultiSigHashMode,
-  hash?: string
+  hash: string
 ): string[] {
-  if (!hash) return publicKeys;
-
   // unsorted
   const hashUnsorted = addressFromPublicKeys(
     0 as any, // only used for hash, so version doesn't matter
--- a/packages/transactions/src/builders.ts
+++ b/packages/transactions/src/builders.ts
@@ -16,6 +16,7 @@ import {
   createStandardAuth,
   SpendingCondition,
   MultiSigSpendingCondition,
+  isNonSequential,
 } from './authorization';
 import { ClarityValue, deserializeCV, NoneCV, PrincipalCV, serializeCV } from './clarity';
 import {
@@ -593,11 +594,23 @@ function deriveNetwork(transaction: StacksTransaction) {
   }
 }
 
+/** @deprecated Not used internally */
 export interface MultiSigOptions {
   numSignatures: number;
   publicKeys: string[];
   signerKeys?: string[];
 }
+
+export interface UnsignedMultiSigOptions {
+  numSignatures: number;
+  publicKeys: string[];
+  /** @experimental Use newer non-sequential multi-sig hashmode for transaction. Future releases may make this the default. */
+  useNonSequentialMultiSig?: boolean;
+}
+
+export type SignedMultiSigOptions = UnsignedMultiSigOptions & {
+  signerKeys: string[];
+};
 
 /**
  * STX token transfer transaction options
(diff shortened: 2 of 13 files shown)


packages/transactions/src/authorization.ts
```diff
@@ -481,7 +502,10 @@ function verifyMultiSig(
     publicKeys.push(foundPubKey);
   }
 
-  if (numSigs !== condition.signaturesRequired)
+  if (
+    (isSequential(condition.hashMode) && numSigs !== condition.signaturesRequired) ||
+    (isNonSequential(condition.hashMode) && numSigs < condition.signaturesRequired)
```
janniks: In new mode it's okay to go over signature count

packages/transactions/src/builders.ts
```diff
@@ -630,16 +643,9 @@ export interface SignedTokenTransferOptions extends TokenTransferOptions {
   senderKey: string;
 }
 
-export interface UnsignedMultiSigTokenTransferOptions extends TokenTransferOptions {
-  numSignatures: number;
-  publicKeys: string[];
-}
+export type UnsignedMultiSigTokenTransferOptions = TokenTransferOptions & UnsignedMultiSigOptions;
```
janniks, line 658: Refactored these types to be more reusable. Does anybody think the change from `interface` to `type` is a problem?

packages/transactions/src/builders.ts
```diff
@@ -747,21 +755,11 @@ export async function makeSTXTokenTransfer(
...
-    for (const key of txOptions.signerKeys) {
-      const pubKey = pubKeyfromPrivKey(key);
-      pubKeys = pubKeys.filter(pk => pk !== bytesToHex(pubKey.data));
-      signer.signOrigin(createStacksPrivateKey(key));
-    }
-
-    for (const key of pubKeys) {
-      signer.appendOrigin(publicKeyFromBytes(hexToBytes(key)));
-    }
+    mutatingSignMultiSig(transaction, txOptions.publicKeys.slice(), txOptions.signerKeys);
```
janniks: Refactored into function (even if mutating) to be more reusable)

packages/transactions/src/builders.ts
```diff
@@ -1692,3 +1663,30 @@ export async function estimateTransactionFeeWithFallback(
...
+    const signerKey = signerKeys.find(key => bytesToHex(pubKeyfromPrivKey(key).data) === publicKey);
+    if (signerKey) {
+      // either sign and append message signature (which allows for recovering the public key)
+      signer.signOrigin(createStacksPrivateKey(signerKey));
+    } else {
+      // or append the public key (which did not sign here)
+      signer.appendOrigin(publicKeyFromBytes(hexToBytes(publicKey)));
+    }
+  }
+}
```
janniks, line 1761: @jbencin you'll notice this now differs from the blockchain `make` methods, but I think is more correct. We try to sign/append in order of public-keys, rather than first all signs and then all appends _(not exactly sure why that was in the code previously, but seems to be inspired by stacks-core methods)_

packages/transactions/src/authorization.ts
```diff
@@ -120,7 +120,14 @@ export function createMultiSigSpendingCondition(
   nonce: IntegerType,
   fee: IntegerType
 ): MultiSigSpendingCondition {
-  const stacksPublicKeys = pubKeys.map(createStacksPublicKey);
+  const publicKeys = pubKeys.slice();
+
+  // sort public keys (only for newer non-sequential multi-sig for backwards compatibility)
```
jbencin: This should be handled the same way in all multisig transaction types: We *recommend* that the user sort their pubkeys when creating, but we can't force it because users may have funds in accounts that don't use the recommended ordering

You could add a `boolean` argument, or create another version of this function, `createMultiSigSpendingConditionSorted`, if you want to provide sorting functionality to the user. But you can't just assume the user wants it

janniks: Got it. I would like to be more opinionated about it, so maybe sorting can be the default and we add a `disablePublicKeySorting` flag. But yeah, this should be a part of all the helpers then as well.. Will change 👍

janniks: Or maybe not, 🤔 technically the developer can still update and replace conditions/modes/etc. on any transaction fairly easily in Stacks.js (assuming they know what they're doing). Similar to how we don't really _expose_ p2wsh multi-sig, we can "hide" the unsorted variant a bit more as well 🤷🏻‍♀️ trying to find a balance for users to be able to get less wrong by accident.

packages/transactions/src/authorization.ts
```diff
@@ -463,7 +484,7 @@ function verifyMultiSig(
       case StacksMessageType.MessageSignature:
         if (field.pubKeyEncoding === PubKeyEncoding.Uncompressed) haveUncompressed = true;
         const { pubKey, nextSigHash } = nextVerification(
-          curSigHash,
+          isNonSequential(condition.hashMode) ? initialSigHash : curSigHash,
```
jbencin: How about this instead:

```typescript
if (isSequential(condition.hashMode)) {
  curSigHash = nextSigHash;
}
```

Think about it this way: In order-independent mode, we just don't update `curSigHash`. It's easier to read than updating `curSigHash`, and then not using it

packages/transactions/src/builders.ts
```diff
@@ -593,12 +594,24 @@ function deriveNetwork(transaction: StacksTransaction) {
...
 export interface MultiSigOptions {
   numSignatures: number;
   publicKeys: string[];
   signerKeys?: string[];
 }
 
+export interface UnsignedMultiSigOptions {
+  numSignatures: number;
+  publicKeys: string[];
+  /** @experimental Use newer non-sequential multi-sig hashmode for transaction. Future releases may make this the default. */
```
jbencin, line 619: I'm sure we will want this to be the default once 3.0 is released

packages/transactions/src/authorization.ts
janniks: Yeah, good point that's fine by me 👍🏻 I was going for more control flow where hashmode is used, but I like the elimination of unnecessary work for the readers view

packages/transactions/src/builders.ts
janniks: Yes, we'll do a second release but that will have to be a breaking change for these helper methods.

packages/transactions/src/builders.ts
```diff
@@ -1692,3 +1663,30 @@ export async function estimateTransactionFeeWithFallback(
...
+/** @internal multi-sig signing re-use */
+function mutatingSignMultiSig(
+  /** **Warning:** method mutates `transaction` */
+  transaction: StacksTransaction,
+  publicKeys: string[],
+  signerKeys: string[]
+) {
+  const signer = new TransactionSigner(transaction);
+
+  // sort public keys (only for newer non-sequential multi-sig for backwards compatibility)
```
jbencin: Again, can't force key ordering for any multisig type

packages/transactions/src/builders.ts
```diff
@@ -1692,3 +1663,30 @@ export async function estimateTransactionFeeWithFallback(
     throw error;
   }
 }
+
+/** @internal multi-sig signing re-use */
+function mutatingSignMultiSig(
```
jbencin: This function can sign *or* append a pubkey. Seems like the name should reflect that

packages/transactions/src/signature.ts
```diff
@@ -65,7 +68,7 @@ export function deserializeTransactionAuthField(bytesReader: BytesReader): Trans
     case AuthFieldType.PublicKeyUncompressed:
       return createTransactionAuthField(
         PubKeyEncoding.Uncompressed,
-        deserializePublicKey(bytesReader)
+        uncompressPublicKey(deserializePublicKey(bytesReader).data)
```
jbencin, line 71: Is the return value of `deserializePublicKey()` always compressed? Is it safe to call `uncompressPublicKey()` on an already uncompressed public key?

packages/transactions/src/signer.ts
```diff
@@ -80,8 +86,12 @@ export class TransactionSigner {
       throw new SigningError('"transaction.auth.spendingCondition" is undefined');
     }
 
-    if (!isSingleSig(this.transaction.auth.spendingCondition)) {
-      const spendingCondition = this.transaction.auth.spendingCondition;
+    const spendingCondition = this.transaction.auth.spendingCondition;
+    if (
+      spendingCondition.hashMode === AddressHashMode.SerializeP2SH ||
+      spendingCondition.hashMode === AddressHashMode.SerializeP2WSH
+    ) {
```
jbencin, line 92: Why not

```suggestion
    if (isSequential(spendingCondition.hashMode)) {
```

packages/transactions/src/signer.ts
```diff
@@ -92,7 +102,11 @@ export class TransactionSigner {
       }
     }
 
-    const nextSighash = this.transaction.signNextOrigin(this.sigHash, privateKey);
+    const curSigHash = isNonSequential(this.transaction.auth.spendingCondition.hashMode)
+      ? this.transaction.signBegin() // non-sequential multi-sig always signs original sighash
+      : this.sigHash;
+
+    const nextSighash = this.transaction.signNextOrigin(curSigHash, privateKey);
     this.sigHash = nextSighash;
```
jbencin: Again, I'd recommend not assigning `nextSigHash` to anything in order-independent mode. Just keep `this.sigHash` at the initial value. This avoids having a variable that is unusable and misleading

packages/transactions/tests/builder.test.ts
```diff
@@ -2359,3 +2370,279 @@ test.each([
...
+        serialized = transactionToHex(tx);
+      }
+
+      const txFinal = deserializeTransaction(serialized);
+      expect(() => txFinal.verifyOrigin()).not.toThrow();
+    });
+
+    test('non sequential multi-sig test vector', async () => {
+      // test vector generated from manual test in stacks-core
+      const EXPECTED =
```
jbencin, line 2574: I assume have validated this hex string with `stacks-inspect decode-tx`?

packages/transactions/tests/builder.test.ts
```diff
@@ -2359,3 +2370,279 @@ test.each([
 
   expect(bytesToHex(transaction.serialize())).toEqual(txBytes);
 });
+
+describe('multi-sig', () => {
```
jbencin, line 2375: I would add a couple tests where you make a valid transaction, flip a bit or two, and expect `verifyOrigin()` to fail

packages/transactions/tests/builder.test.ts
```diff
@@ -2359,3 +2370,279 @@ test.each([
...
+      fee: 1_000_000n,
+      nonce: 0n,
+      network: 'testnet',
+      anchorMode: AnchorMode.Any,
+      numSignatures: 2,
+      publicKeys,
+      signerKeys,
+    });
+
+    expect(() => tx.verifyOrigin()).not.toThrow();
```
jbencin, line 2396: In addition to `verifyOrigin()` you could add a few asserts at the end of these tests, like check that we have the expected number of auth fields and signatures, that they match the keys we passed, etc. Right now `MakeSTXTokenTransfer()` could be ignoring most of its args and returning a valid singlesig tx, and these unit tests would still pass

packages/transactions/src/authorization.ts
jbencin: Even with sequential multisig transactions, it was customary to sort the pubkeys before creating a transaction, but it was never enforced. So new multisig types should not handle this differently

One thing you could do, which would work for both types and would be backwards compatible, is to add an optional `address` parameter. Users would be encouraged to supply the address of the multisig account, and you could check if the pubkeys generate that address, first in the order given, then sort them and try in that order. If it doesn't match in either order, return an error

packages/transactions/tests/builder.test.ts
jbencin: I just tried this, it does decode and the signatures are valid

packages/transactions/tests/builder.test.ts
janniks: I like that, good idea

packages/transactions/tests/builder.test.ts
janniks: Thanks for checking. I didn't, but I ran the rust verify version on the instance when constructing it and parsing it from hex, (assuming the command does something similar)

packages/transactions/src/authorization.ts
janniks: Lovely approach!!! I really like this more than another config flag!! 👏

janniks: Just thinking about next breaking release, I would like to change the default behavior to "sorted" to guide the ecosystem to the preferred way
.

packages/transactions/src/signature.ts
janniks: In this case yes, since the wire format assumes compressed public-keys (from what I can tell, always reading the same amount of bytes). So this was a bug previously in Stacks.js

packages/transactions/src/signer.ts
janniks: Right, this didn't work due to type info here, ~I now updated the helpers to indicate they're only used for multi-sig hashmodes going in ✅~ nvm didn't work everywhere

janniks: nvm, updated -- mb

janniks:
Ready for re-review @jbencin 🙏

packages/transactions/src/builders.ts
```diff
@@ -1692,3 +1723,70 @@ export async function estimateTransactionFeeWithFallback(
...
+      signer.signOrigin(createStacksPrivateKey(signerKey));
+    } else {
+      // or append the public key (which did not sign here)
+      signer.appendOrigin(publicKeyFromBytes(hexToBytes(publicKey)));
+    }
+  }
+}
+
+/** @internal Get the matching public-keys array for a multi-sig address */
+function ensureValidMultiSigPublicKeyOrder(
```
jbencin: Using the word "ensure" implies this function either returns a boolean, or modifies it's arguments. Instead, it returns the pubkeys in the correct order. I would rename this to something like `sortPublicKeysForAddress()`

packages/transactions/src/builders.ts
```diff
@@ -1692,3 +1723,70 @@ export async function estimateTransactionFeeWithFallback(
...
+  publicKeys: string[],
+  numSigs: number,
+  hashMode: MultiSigHashMode,
+  hash?: string
```
jbencin: There's really no point in calling this function without the address

packages/transactions/src/builders.ts
janniks: We typically have our ensures `throw` if the bad condition occurs, but this is internal so idc either way 👍

packages/transactions/src/builders.ts
janniks: That's true, this was to push the check down, but I can try to improve this 🤔

packages/transactions/src/builders.ts
```diff
@@ -1692,3 +1723,70 @@ export async function estimateTransactionFeeWithFallback(
...
+    throw new Error('Transaction is not a multi-sig transaction');
+  }
+
+  const signer = new TransactionSigner(transaction);
+
+  const pubs = ensureValidMultiSigPublicKeyOrder(
+    publicKeys,
+    transaction.auth.spendingCondition.signaturesRequired,
+    transaction.auth.spendingCondition.hashMode,
+    address ? createAddress(address).hash160 : undefined
```
jbencin: The ternary operator here is kinda pointless, it would just map any "falsy" value like `0`, `false`, or `null` to `undefined`

packages/transactions/src/signature.ts
jbencin: Shouldn't you be calling `uncompressPublicKey()` under `case AuthFieldType.PublicKeyCompressed` though? This seems backwards

packages/transactions/src/builders.ts
janniks: Removed the previous optionality of this param, but kept the ternary (that's fine in this case, right? We want falsy values not to be treated as an address. We typically don't strictly check each param type and trust most developers will have JS/TS tooling to disallow most types. At least that's a common pattern in the codebase.)

packages/transactions/src/signature.ts
janniks: This is in the deserializer code. The wire format comes as compressed public-keys and an encoding flag, which indicate how they should be used. So the pub-key-encoding here is telling us we need to uncompress first if we want to use the public-key for something.

janniks: I'm working under this assumption from SIP-005
> The spending field body, which will be the following, depending on the field ID:
0x00 or 0x01: The next 33 bytes are a compressed secp256k1 public key. If the field ID is 0x00, the key will be loaded as a compressed secp256k1 public key. If it is 0x01, then the key will be loaded as an uncompressed secp256k1 public key.

packages/transactions/src/signer.ts
```diff
@@ -93,7 +102,13 @@ export class TransactionSigner {
     }
 
     const nextSighash = this.transaction.signNextOrigin(this.sigHash, privateKey);
-    this.sigHash = nextSighash;
+
+    if (
+      isSingleSig(this.transaction.auth.spendingCondition) ||
```
jbencin, line 107: Probably not necessary

packages/transactions/src/transaction.ts
```diff
@@ -164,10 +170,11 @@ export class StacksTransaction {
     if (isSingleSig(condition)) {
       condition.signature = nextSig;
     } else {
-      const compressed = bytesToHex(privateKey.data).endsWith('01');
       condition.fields.push(
         createTransactionAuthField(
-          compressed ? PubKeyEncoding.Compressed : PubKeyEncoding.Uncompressed,
+          privateKey.data.byteLength === PRIVATE_KEY_COMPRESSED_LENGTH
+            ? PubKeyEncoding.Compressed
+            : PubKeyEncoding.Uncompressed,
```
jbencin, line 177: Would be safer to have a `switch` here, and fail if `privateKey.data.byteLength` isn't `PRIVATE_KEY_COMPRESSED_LENGTH` or `PRIVATE_KEY_UNCOMPRESSED_LENGTH`

packages/transactions/src/signer.ts
janniks: Yeah, this is just for typescript.  I can do it with a type cast but I typically try not to. Or keep the raw comparisons which can type infer in TS

packages/transactions/src/builders.ts
jbencin: Makes sense, but could also return a `null` and let the calling function decide whether to `throw` or recover. Seems like the function is trying to do to things that maybe shouldn't go together

packages/transactions/src/signature.ts
jbencin: Reading the SIP it looks like you are correct. The naming here is just confusing

jbencin (approved):
Good work here! Any of the discussions that are still open I consider very minor and won't hold up approval for them

PR #1714, opened 2024-07-02
fix: update multi-sig signer state

janniks:
> This PR was published to npm with the version `6.16.0`
> e.g. `npm install @stacks/common@6.16.0 --save-exact`

- minor update to the multi sig changes (signer sighash was not correct in all states)

### DIFF
--- a/packages/transactions/src/signer.ts
+++ b/packages/transactions/src/signer.ts
@@ -1,6 +1,7 @@
 import { StacksTransaction } from './transaction';
 import { StacksPrivateKey, StacksPublicKey } from './keys';
 import {
+  isNonSequentialMultiSig,
   isSequentialMultiSig,
   isSingleSig,
   nextVerification,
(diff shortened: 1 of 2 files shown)


packages/transactions/src/signer.ts
```diff
@@ -38,16 +39,19 @@ export class TransactionSigner {
...
+        const nextVerify = nextVerification(
+          this.sigHash,
+          transaction.auth.authType,
+          spendingCondition.fee,
+          spendingCondition.nonce,
+          PubKeyEncoding.Compressed, // always compressed for multisig
+          signature
+        );
+
+        if (!isNonSequentialMultiSig(spendingCondition.hashMode)) {
```
jbencin, line 54: Could also do

```suggestion
        if (isSequentialMultiSig(spendingCondition.hashMode)) {
```

jbencin (approved):

PR #1719, opened 2024-07-08
feat!: Update post-condition representation to humand readable types

janniks:
- update post-condition representations for [SIP](https://github.com/stacksgov/sips/pull/166)

---

**Example** Post-conditions should now be human-readable and serializable. This is a breaking change.

```ts
const condition = {
  type: 'stx-postcondition',
  address: 'SP...123',
  condition: 'eq',
  amount: 10000
}
```

### DIFF
--- a/.github/MIGRATION.md
+++ b/.github/MIGRATION.md
@@ -7,6 +7,7 @@
   - [StacksNodeApi](#stacksnodeapi)
   - [StacksNetwork to StacksNodeApi](#stacksnetwork-to-stacksnodeapi)
   - [Clarity Representation](#clarity-representation)
+  - [Post-conditions](#post-conditions)
   - [`serialize` methods](#serialize-methods)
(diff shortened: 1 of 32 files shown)


hugoclrd (approved):
👏 👏 

Great changes!

I have a doubt regarding the `sent` `not-sent` keywords.
Does the past tense make more sense than the previous version (Sends, DoesNotSend)?

janniks:
> 👏 👏 
> 
> 
> 
> Great changes!
> 
> 
> 
> I have a doubt regarding the `sent` `not-sent` keywords.
> 
> Does the past tense make more sense than the previous version (Sends, DoesNotSend)?

Yeah, they're not perfect. I also prefer "will" "does" (from the perspective of the person constructing the PC before the tx, but folks on the wallets standards calls prefer red "post"-tense for "post" conditions 😅

PR #1720, opened 2024-07-12
refactor: move internal wire types to reduce circular dependencies

janniks:
> This PR was published to npm with the version `6.14.1-pr.46+dc9514f8`
> e.g. `npm install @stacks/common@6.14.1-pr.46+dc9514f8 --save-exact`

- move files for serialization to reduce possibility of circular deps — splitting functions by style (create.., serialize.., etc.) rather than type (signature, payload, etc.)
Screenshot 2024-07-12 at 17 03 17


packages/transactions/src/BytesReader.ts
zone117x, line 1: Is the casing on this intentional?

zone117x (approved):

packages/transactions/src/BytesReader.ts
janniks: Yeah, although I don't have a strong preference. I was trying to highlight that this file is only used for a single Class.

PR #1724, opened 2024-07-18
feat!: Rename networking/fetch methods to include `fetch` prefix

janniks:
> This PR was published to npm with the version `6.14.1-pr.50+348b9395`
> e.g. `npm install @stacks/common@6.14.1-pr.50+348b9395 --save-exact`

- Rename "fetch" methods for better discoverability


rafaelcr (approved):
Nice

PR #1726, opened 2024-07-18
feat: Add STX unit helpers

janniks:
> This PR was published to npm with the version `6.14.1-pr.53+91f0bd20`
> e.g. `npm install @stacks/common@6.14.1-pr.53+91f0bd20 --save-exact`

- Add unit helpers based on old `stx-utils` https://github.com/stacks-network/stacks-utils/blob/master/src/units.js


rafaelcr (approved):
Cool

PR #1735, opened 2024-07-31
refactor!: Remove wrapper type for message signatures

janniks:
> This PR was published to npm with the version `6.14.1-pr.61+4b7fd662`
> e.g. `npm install @stacks/common@6.14.1-pr.61+4b7fd662 --save-exact`

- Prefer direct `string` instead of message signature wrapper type
- Add `PublicKey` super type where TS complained


beguene (approved):
This is much cleaner !

PR #1737, opened 2024-08-15
feat: create @stacks/rpc-client package

zone117x:
The RPC schema is being removed from the `stacks-blockchain-api` codebase. This means the `@stacks/blockchain-api-client` library will no longer include auto-generated code for those endpoints. See more details over at that repo: [PR](https://github.com/hirosystems/stacks-blockchain-api/pull/2045), [client migration notes](https://github.com/hirosystems/stacks-blockchain-api/blob/refactor/fastify-wip/client/MIGRATION.md).

This PR creates a new library `@stacks/rpc-client` which generates a client library using the canonical RPC schema at 
https://github.com/stacks-network/stacks-core/blob/master/docs/rpc/openapi.yaml

The library http request types are generated using [`openapi-typescript`](https://openapi-ts.dev/introduction), and depends on [`openapi-fetch`](https://openapi-ts.dev/openapi-fetch/).

Some glue code is included to easily use the existing `StacksNetwork` objects and middleware, and also to export some common types.


packages/network/src/fetch.ts
```diff
@@ -108,9 +108,11 @@ export function createApiKeyMiddleware({
       const reqUrl = new URL(context.url);
       if (!hostMatches(reqUrl.host, host)) return; // Skip middleware if host does not match pattern
 
-      const headers = new Headers(context.init.headers);
+      const headers =
+        context.init.headers instanceof Headers
+          ? context.init.headers
+          : (context.init.headers = new Headers(context.init.headers));
       headers.set(httpHeader, apiKey);
-      context.init.headers = headers;
```
zone117x, line 113: This fixes a bug where in some instances the `headers` property is a read-only `getter` and cannot be re-assigned.

scottsolowild:
This is going into [its own repo](https://github.com/hirosystems/stacks-rpc-client) and will be closed when that's ready. /cc @janniks

PR #1746, opened 2024-10-17
feat: detect chainID by default for testnet networks in tx builder functions

zone117x:
> This PR was published to npm with the version `6.16.1`
> e.g. `npm install @stacks/common@6.16.1 --save-exact`

Updates the transaction builder functions to make them lookup the network `chain_id` by default when constructing txs for a custom testnet network. Behavior for mainnet and for primary testnet is unchanged.

**Non-breaking**
If the `/v2/info` request fails, e.g. when constructing a tx either offline or without expecting the network RPC to be available, then by default it will fallback to using the chainID already specified in the `network` object (which is typically either the default primary testnet chainID, or a user-specified chainID).

### DIFF
--- a/packages/cli/tests/cli.test.ts
+++ b/packages/cli/tests/cli.test.ts
@@ -24,7 +24,7 @@ import {
   WalletKeyInfoResult,
 } from './derivation-path/keychain';
 import * as fixtures from './fixtures/cli.fixture';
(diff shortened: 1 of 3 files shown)


janniks:
Closing as not-needed for now. Can reopen if this comes up again.

PR #1747, opened 2024-10-18
fix: Fix string parsing in signature decode

janniks:
> This PR was published to npm with the version `6.14.1-pr.75+b2acdcaf`
> e.g. `npm install @stacks/common@6.14.1-pr.75+b2acdcaf --save-exact`

- Fixes what I think is a bug, but am not sure @beguene do you remember how this was intended. Since it's a signature I'm assuming it should NOT be parsed as UTF8 ever. 🤷🏻
- Minor refactoring, comments, migrations, deprecations, etc.


beguene (approved):
Great catch ! 
You are right, it's definitely a bug because cryptographic signatures are usually in hex and parsed as hex (or byte arrays). And parsing as UTF-8 could lead to some incorrect byte conversion.

👍

PR #1748, opened 2024-10-21
fix!: Switch to a network and client param solution

janniks:
> This PR was published to npm with the version `6.14.1-pr.78+8cf2e30c`
> e.g. `npm install @stacks/common@6.14.1-pr.78+8cf2e30c --save-exact`

- another pivot on network params

---

The previous solution had some confusing aspects. E.g when using devnet (which is only a URL, and otherwise not different from testnet)

Now, all functions that need fetch, can take a `network` param to specify which URL+fetchFn to use. All of these function ALSO take the new `client` param. But it's okay to supply `network`, `client`, both or neither. Stacks.js will default to `mainnet` or infer which network to use. This way we can easily override URL for APIs, but also create manageable `network` objects that can be passed around everywhere (what people do currently in Stacks.js code).

Want to create your custom network?
```ts
const network = {
  ...STACKS_TESTNET,
  chainId: 666,
  client: { baseUrl: "node6.org" }
}

makeSTXTokenTransfer({ network, // ...
```

Override things if you want for `client` stuff.
```ts
makeSTXTokenTransfer({
  network,
  client: { baseUrl: "node7.com" } // client overrides client in network
  // ...
```

This also makes it easier to combine network and API/RPC clients.

```ts
const client = createClient();

makeSTXTokenTransfer({
  client, // objects are mostly compatible
```

This way we always know if a function does something with it's fetching/`client` -- e.g. 
```ts
selectStxDerivation({
  network,
  // doesn't take `client` param, we know it doesn't do networking
```

### DIFF
--- a/.github/MIGRATION.md
+++ b/.github/MIGRATION.md
@@ -2,13 +2,12 @@
 
 - [Stacks.js (\>=5.x.x) → (7.x.x)](#stacksjs-5xx--7xx)
   - [Breaking Changes](#breaking-changes)
+  - [Reducing Wrapper Types](#reducing-wrapper-types)
   - [Stacks Network](#stacks-network)
     - [Impacts](#impacts)
-  - [Fetch Methods](#fetch-methods)
-  - [Reducing Wrapper Types](#reducing-wrapper-types)
-  - [Stacks Network](#stacks-network-1)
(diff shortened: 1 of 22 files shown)


zone117x (approved):
Solid solution 👍

PR #1749, opened 2024-10-22
refactor: Minor refactoring to match common theme

janniks:
> This PR was published to npm with the version `6.14.1-pr.81+e6f2832a`
> e.g. `npm install @stacks/common@6.14.1-pr.81+e6f2832a --save-exact`

- refactor some uint8array methods to match the general theme
- remove unneeded overloads
- add @ignore and deprecation comments


packages/stacking/src/utils.ts
```diff
@@ -261,26 +272,28 @@ function legacyHashModeToBtcAddressVersion(
 
 function _poxAddressToBtcAddress_Values(
   version: number,
-  hashBytes: Uint8Array,
+  hash: string | Uint8Array,
   network: StacksNetworkName
 ): string {
   if (!StacksNetworks.includes(network)) throw new Error('Invalid network.');
 
+  hash = typeof hash === 'string' ? hexToBytes(hash) : hash;
```
hugoclrd:
```suggestion
  if (typeof hash === 'string') hash = hexToBytes(hash);
```

avoid the `hash = hash` situation, avoiding the re-assignation, also making the logic very slightly more readable i think

hugoclrd (changes requested):
Small improvement suggestion

packages/stacking/src/utils.ts
janniks: Done ✅

hugoclrd (approved):
lgtm

PR #1751, opened 2024-10-23
refactor update stackstransaction constructor to options object

janniks:
> This PR was published to npm with the version `6.14.1-pr.87+6d66e7a4`
> e.g. `npm install @stacks/common@6.14.1-pr.87+6d66e7a4 --save-exact`

- Rename StacksTransaction, so we can use the name for a better more human readable alternative in the future
- Update to accept an options object instead

### DIFF
--- a/packages/transactions/src/postcondition-types.ts
+++ b/packages/transactions/src/postcondition-types.ts
@@ -40,3 +40,5 @@ export type NonFungiblePostCondition = {
 };
 
 export type PostCondition = StxPostCondition | FungiblePostCondition | NonFungiblePostCondition;
+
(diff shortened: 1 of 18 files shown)


packages/transactions/src/postcondition-types.ts
```diff
@@ -40,3 +40,5 @@ export type NonFungiblePostCondition = {
 };
 
 export type PostCondition = StxPostCondition | FungiblePostCondition | NonFungiblePostCondition;
+
+export type PostConditionModeName = 'allow' | 'deny';
```
beguene, line 44: 👍

beguene (approved):

PR #1756, opened 2024-11-04
fix: Default contract deploys to Clarity version 3

janniks:
- fix: default to clarity 3, since this is now live and we default to latest as expected behavior


packages/transactions/tests/builder.test.ts
```diff
@@ -716,6 +716,7 @@ test('Make smart contract deploy (defaults to versioned smart contract, as of 2.
     senderKey,
     fee,
     nonce,
+    clarityVersion: 2,
```
beguene, line 719: Should we have a test for version 3 ?

beguene (approved):
LGTM 🚀 !

(there is a [test](https://github.com/hirosystems/stacks.js/actions/runs/11670385857/job/32494477251?pr=1756) failing though)

PR #1766, opened 2024-11-18
fix: Update fetch input types to account for Request object as first …

janniks:
> This PR was published to npm with the version `7.0.3-pr.4+64a49b58`
> e.g. `npm install @stacks/common@7.0.3-pr.4+64a49b58 --save-exact`

- I think we need this for better compat with the `createClient` approach. Mainly for types, but also a fetch param could look different from the only string format supported currently (even though maybe this mainly already works... probably)
- Added a `url: fetchParams.input` so this won't impact current code (just show deprecation notice)

### DIFF
--- a/packages/common/src/fetch.ts
+++ b/packages/common/src/fetch.ts
@@ -37,7 +37,10 @@ export const setFetchOptions = (ops: RequestInit): RequestInit => {
 };
 
 /** @ignore */
-export async function fetchWrapper(input: RequestInfo, init?: RequestInit): Promise<Response> {
+export async function fetchWrapper(
+  input: RequestInfo | URL,
+  init?: RequestInit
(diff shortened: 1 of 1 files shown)


packages/common/src/fetch.ts
```diff
@@ -46,11 +52,11 @@ export async function fetchWrapper(input: RequestInfo, init?: RequestInit): Prom
   return fetchResult;
 }
 
-export type FetchFn = (url: string, init?: RequestInit) => Promise<Response>;
+export type FetchFn = (input: RequestInfo | URL, init?: RequestInit) => Promise<Response>;
```
zone117x, line 55: Should this just be:
```ts
export FetchFn = typeof globalThis.fetch;
```
?

PR #1784, opened 2025-02-03
fix: use correct types for postConditions

zone117x:
> This PR was published to npm with the version `7.0.3-pr.9+878694d9`
> e.g. `npm install @stacks/common@7.0.3-pr.9+878694d9 --save-exact`

The postConditions property on the txs object was a `LengthPrefixedList` containing values of type `StacksWire` rather than `PostConditionWire`.

Before:
```ts
    const postConditions = tx.postConditions.values.map(
      p => p as PostConditionWire
    );
    postConditions.forEach(p => console.log(p.principal));
```

After:
```
    tx.postConditions.forEach(p => console.log(p.principal));
```

Before:
```ts
  const badPostConditions = createLPList([createMemoString('1234')]);
  new StacksTransactionWire({
    // Compiles without error
    postConditions: badPostConditions,
  });
```

After:
```ts
  const badPostConditions = createLPList([createMemoString('1234')]);
  new StacksTransactionWire({
    // @ts-expect-error
    postConditions: badPostConditions,
  });
```


janniks (approved):
Beautiful change! Thanks for the type magic 🙏

PR #1785, opened 2025-02-03
fix: regression in 'NoEstimateAvailable' check

zone117x:
> This PR was published to npm with the version `7.0.3-pr.6+dc8b1728`
> e.g. `npm install @stacks/common@7.0.3-pr.6+dc8b1728 --save-exact`

The stacks-core RPC still has a bug where the `NoEstimateAvailable` response can be a plain string rather than structured json. See https://github.com/stacks-network/stacks-core/issues/4145

```
Estimator RPC endpoint failed to estimate tx ContractCall: NoEstimateAvailable
```

This PR detects `NoEstimateAvailable` when the response is a string.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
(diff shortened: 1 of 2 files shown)


janniks (approved):
👍

PR #1795, opened 2025-05-07
fix: Fix a regression from the v7.x.x update where the localnet/mock…

janniks:
> This PR was published to npm with the version `7.0.7-pr.2+de94b530`
> e.g. `npm install @stacks/common@7.0.7-pr.2+de94b530 --save-exact`

- Fixes a regression that always used testnet/mainnet when we switched to the static network object style.
- Adds tests to avoid future regressions


zone117x (approved):
LGTM!

PR #1809, opened 2025-11-12
ci: switch to trusted publisher

jannik-stacks:
- add trusted publisher setup
- merge publish and publish-beta workflows

Follows the guidelines from https://docs.npmjs.com/trusted-publishers.
Keeps the same lerna/versioning setup as before.


.github/workflows/publish.yml
```diff
@@ -21,28 +25,58 @@ jobs:
...
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-
       - name: Publish
         run: |
           npx lerna publish from-package --no-verify-access --yes
           RELEASE=$(cat lerna.json | jq -r '.version')
           git tag -a v$RELEASE -m v$RELEASE
           git push origin v$RELEASE
+
+  publish-beta:
```
hstove-stacks, line 49: I think this is a copypasta
```suggestion
  publish:
```

jannik-stacks: nope, this was merging two workflow files into one, two sections. i should have specified in the description

hstove (approved):
LGTM!

hstove-stacks (approved):

hstove-stacks (approved):

PR #1810, opened 2025-11-12
feat: Add ClarityVersion `.Clarity4`

jannik-stacks:
Small change to default to Clarity4 for new contracts and allow `4` as the encoded enum.


hstove-stacks (approved):
lgtm!

PR #1813, opened 2025-11-13
build: update lerna

jannik-stacks:
- Update `lerna` dependency (potentially needed for trusted publishers)
- Update workflows to latest node version, when using `lerna`


hugo-stacks (dismissed):
lgtm

hugo-stacks (approved):
lgtm

PR #1839, opened 2026-02-18
fix: Throw on post-condition amounts not in valid range

### LINKED ISSUE #1342: `createSTXPostCondition` and `createFungiblePostCondition` allows to create invalid post conditions.
**Describe the bug**
Both `createSTXPostCondition` and `createFungiblePostCondition` accepts values higher than `18446744073709551615` (max value of u64) and as a result they produce invalid post-conditions.

From https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md
> ### A STX post condition body is encoded as follows:
> ...
> An **8-byte** value encoding the literal number of microSTX
>
> ### A Fungible token post-condition body is encoded as follows:
> ...
> An **8-byte** value encoding the literal number of token units

8-byte value (u64) not 16-byte value (u128)

jannik-stacks:
- fixes https://github.com/stx-labs/stacks.js/issues/1342

### DIFF
--- a/packages/transactions/src/pc.ts
+++ b/packages/transactions/src/pc.ts
@@ -1,4 +1,4 @@
-import { IntegerType, intToBigInt } from '@stacks/common';
+import { IntegerType } from '@stacks/common';
 import { ClarityValue } from './clarity';
 import {
   FungibleComparator,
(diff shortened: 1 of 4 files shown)


hugo-stacks (approved):
lgtm!
It's nice closing old issues!

PR #1843, opened 2026-03-13
fix: Add Clarity5 option (without defaulting to it yet)

jannik-stacks:
- Add `5` option in Clarity version.
- Do NOT add defaults to `5` yet, since it's not live on mainnet yet (or even testnet) -- We will make those changes later, since the assumption for some methods is to use the `latest` clarity version (hardcoded).


hugo-stacks (approved):
lgtm!

PR #1853, opened 2026-04-08
chore: pre commit signed check

jannik-stacks:
- add a new hook to detect git signing, so contributors add verified commits
- update husky dep


adriano-stacks (dismissed):

edgar-stacks (dismissed):

hugo-stacks (approved):
lgtm

PR #1854, opened 2026-04-09
feat: Add bitcoin-staking package

jannik-stacks:
## Add `@stacks/bitcoin-staking` package for PoX-5 staking flows

- Adds a new `packages/bitcoin-staking` workspace package: unsigned transaction builders for every PoX-5 entrypoint (`buildSetupBond`, `buildRegisterForBond` with BTC/sBTC lockup shapes, `buildStake`/`buildStakeUpdate`/`buildUnstake`, `buildAnnounceL1EarlyExit`, rewards and signer-grant builders), plus read helpers covering PoX info, bonds, memberships, signer sets, and the full rewards ledger.
- Ships client-side mirrors of the contract's consensus-critical logic: P2WSH lockup script construction and address derivation (`buildLockScript`, `buildLockAddress`), SPV proof assembly for L1 registrations (`buildLockProof`, `computeMerkleBranch`), reclaim transaction building/signing for expired or early-exited locks, and SIP-018 signer-key grant signing/verification — each pinned against on-chain results by golden-vector tests.
- Adds an eligibility preflight layer (`fetchEligible*` per entrypoint) that re-runs the contract's guard chain off-chain and returns the abort reasons in contract evaluation order, plus cycle/bond timing math (`bondStatus`, `isInPreparePhase`, `minUstxForSatsAmount`) and a `describePox5Error` code dictionary.
- Two small fixes outside the new package: `hexToBigInt` in `@stacks/common` accepts an optional `0x` prefix, and `Address.parse` in `@stacks/transactions` returns `Required<AddressRepr>` (runtime shape unchanged).

Tests run in three layers, which is worth knowing when reviewing: pure offline unit tests, a dockerized regtest suite, and a suite against a hosted private testnet — the latter two record live HTTP fixtures once (`RECORD=1`) and replay them offline in CI, so fixture JSON dominates the diff line count. Flows that mutate one-shot chain state (e.g. reward claims) are live-only and skipped in replay.

### Why

Integrators currently have to hand-roll PoX-5 contract calls, lockup scripts, and SPV proofs; this package provides those as tested, composable helpers.

### Example

```ts
import {
  buildLockOutputScript,
  buildLockProof,
  buildRegisterForBond,
  buildUnlockScript,
  fetchEligibleRegisterForBond,
} from '@stacks/bitcoin-staking';

const unlockBytes = buildUnlockScript(stakerPublicKey);
const outputScript = buildLockOutputScript({ stxAddress, unlockHeight, unlockBytes, earlyUnlockBytes });

// SPV proof for the confirmed L1 lockup tx (inputs come from any Esplora API).
const lockupOutput = buildLockProof({ txHex, header, merkleProof, txCount, unlockHeight, outputScript });

const unsigned = await buildRegisterForBond({
  bondIndex,
  signerManager,
  amountUstx,
  lockup: { kind: 'btc', outputs: [lockupOutput], unlockBytes },
  publicKey: stakerPublicKey,
  fee,
  nonce,
  network,
  postConditionMode: 'allow', // register locks STX; Deny mode would abort it
});

// Optional preflight: returns the contract's abort reasons in evaluation order.
const { eligible, reasons } = await fetchEligibleRegisterForBond({ ... });
```


packages/bitcoin-staking/src/script.ts
```diff
@@ -0,0 +1,492 @@
...
+      32,
+      'EQUALVERIFY',
+      'SHA256',
+      stakerHash,
+      'EQUALVERIFY',
+    ]),
+    earlyUnlockBytes, // per-bond early-unlock subscript (ELSE branch)
+    btc.Script.encode(['ENDIF', 'VERIFY']),
+    unlockBytes // staker subscript — runs last in BOTH branches
+  );
```
edgar-stacks: Potentially critical

buildLockScript concatenates earlyUnlockBytes (an opaque per‑bond blob from on‑chain protocol-bonds, via fetchBond) raw into the OP_ELSE branch (line 251) with only a hex→bytes decode — no non‑empty check, no Script.decode‑ability check, no stack‑balance check. This is asymmetric with the function's two other inputs (unlockHeight is validated by serializeCScriptNum; unlockBytes is 33‑byte‑checked by buildUnlockScript), and contradicts the function's own docstring that both subscripts "MUST leave a valid (boolean) result on the stack."

Verified by reproduction: a malformed/truncated push (e.g. 0x21 followed by only 2 bytes) makes the trailing push opcode swallow the scaffold's ENDIF/VERIFY and the staker CHECKSIG tail, yielding an unbalanced OP_IF with no OP_ENDIF — a hard Bitcoin consensus failure in both branches. Yet btc.p2wsh still derives a valid, fundable mainnet address. The staker funds lockAddress (from buildRegisterMetadata) before any on‑chain check, so the sats become permanently unspendable.

packages/bitcoin-staking/src/btc-address.ts
```diff
@@ -0,0 +1,197 @@
...
+      const words = bech32.toWords(data);
+      return bech32.encode(SegwitPrefix[networkName], [SEGWIT_V0, ...words]);
+    }
+    case PoXAddressVersion.P2TR: {
+      const words = bech32m.toWords(data);
+      return bech32m.encode(SegwitPrefix[networkName], [SEGWIT_V1, ...words]);
+    }
+    default:
+      throw new Error(`Unexpected address version: ${version}`);
+  }
```
edgar-stacks, line 241: stringify() switches on the PoX version but never checks data.length. The base58 branch calls base58CheckEncode, which does hash.slice(0, 20) (packages/encryption/src/keys.ts:81) and silently truncates. The segwit branches have no length check at all (verified against @scure/base genBech32().encode — it only enforces a 90‑char total limit, no witness‑program validation). So a P2WPKH version with 32‑byte data emits a P2WSH‑shaped program, etc. parse() enforces 20‑vs‑32 via nativeAddressToSegwitVersion, but the encode path has no symmetric guard — and unvalidated data reaches it from two documented entry points: fromPoxTuple() (reads on‑chain Clarity buffers) and parseSignerCalldata() (signer.ts:155‑159), whose docstring shows feeding the result straight into stringify.

packages/bitcoin-staking/tests/helpers/utils.ts
```diff
@@ -0,0 +1,376 @@
...
+    // multi-account test reads e.g. get-bond-membership for two stakers, so key
+    // by sender + args, not just the path.
+    if (url.pathname.includes("/contracts/call-read/")) {
+      return `${path}#${String(parsed.sender ?? "")}:${JSON.stringify(parsed.arguments ?? [])}`;
+    }
+  } catch {
+    // not JSON — fall through to path keying
+  }
+  return path; // e.g. /v2/transactions (broadcast body varies; response unused)
+}
```
edgar-stacks, line 276: fixtureKey() keys Stacks broadcast POSTs by path alone (return path; // broadcast body varies). In replay mode every broadcast resolves to the same recorded success regardless of the serialized tx bytes. The regtest action suite runs in CI under replay, so it appears to exercise build→broadcast but never inspects the encoded Clarity args, post‑conditions, PoX address, or SPV proof. A regression in arg order/types, a missing post‑condition, or a wrong proof encoding would still "broadcast successfully" and pass. Confirmed no test asserts a golden serialized txHex for any Stacks builder.

packages/bitcoin-staking/tests/reclaim.test.ts
```diff
@@ -0,0 +1,159 @@
...
+const LOCK_SCRIPT = buildLockScript({
+  stxAddress: STX_ADDRESS,
+  unlockHeight: UNLOCK_HEIGHT,
+  unlockBytes: buildUnlockScript(STAKER_PUB),
+  earlyUnlockBytes: buildUnlockScript(COSIGNER_PUB),
+});
+
+const UTXO: Utxo = { txid: 'a'.repeat(64), vout: 0, value: 30_000n };
+
+describe('buildReclaim', () => {
```
edgar-stacks, line 63: reclaim.test.ts builds/signs/finalizes both spend paths but every assertion is structural (finalScriptWitness length 3/5, isFinal===true, byte‑identical sigs). reclaim.ts builds every tx with disableScriptCheck:true (TX_OPTS), and @scure/btc-signer ships no script interpreter — so nothing executes the lockup opcodes. The only tests that prove the witness satisfies the script are test.skip. A swapped staker/cosigner pubkey, wrong CLTV/sequence pairing, or wrong opcode order would still pass CI yet be rejected by consensus → unspendable BTC.

edgar-stacks:
@jannik-stacks tagged a few things i found with claude review, these could be false positives but i wanted to flag these just in case

packages/bitcoin-staking/src/script.ts
```diff
@@ -0,0 +1,584 @@
...
+ *
+ * Structural checks (always): the bytes are non-empty and decode as Bitcoin
+ * script (a truncated push fails to decode).
+ *
+ * Shape heuristic (`shape: true`, the default): the fragment contains at least
+ * one 33-byte public-key push, ends in `OP_CHECKSIG`/`OP_CHECKMULTISIG` (it
+ * must leave a boolean for the shared `OP_VERIFY`), and contains no
+ * `OP_IF`/`OP_NOTIF`/`OP_ELSE`/`OP_ENDIF` (which would unbalance the outer
+ * `OP_IF … OP_ENDIF` scaffold). Disable for bond templates that are
+ * intentionally exotic but chain-valid.
```
djordon-stacks: We can make this requirement tighter, since we know that there is only one key in the early unlock part of the script.

rafa-stacks (dismissed):

edgar-stacks (dismissed):

edgar-stacks (dismissed):

edgar-stacks (approved):

packages/bitcoin-staking/src/script.ts
jannik-stacks: Can you recheck the latest tightness? 🙋‍♂️

PR #1862, opened 2026-05-15
ci: pin gh actions versions

hugo-stacks:
### Description

Pin github actions version, fix ci


jannik-stacks (approved):
yay, thanks so much for this

PR #1874, opened 2026-06-29
ci: changeset version update

jannik-stacks:
- add github token where needed
- tighten permissions and move to per job instead or per whole file


hstove-stacks (approved):
LGTM!

PR #1886, opened 2026-09-15
chore: remove unused dependencies, dead files, and internal exports found by knip

hugo-stacks:
Ran [knip](https://knip.dev) across the monorepo and audited every finding by hand.

- Remove 14 dead files: 12 duplicate `tests/setup.js` (the shared `configs/jestSetup.js` is what jest loads), an empty `hmacSha256.ts`, an orphaned CLI `auth.ts`, and a superseded transactions test that never matched jest's pattern.
- Remove 30 unused dependencies (legacy express stack in CLI, webpack 4 browser polyfills, `@stacks/network` listed twice in storage, `yalc`/`tsdx`/`depcheck` and their broken wallet-sdk scripts, etc.). Net −136 packages in the lockfile.
- Add 3 devDependencies bitcoin-staking tests were importing via hoisting only (`c32check`, `@scure/bip32`, `@jest/test-sequencer`), plus `process`/`util` for the webpack configs.
- Drop `export` from ~45 internal-only symbols, delete 6 fully dead functions, export `ProfileImage` since it appears in a public type.
- Add `knip.jsonc` encoding the false positives (commitlint via npx, bitcoin-staking dev scripts, deprecated aliases). Knip is not a dependency; run with `npx knip@latest`.

Not touched: 23 unused exports in `bitcoin-staking/tests/helpers`, left for that package's owner.

Typecheck, lint, and all package test suites pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/transactions/tests/namespaces/address.ts
jannik-stacks, line 1: let's keep this file and rename to .test.ts

jannik-stacks (changes requested):
generally good, minor change request

packages/transactions/tests/namespaces/address.ts
hugo-stacks: This one is already covered

1. `address.test.ts` already exists in this same directory, so the rename would collide. It's a strict superset of this file — same three `Address.stringify` assertions, plus `Address.parse`, `Address.fromPrivateKey`, and `Address.fromPublicKey`.
2. Because this file lacks the `.test.` suffix it never ran, and its third assertion is wrong. It expects `ST000000000000000000002Q6VF78.pox` (the mainnet checksum with a `T` prefix), but `Address.stringify` actually returns `ST000000000000000000002AMW42H.pox`. I confirmed by running it:

```
Expected: "ST000000000000000000002Q6VF78.pox"
Received: "ST000000000000000000002AMW42H.pox"
```

`address.test.ts` asserts the correct value. So renaming this file would add a failing test alongside duplicate coverage.

Happy to restore it under a different name if you'd rather keep it around

jannik-stacks: oh weird, yeah, then good to remove

jannik-stacks (approved):
nice cleanup

